PR #50, opened 2023-01-27
Exchange improvements

wouterkool:


N-Wouda:
I think we can revisit this soon-ish, probably next week.

wouterkool:
> I think we can revisit this soon-ish, probably next week.

Took a bit longer but I'm closing this as it will be implemented and generalized upon by https://github.com/PyVRP/PyVRP/pull/253.

PR #58, opened 2023-02-02
CI wheel builds + deploy to PyPI

### LINKED ISSUE #8: Python package
At some point we should provide pre-compiled binaries on the Python package index. It is probably easiest to do so if we do the deployment through GitHub Actions.

I have experience deploying regular Python packages to PyPI. What's new here are the binaries. There are some examples available of how to build binary wheels in the workflows: [here](https://github.com/pypa/cibuildwheel). 

We also need to provide typing stubs for our binary wheels. An example of how all that works is [more-itertools](https://github.com/more-itertools/more-itertools), but it's pretty simple: just provide the relevant `.pyi` files.

### LINKED ISSUE #40: CMake structure
The current CMakeLists define targets/libraries in a pretty ad-hoc manner. That's not that easy to maintain. I'll have a look at some other projects to figure out how other people deal with this.

N-Wouda:
This PR closes #8, and closes #40, using the meson build system and cibuildwheel. In particular, it:
- [x] Build wheels in CI (cibuildwheel).
- [x] Add a saner build system (meson).
- [x] Adds checked/corrected type stubs for all compiled assets.


N-Wouda:
Nice, code coverage works again with the new meson build system!

N-Wouda:
@leonlan @wouterkool can you review this? There's a lot of internal movement going on, with bindings moving into the cpp files. But very little of that is actual content changes.

The wheel builds in the CI runner currently work for Linux/x86-64. They do not yet work for Windows, OSX, or x86-32 builds. Doing that will require changes to the C++ codebase. I'll make a separate issue for that, since this PR is already big enough.

leonlan:
Will look at this tomorrow!

meson.build
```diff
@@ -0,0 +1,179 @@
+project('pyvrp', 'cpp', default_options : ['cpp_std=c++20'])
+
+# The following lines define compiler and linker options. The options are set
+# per buildtype: for example, debug builds are also given a few coverage 
+# settings, whereas release builds optimise for speed.
+if get_option('buildtype') == 'release'
+    add_project_arguments('-O3', language: 'cpp')
```
eli-schwartz, line 7: -O3 is already included in the release buildtype, see the chart at https://mesonbuild.com/Builtin-options.html#core-options

meson.build
```diff
@@ -0,0 +1,179 @@
...
+    add_project_link_arguments('-flto', language: 'cpp')
```
eli-schwartz, line 8: You could set `-Db_lto=true` as an option to Meson, or add to the `project()` function the keyword argument:
```meson
project(..., default_options: ['b_lto=true'])
```

to make that option be the default without manually specifying it on the meson command line.

meson.build
```diff
@@ -0,0 +1,179 @@
...
+
+    if compiler.get_id() == 'gcc'
+        # clang does not have this particular flag, so we only add it when
+        # compiling with gcc. It's helpful in determining code coverage.
+        add_project_arguments('-fprofile-abs-path', language: 'cpp')
+    endif
+
+    add_project_arguments('--coverage', language: 'cpp')
+    add_project_link_arguments('--coverage', language: 'cpp')
+endif
```
eli-schwartz, line 22: This can be handled by the `-Db_coverage=true` option, more or less. It hooks up `--coverage` and produces build targets for `ninja coverage`.

See https://mesonbuild.com/Feature-autodetection.html#coverage for discussion of coverage support.

...

I do not think that fprofile-abs-path is needed since Meson only uses a single working directory -- anyway, the coverage report generator should set things up to work. I guess that flag couldn't hurt...

meson.build
```diff
@@ -0,0 +1,179 @@
...
+
+assert(pybind11.found(), 'Could not find pybind11!')
+
+# Now that everything is in place, let's defines the extensions to be built by
+# Meson. These are given as tuples on the following lines. The tuples look like
+# (extension name, source files, install dir), where 'extension name' is the
+# eventual module name, the source files name cpp source files (in pyvrp/cpp), 
+# and 'install dir' gives the directory where the extension should be placed.
+
+INSTALL_DIR = meson.source_root() / 'pyvrp'
```
eli-schwartz, line 47: I'm not familiar with how poetry handles this (I know more about https://github.com/mesonbuild/meson-python). Do you really need to pass the install dir like this? Maybe it's possible to configure `meson setup -Dpython.platlibdir={source_dir}`?

meson.build
```diff
@@ -0,0 +1,179 @@
...
+
+foreach extension : extensions
+    message('Going to build ' + extension[0])
+
+    py.extension_module(
+        extension[0],
+        extension[1],
+        dependencies : [py.dependency(), pybind11],
+        install : true,
+        install_dir: extension[2],
```
eli-schwartz, line 165: ... and here, you could use `subdir: 'pyvrp/crossover'` or similar, and Meson would install to the configured platlibdir override, as '{platlibdir}/pyvrp/crossover/'.

meson.build
```diff
@@ -0,0 +1,179 @@
...
+            include_directories(SRC_DIR),
+        ],
+        cpp_args: [
+            '-Werror',
+            '-Wall',
+            '-Wextra',
+            '-Wpedantic',
+            # Ignore unused parameters, because not all LS operators use every
+            # argument they are given.
+            '-Wno-unused-parameter'
```
eli-schwartz, line 176: I would set -Werror as `default_options: ['werror=true']`.

-Wall is already the default because the default `warning_level=1` option sets that. But you can add `default_options: ['warning_level=3']`, 2 sets -Wextra and 3 sets -Wpedantic.

The only one you need to set manually is -Wno-unused-parameter. This won't work on Windows compilers, but you could use [`compiler.get_supported_arguments()`](https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerget_supported_arguments) to check if it's supported.

(Using the builtin core options for werror/warning_level gets you cross-platform flag handling for free.)

meson.build
```diff
@@ -0,0 +1,179 @@
...
+    ['ProblemData', SRC_DIR / 'ProblemData.cpp', INSTALL_DIR],
+    ['TimeWindowSegment', SRC_DIR / 'TimeWindowSegment.cpp', INSTALL_DIR],
+    ['XorShift128', SRC_DIR / 'XorShift128.cpp', INSTALL_DIR],
+    [
+        'Individual', 
+        [
+            SRC_DIR / 'Individual.cpp', 
+            SRC_DIR / 'ProblemData.cpp', 
+            SRC_DIR / 'PenaltyManager.cpp', 
+            SRC_DIR / 'XorShift128.cpp'
```
eli-schwartz, line 62: A bunch of these files seem to be the same files used in multiple extensions.

Have you considered building them into a `libcommon = static_library('common', sources_list, dependencies : [py.dependency(), pybind11])` and adding `link_with: libcommon` to the extensions?

This would mean that each file is only built once, and then whichever functions are referenced from the main module implementation would be included at link time of the extension.

(It's not clear to me what this is doing... are you building lots of code in each extension including multiple `PyInit_modulename` initialization functions, while only one initialization function is used at runtime depending on which module was actually built?)

The benefit of using a private "libcommon" style approach is that you save a bit of work while compiling, and this can add up over time.

N-Wouda:
Hi @eli-schwartz! 👋 Thanks for the review on the meson stuff! I'll update our `meson.build` later today, when I've had some time to tinker with it more :).

Great work on meson btw; I'm very impressed with it so far!

meson.build
N-Wouda: No, that's not required by poetry. I'm pretty new to this whole thing, so I tinkered with it until it worked. I based this particular bit on how [pendulum](https://github.com/sdispater/pendulum/blob/master/meson.build) does it. Your suggestion of generalising this is nice and works, so I am using that now!

N-Wouda:
@eli-schwartz:
- I have implemented a `libcommon` approach to our build, and moved the binding definitions into their own files. That has sped everything up quite a bit: thanks!
- I'm using `subdir` and a custom `python.platlibdir` for extension installation.
- I've moved most default CPP options into the project's `default_options` argument, and use `compiler.has_argument(.)` to check for `-Wno-unused-parameter`. 

The coverage stuff I'm leaving as-is, for now. The current way works well enough for gcovr to provide good reports, and that was my main motivation. I can probably make that work using meson's coverage setting (and then hoepfully also drop the `-fprofile-abs-path` flag), but that is not yet high on my todo list. I have copied your comment to a new issue (#68), and will revisit this in the near future to set coverage up a bit nicer.

Thank you for your review!

pyproject.toml
```diff
@@ -50,12 +50,14 @@ pytest = ">=6.0.0"
 pytest-cov = ">=2.6.1"
 codecov = "*"
 
-# These are used in the build script: for compiling the library (cmake/pybind11)
-# and for generating type stubs (mypy).
-cmake = "^3.25.0"
+black = "^22.12.0"
```
leonlan, line 53: Question: do we want to keep `black` as dependency in poetry or should we leave it to pre-commit?

N-Wouda: Good point. I added it initially so I can tell my IDE to use it for formatting, but I can also just rely on the system-wide Python's `black` for that. I've removed it from the dependency list.

wouterkool (approved):
Looks great to me!

Only small comment: I see different style imports, do we have a preference/guidelines which to use where? I think this also relates to the overall structure of how we organize the Python code, so let's discuss this separately.
* `from pyvrp import PenaltyManager` (e.g. in `pyvrp/crossover/selective_route_Exchange.pyi`)
* `from pyvrp.PenaltyManager import PenaltyManager` (e.g. in `pyvrp/educate/Exchange.pyi`)
* `from .ProblemData import ProblemData` (in `pyvrp/Result.py`).

N-Wouda:
@wouterkool agree. I don't really have a strong philosophy on these. In tests I typically prefer absolute imports, but in package code I use relative imports for siblings, and absolute imports (from `pyvrp`) for everything else. We should have a look at writing this down at some point. Would you be willing to open an issue for this?

PR #62, opened 2023-02-03
Example notebooks

### LINKED ISSUE #70: Move ProblemData.from_file to Python
I just released [`vrplib`](https://github.com/leonlan/VRPLIB), so it's a good moment to move `ProblemData.from_file` to the Python side and use `vrplib` to parse the instances.

https://github.com/N-Wouda/pyvrp/blob/7ca863f1893064c69df85fdeecfe9262f8e0407b/pyvrp/read.py#L23

wouterkool:
Closes #70. This PR:

- Implements various plotting tools and some diagnostic calculations.
- Provides an example VRPTW notebook
- Uses `vrplib` to read the problem data from a file.


pyvrp/diagnostics.py
```diff
@@ -0,0 +1,106 @@
...
+        Route (list of indices) to compute statistics for.
+
+    Returns
+    -------
+    RouteStatistics
+        Object with statistics for the route.
+    """
+    assert 0 not in route, "Route should not contain depot"
+
+    depot = data.client(0)  # For readability, define variable
```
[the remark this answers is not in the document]
wouterkool: Function or property? E.g. `data.depot` and `data.client[i]`? If that is difficult, what do you think about renaming the function to `get_client()` (holds at other places as well)?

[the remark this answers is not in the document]
wouterkool: Sorry I didn't read properly that the function was already there. Using it now.

pyproject.toml
```diff
@@ -53,6 +54,13 @@ pytest = ">=6.0.0"
 pytest-cov = ">=2.6.1"
 codecov = "*"
 
+[tool.poetry.group.examples]
+optional = true
+
+[tool.poetry.group.examples.dependencies]
+jupyter = ">=1.0.0"
+pandas = ">=1.5.0"
+
```
N-Wouda, line 63: The build's failing because this group is now in the middel of the dev-dependencies, so poetry thinks everything below it (including meson and pybind) is optional, too. I suggest we move the examples group up, to be between `tqdm = "^4.61.1"` and `[tool.poetry.group.dev.dependencies]` (currently, that is line 50).

wouterkool: Sorry, I missed that because of the blank line.

wouterkool:
Closes #70

pyvrp/tests/test_read.py
```diff
@@ -1,29 +1,32 @@
...
+        ("data/UnknownEdgeWeightType.txt", ValueError),
+        ("somewhere that does not exist", FileNotFoundError),
+        ("data/FileWithUnknownSection.txt", ValueError),
+        ("data/DepotNotOne.txt", ValueError),
+        # TODO: check must be done by VRPLIB
+        # ("data/DepotSectionDoesNotEndInMinusOne.txt", RuntimeError),
+        ("data/MoreThanOneDepot.txt", ValueError),
+        ("data/NonZeroDepotServiceDuration.txt", ValueError),
+        # TODO: release times not in VRBLIB format, so not supported by VRPLIB
+        # ("data/NonZeroDepotReleaseTime.txt", ValueError),
```
leonlan, line 22: Question: does `vrplib` not work with this file? It looks like it's correctly formatted in `VRPLIB` style.

N-Wouda:
I've updated a few minor things. One thing we should add is to solidify the current plots using e.g. https://matplotlib.org/stable/devel/testing.html#writing-an-image-comparison-test. I'll try to set that up tomorrow.

N-Wouda:
Here's the example in the docs: https://pyvrp--62.org.readthedocs.build/en/62/examples/vrptw.html

pyvrp/tests/test_read.py
wouterkool: My bad :) fixed now

N-Wouda:
I've added a bit more documentation, particularly to `pyvrp.diagnostics`. I'll wait for the coverage and documentation results to come in, but overall I think this is ready to be merged.

We will have to revisit the VRPTW example soon, since that needs more explanation to really work as an intro to pyvrp.

N-Wouda (approved):

PR #72, opened 2023-02-07
Compare individuals for equality

N-Wouda:
This PR adds an equality operator for individuals, so we can compare them to see if they're the same.

### DIFF
--- a/pyvrp/Population.py
+++ b/pyvrp/Population.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Callable, Iterator, List, Tuple
+from typing import Callable, Iterator, List, NamedTuple, Tuple
 
 import numpy as np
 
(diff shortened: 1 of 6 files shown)


pyvrp/Population.py
```diff
@@ -87,20 +87,31 @@ def add(self, individual: Individual):
...
+                if id(other) == id(individual):
                     del prox[idx]
                     break
 
         for item in self:  # remove individual from subpopulation.
-            if item[0] == individual:
+            if id(item[0]) == id(individual):
                 self._items.remove(item)
                 break
         else:
```
N-Wouda, line 114: This part used to already be by identity (in CPython: memory location), but it's more important to be explicit about this now that we can also compare by equality using ==.

pyvrp/Population.py
```diff
@@ -87,20 +87,31 @@ def add(self, individual: Individual):
...
         for _, _, prox in self:  # remove individual from other proximities
             for idx, (other, _) in enumerate(prox):
-                if other == individual:
+                if id(other) == id(individual):
                     del prox[idx]
                     break
 
         for item in self:  # remove individual from subpopulation.
-            if item[0] == individual:
+            if id(item[0]) == id(individual):
```
leonlan, line 111: Question: Is `id(x) == id(y)` preferred over/different from `x is y`?

leonlan, line 111: Suggestion: can we change `_Item` from tuple to a dataclass? Then we can be more explicit about what we're accessing from the items: `id(item.individual) == id(individual)`.

N-Wouda: No, I momentarily forgot that existed!

N-Wouda: I do want to keep indexing and unpacking around, so I'll go for a named tuple. That way we get the best of both worlds!

leonlan (approved):

PR #74, opened 2023-02-07
Stub for documentation

N-Wouda:


N-Wouda:
I'm merging these stubs in without review because we need this on `main` before RTD will build anything from PRs.

Issue #46, opened 2023-01-22
SearchMethod protocol


N-Wouda:
We currently have one search method, the `LocalSearch` class. If we are to support additional methods, we should standardise an interface for such methods that e.g. `GeneticAlgorithm` can rely on. This issue exists to determine that interface.

leonlan:
> Slightly off-topic: In view of our long-term ambition to possibly include algorithms other than HGS, to what extent should be keep the docs (and code structure) specific to HGS? E.g., `educate` is a HGS-specific term, and it would probably be more future proof to rename this to `local_search`.
>  ...
> _Originally posted by @leonlan in https://github.com/N-Wouda/PyVRP/pull/75#discussion_r1101053697_


>  I've given this some thought, and settled on educate because it'd also cover non-LS type of educators. I'm not entirely sure what we want to do, but we probably do want to facilitate people hybridizing with, say, ALNS (I recall a paper doing HGS + ALNS) or something else.
> 
> This also ties into #46, which might have LS derive from a base Educator class to express this intent more clearly. A discussion about this can probably take place there.
> 
> _Originally posted by @N-Wouda in https://github.com/N-Wouda/PyVRP/pull/75#discussion_r1101131530_

N-Wouda:
I suggest we call it `SearchMethod`, rather than `Educator`. We should think of an overall `SearchMethod` protocol that e.g. `GeneticAlgorithm` can rely on. In that sense, #122 duplicates this issue somewhat.

N-Wouda:
I think we can keep this pretty simple, and only impose something like this:
```python
class SearchMethod(Protocol):

  def __call__(self, sol: Solution) -> Solution:
    pass

  def on_best(self, best: Solution) -> Solution:
    pass
```
The `on_best` thing is something we'll need if we want to replicate our current GA exactly. We might eventually want to do that differently so we can simplify this further. My ideal target is a search method that also allows a simple function, that is, only requires `__call__`.

leonlan:
`on_best` was removed in https://github.com/PyVRP/PyVRP/pull/287.

N-Wouda:
Yeah you're right! I still had the old situation in my head. Great, then just a function it is!

wouterkool:
[...]

PR #75, opened 2023-02-07
API reference (docs)

N-Wouda:
This PR adds an API reference to the documentation.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,7 +15,7 @@ now = datetime.date.today()
 
 project = "PyVRP"
 author = "PyVRP contributors"
-copyright = f"2019 - {now.year}, {author}"
+copyright = f"2022 - {now.year}, {author}"
 
 with open("../../pyproject.toml", "rb") as fh:
     pyproj = tomli.load(fh)
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -14,8 +14,8 @@ sys.path.insert(0, os.path.abspath("../../"))
 now = datetime.date.today()
 
 project = "PyVRP"
-author = "PyVRP contributors"
-copyright = f"2022 - {now.year}, {author}"
+authors = "PyVRP contributors"
+copyright = f"2022 - {now.year}, {authors}"
 
 with open("../../pyproject.toml", "rb") as fh:
     pyproj = tomli.load(fh)


N-Wouda:
PenaltyParams is not consistent with e.g. GeneticAlgorithmParams (it should be PenaltyManagerParams).

docs/source/api/educate.rst
```diff
@@ -0,0 +1,67 @@
+.. module:: pyvrp.educate
+   :synopsis: Education
+
+
+Education
+=========
+
+The :mod:`pyvrp.educate` module contains classes and methods responsible for educating (improving) a newly created offspring solution.
+This happens just after :mod:`pyvrp.crossover` is performed by the :class:`~pyvrp.GeneticAlgorithm.GeneticAlgorithm`.
```
leonlan, line 9: Slightly off-topic: In view of our long-term ambition to possibly include algorithms other than HGS, to what extent should be keep the docs (and code structure) specific to HGS? E.g., `educate` is a HGS-specific term, and it would probably be more future proof to rename this to `local_search`.

This is probably not the place to discuss this in detail, but it's useful to discuss this soon/open an issue about it. What do you think?

pyvrp/ProblemData.pyi
```diff
@@ -28,16 +90,128 @@ class ProblemData:
...
+            Client or depot number.
+        second
+            Client or depot number.
+
+        Returns
+        -------
+        int
+            Travel duration between the given clients.
+        """
+    def distance_matrix(self) -> Any:
```
leonlan, line 172: Question: why is this typed `Any` instead of an array or lists of lists?

N-Wouda: Because this will very soon become a `nd.ndarray` but it isn't yet, and I didn't want to bother typing this yet :)

docs/source/api/educate.rst
N-Wouda: I've given this some thought, and settled on educate because it'd also cover non-LS type of educators. I'm not entirely sure what we want to do, but we probably do want to facilitate people hybridizing with, say, ALNS (I recall a paper doing HGS + ALNS) or something else.

This also ties into #46, which might have LS derive from a base Educator class to express this intent more clearly. A discussion about this can probably take place there.

leonlan: OK, I'll quote these comments and continue in #46.

docs/source/api/crossover.rst
```diff
@@ -0,0 +1,12 @@
...
+
+
+Crossover
+=========
+
+The :mod:`pyvrp.crossover` module provides operators that are responsible for generating a new offspring :class:`pyvrp.Individual.Individual` solution from two parent solutions.
+
+.. autoapimodule:: pyvrp.crossover.selective_route_exchange
+
+   .. autofunction:: selective_route_exchange
```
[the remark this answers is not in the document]
N-Wouda: Turns out RTD needs the compiled extensions as well, just the pyi files are not enough. Annoying. I'll try to set up our own build system on RTD in the next few commits, so we can build those extensions there.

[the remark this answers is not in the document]
N-Wouda: The nice thing about this is that we can now have our documentation dependencies in a group in `pyprojec.toml` as well.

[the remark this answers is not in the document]
N-Wouda: Seems to work again!

leonlan (approved):

PR #76, opened 2023-02-10
Test SREX

### LINKED ISSUE #36: Test SelectiveRouteExchange

leonlan:
Closes #36.


leonlan:
@N-Wouda I have implemented three tests (one easy, two more involved). 
- I can't think of more easy test cases. 
- I could add more involved test cases like the last test in order to increase coverage on SREX and greedy repair. In particular, we let the startB index move one place, and greedy repair should insert customer 3 after 1 (which is infeasible in greedy repair).

pyvrp/crossover/tests/test_selective_route_exchange.py
```diff
@@ -0,0 +1,72 @@
...
+    indiv1 = Individual(data, pm, [[1, 2], [3, 4], []])
+    indiv2 = Individual(data, pm, [[2, 3], [4, 1], []])
+    parents = (indiv1, indiv2)
+
+    # The start indices do not change because there are no improving moves.
+    # So, indiv1's route [1, 2] will be replaced by indiv2's route [2, 3].
+    # This results in two incomplete offspring [[2, 3], [4]] and [[2], [3, 4]],
+    # which are both repaired using greedy repair. After repair, we obtain the
+    # offspring [[2, 3, 1], [4]] with cost 8735, and [[1, 2], [3, 4]] with
+    # cost 9725.The first one is returned since it has the lowest cost.
```
N-Wouda, line 69:
```suggestion
    # cost 9725. The first one is returned since it has the lowest cost.
```

N-Wouda:
This is already pretty good, even though there are only a few tests. The one issue I have is that we're [not yet covering all the other improvement cases](https://app.codecov.io/gh/N-Wouda/pyvrp/blob/test-srex/pyvrp/cpp/crossover/selective_route_exchange.cpp):

image

We only currently test the case where the improvement either does not exist, or when it's achieved by moving left in A. Can we also cover the other cases? I'm not sure if that's a lot more work. If it's easy I'd like to do it now, if it's not we can add a TODO to the tests file for later.

leonlan:
I think it should be doable. Will cover them tonight.

N-Wouda:
@leonlan I think you can get `differenceBLeft` by switching the parent order of the 'A left' case. And similarly you should only need a single case for the 'right' differences, since that should (?) be symmetric as well.

leonlan:
> @leonlan I think you can get `differenceBLeft` by switching the parent order of the 'A left' case. And similarly you should only need a single case for the 'right' differences, since that should (?) be symmetric as well.

This didn't work unfortunately, but it did help me to find that there is quite some symmetry in the route ordering. I made new 3 route examples to get the "left" and "right" differences. 

I spend a bit too much time on this but the coverage is up to 95%+ for both SREX and greedy repair now :-)

pyvrp/crossover/tests/test_selective_route_exchange.py
```diff
@@ -0,0 +1,173 @@
...
+    pm = PenaltyManager(data.vehicle_capacity)
+    rng = XorShift128(seed=4)  # nMovedRoutes = 3
+
+    indiv1 = Individual(data, pm, [[1], [2], [3, 4]])
+    indiv2 = Individual(data, pm, [[1, 2], [3], [4]])
+    parents = (indiv1, indiv2)
+
+    offspring = srex(parents, data, pm, rng)
+
+    assert_equal(offspring.get_routes(), indiv2.get_routes())
```
N-Wouda, line 23: Question: how does this work? I get that they are all replaced, but why?

N-Wouda: Oh because `nMovedRoutes = 3`?

N-Wouda: If so, can we maybe have something along the lines of: "this RNG seed will cause `nMovedRoutes = 3`", in this test and in others? It wasn't immediately obvious to me what hte comment behind the rng meant.

N-Wouda (approved):
Looks OK to me, but maybe add a comment here and there to better explain what the RNG seeds do to set the (internal) SREX variables `startA` and `nMovedRoutes`.

pyvrp/crossover/tests/test_selective_route_exchange.py
N-Wouda: I don't like that we tie these lower-level parameters to the RNG state so tightly. Perhaps we should split off SREX's implementation into a separate function that takes `startA` and `nMovedRoutes` as arguments, with a shim around it that keeps the original interface.

But that's maybe something for a new issue.

PR #81, opened 2023-02-12
Unified commandline interface

### LINKED ISSUE #69: CLI
We currently have the `solve` and `benchmark` scripts that, well, solve a single instance (`solve`), or a set of instances (`benchmark`), respectively. These scripts are installed along with the package, so they are available on the command line (within the poetry environment: `pyvrp run solve <args>`).

Of course, we cannot offer general `solve` and `benchmark` entrypoints on the command line, since those names mean nothing outside the `pyvrp` context. I propose we merge both into a single CLI that we expose as `pyvrp`. This issue serves to detail what that CLI should contain.

N-Wouda:
Closes #69. This PR adds a single CLI that covers the old benchmark and solve scripts as special cases.


.github/codecov.yml
```diff
@@ -1,2 +1,3 @@
 ignore:
   - "**/*_bindings.cpp"  # ignore pybind11 bindings
+  - "pyvrp/cli.py"  # tested in other ways than unit tests
```
N-Wouda, line 3: I think this is OK, since we're going to be using the CLI ourselves a lot anyway, and it's not really a functional, standalone component.

N-Wouda:
@leonlan I think this is basically done. The interface marries the best of the previously available `solve` and `benchmark` scripts, in a single `pyvrp` entrypoint (`cli.py` is run whenever `poetry run pyvrp <args>` is called).

pyvrp/cli.py
```diff
@@ -0,0 +1,209 @@
...
+        Filesystem location of the VRPLIB instance.
+    seed
+        Seed to use for the RNG.
+    max_runtime
+        Maximum runtime (in seconds) for solving. Either ``max_runtime`` or
+        ``max_iterations`` must be specified.
+    max_iterations
+        Maximum number of iterations for solving. Either ``max_runtime`` or
+        ``max_iterations`` must be specified.
+
```
leonlan, line 74: `instance_format` and `round_func` are not described in the docstring.

leonlan (approved):
LGTM!

PR #82, opened 2023-02-12
Remove Problem::fromFile

leonlan:
Removes `Problem::fromFile` from the C++ source code, which was replaced in #70.


pyvrp/cpp/ProblemData.cpp
```diff
@@ -250,8 +30,6 @@ ProblemData::ProblemData(std::vector<std::pair<int, int>> const &coords,
       numVehicles_(numVehicles),
       vehicleCapacity_(vehicleCap)
 {
-    // TODO argument checks (partially from ProblemData::fromFile)
-
```
N-Wouda, line 254: Can we keep the TODO here, and only get rid of the bit in parens? I do think we should do argument checks here at some point.

N-Wouda (approved):

PR #83, opened 2023-02-12
Speed up Population

N-Wouda:


N-Wouda:
TODO: add tests that verify proximity and pop are sorted.

N-Wouda:
This branch seems somewhat broken in various tests I tried, and has become superfluous with #120.

PR #85, opened 2023-02-13
Bugfix: population individual sorting:

wouterkool:
* Fix incorrect comparison function for proximity structure (was non-complete). Note that the old version was not hierarchical, it was incorrect. While we can implement a hierarchical function, the cost is actually not relevant as the sorting is only used for computing the diversity.
* Fix tie-breaker in ordering of individuals by diversity for computation of biased fitness. The function now sorts by decreasing diversity and increasing cost (like the original implementation).


N-Wouda:
Thanks!

PR #86, opened 2023-02-14
Education tests (LS operators)

### LINKED ISSUE #84: Individual == comparison based on neighbours
Implement `Individual::operator==` by comparing `Individual::neighbours` rather than `Individual::routes` which is more robust as it does not depend on the sorting of the routes in the individual

N-Wouda:
Closes #84. This PR adds rudimentary tests for various LS node and route operators.


pyvrp/cpp/educate/LocalSearch.cpp
```diff
@@ -120,8 +120,8 @@ void LocalSearch::intensify(Individual &indiv)
                     continue;
             }
 
-            if (lastModified[U.idx] > lastTested)
-                enumerateSubpaths(U);
+            // if (lastModified[U.idx] > lastTested)
+            //     enumerateSubpaths(U);
```
N-Wouda, line 124: See #90.

pyvrp/educate/tests/test_Exchange.py
```diff
@@ -0,0 +1,190 @@
...
+    ls = LocalSearch(data, pm, rng, params)
+    op = Exchange22(data, pm)
+    ls.add_node_operator(op)
+
+    individual = Individual(data, pm, [[1, 2, 3, 4]])
+    copy = Individual(individual)
+
+    ls.search(individual)
+
+    assert_equal(individual, copy)
```
wouterkool, line 167: Are you sure [3, 4, 1, 2] is an improvement over [1, 2, 3, 4], otherwise that may be the reason the solution is unchanged?

wouterkool (approved):
Nice work!

pyvrp/educate/tests/test_Exchange.py
N-Wouda: I thought so, but I'll check to be sure and make this a bit more explicit.

leonlan (approved):
LGTM!

N-Wouda:
@leonlan I've updated your restart test because it failed in the CI, but this should work.

I'm merging this in now.

PR #88, opened 2023-02-15
Population-generics

wouterkool:
- To separate VRP logic from Genetic algorithm logic, Implement population as a generic type that can hold arbitrary typed individuals.
- Add extra implementation of subpopulation that separates the computation of BiasedFitness. The new BiasedFitness class uses a numpy implementation but could be implemented in c++ as well.
- Added tests for population


wouterkool:
See https://github.com/N-Wouda/PyVRP/issues/44#issuecomment-1431249523 which suggests that the numpy implementation is not faster. Maybe we should profile actual time spent in population vs local search for these runs.

pyvrp/Population.py
```diff
@@ -2,18 +2,49 @@
...
 from .XorShift128 import XorShift128
 
-_DiversityMeasure = Callable[[ProblemData, Individual, Individual], float]
+
+class PopulationIndividual(Protocol):  # pragma: no cover
+    """
+    Defines an individual that can be put in a population target protocol.
+    """
+
+    def cost(self) -> Union[int, float]:
```
N-Wouda, line 33: `float` already covers `int`, so the union type is AFAIK unnecessary.

pyvrp/Population.py
```diff
@@ -2,18 +2,49 @@
...
+        ...
+
+    def is_feasible(self) -> bool:
+        ...
+
+
+TIndiv = TypeVar("TIndiv", bound=PopulationIndividual)
```
N-Wouda, line 40: Maybe `Solution`? Or `SolutionType`? `TIndiv` does not immediately ring a bell for me.

pyvrp/Population.py
```diff
@@ -118,13 +167,17 @@ def remove(self, individual: Individual):
         ValueError
             When the given individual could not be found in the subpopulation.
         """
-        for _, _, prox in self:  # remove individual from other proximities
+        for (
+            _,
+            _,
+            prox,
+        ) in self._items:  # remove individual from other proximities
```
N-Wouda, line 174: Can we keep something like the original here? The one-line for loop was easier to read than this five-line thing, IMHO.

pyvrp/Population.py
```diff
@@ -398,3 +564,51 @@ def get_best_found(self) -> Individual:
...
+            lambda first, second: diversity_op(data, first, second),
+            rng.randint,
+            params,
+        )
+
+        if initial_solutions:
+            for indiv in initial_solutions:
+                self.add(indiv)
+        else:
+            self.initialize(lambda: Individual(data, penalty_manager, rng))
```
N-Wouda, line 614: After going through this class, I am a little overwhelmed. So much of this exists just to satisfy the type checker, when I do not immediately see the benefit w.r.t. the code that we actually have, rather than what we might write in a few months.

I see the value of having this work for general types that satisfy some constraints (e.g., have an objective value), but I question the relatively heavy-handed use of classes and subclasses here. Is there a way to do this with less work? Do we need to branch on numpy/pure python in the type system, for example? 

(I don't know, but that's a concern I have)

N-Wouda (commented):
I don't really have issues with the code per se, but I do have a concern. That concern is that in our drive to keep the Population and fitness calculations (a) in Python, and (b) fast, we now have something that's 400-ish lines for `BiasedFitness`, and a few hundred lined for the `Population.py` file (although that does more, including a lot of typing stuff).

Say, 600 lines total. That's considerably more than the Population class in C++ was: our Euro/NeurIPS population was 100 lines in the header, and another 100-or-so in the source file. Say, 250 total. I personally find those 250 lines considerably easier to understand than the biased fitness calculation in this PR, which heavily relies on numpy to do the heavy lifting. 

I think a PR that does all this in C++, including a few of the ideas of the numpy implementation, would be much smaller, simpler, and at least as fast. Such a C++ implementation would be more heavily tied to the Individual (although we can probably use a template, and bind an instantiation of `Population<Individual>` only in the Python bindings), but we gain simplicity as a result.

So, in brief: I think there's a lot of value here (including almost everything that's in the tests, which we should definitely keep), but I am not sure if this is the right way to fix our Population.

pyvrp/Population.py
wouterkool: I copied it from here so maybe then also remove it there: https://github.com/N-Wouda/PyVRP/blob/e04aba4a0849a7c9579eac1e76d1283bf3e67170/pyvrp/OptimisationTarget.py#L9

wouterkool:
I think the only reason for implementing the population in c++ is the biased fitness computation, so I suggest implementing only the BiasedFitness class in c++, then keeping only the SubPopulation that uses the BiasedFitness internally. To me, that would be a nice sweet spot keeping things clearly separated.

wouterkool:
Agreed that we will run some more GH1000 benchmarks to compare the different implementations, and the overall importance of an efficient implementation of Population.
* If it is not important, we will stick with the Python implementation and remove the numpy version, but we will keep the generic interface.
* If it turns out to be important, we will keep the numpy version implemented in BiasedFitness, and we should look into implementing the BiasedFitness class in c++, but we can keep the population in Python.

N-Wouda:
Current main:
```
      Avg. objective: 335165
     Avg. iterations: 130764
   Avg. run-time (s): 3600.01
        Total not OK: 0
```
This PR:
```
TODO
```

pyvrp/Population.py
N-Wouda: Yes, thanks!

pyvrp/tests/test_Population.py
```diff
@@ -215,36 +234,233 @@ def test_select_returns_same_parents_if_no_other_option():
...
+    )
+
+    assert_equal(len(pop.feasible_subpopulation), 1)
+    assert_equal(len(pop.infeasible_subpopulation), 0)
+    assert_equal(pop.feasible_subpopulation.avg_distance_closest(0), 0)
+
+
+@mark.parametrize("use_numpy", [False, True])
+@mark.parametrize("nb_elite", [5, 25])
+def test_elite_not_purged_computation(nb_elite: int, use_numpy: bool):
```
N-Wouda, line 297: I moved this test over to #120 just yet.

pyvrp/tests/test_Population.py
```diff
@@ -215,36 +234,233 @@ def test_select_returns_same_parents_if_no_other_option():
...
+    assert_equal(params.min_pop_size, len(pop))
+
+    # Make sure elite individuals are among these
+    for elite_idx in np.argsort(costs)[: params.nb_elite]:
+        # Note: O(n) not most efficient way to check but will do for testing
+        assert_(indivs[elite_idx] in pop.feasible_subpopulation)
+
+
+@mark.parametrize("use_numpy", [False, True])
+def test_biased_fitness_computation(use_numpy: bool):
```
N-Wouda, line 345: Working on this test now.

wouterkool:
Not merging this as we decided to move implementation to C++, but keep around the branch as it may be useful for reference if we decide to move population to Python again.

PR #89, opened 2023-02-17
Population restart

leonlan:
This PR:
- Adds population restart back to the algorithm.
- Introduces a nearest integer rounding option.

### DIFF
--- a/pyvrp/GeneticAlgorithm.py
+++ b/pyvrp/GeneticAlgorithm.py
@@ -25,6 +25,7 @@ class GeneticAlgorithmParams:
     repair_probability: float = 0.80
     collect_statistics: bool = False
     should_intensify: bool = True
+    nb_iter_no_improvement: int = 10000
 
     def __post_init__(self):
         if self.nb_penalty_management < 0:
@@ -93,13 +94,27 @@ class GeneticAlgorithm:
         start = time.perf_counter()
         stats = Statistics()
         iters = 0
+        iters_no_improvement = 1
 
         while not stop(self._pop.get_best_found()):
             iters += 1
 
+            if iters_no_improvement == self._params.nb_iter_no_improvement:
+                self._pop.restart()
+                iters_no_improvement = 1
+
+            curr_best = self._pop.get_best_found().cost()
+
             parents = self._pop.select()
             offspring = self._op(parents, self._data, self._pm, self._rng)
             self._educate(offspring)
+
+            new_best = self._pop.get_best_found().cost()
+
+            if new_best < curr_best:
+                iters_no_improvement = 1
+            else:
+                iters_no_improvement += 1
 
             if iters % self._params.nb_penalty_management == 0:
                 self._update_penalties()
--- a/pyvrp/GeneticAlgorithm.py
+++ b/pyvrp/GeneticAlgorithm.py
@@ -25,7 +25,7 @@ class GeneticAlgorithmParams:
     repair_probability: float = 0.80
     collect_statistics: bool = False
     should_intensify: bool = True
-    nb_iter_no_improvement: int = 10000
+    nb_iter_no_improvement: int = 20_000
 
     def __post_init__(self):
         if self.nb_penalty_management < 0:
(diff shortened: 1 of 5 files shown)


leonlan:
I removed the bounds for now (they're still in the benchmark-cvrp branch).

pyvrp/read.py
```diff
@@ -85,29 +88,37 @@ def read(
...
     if "service_time" in instance:
         service_times = round_func(instance["service_time"])
     else:
         service_times = np.zeros(num_clients, dtype=int)
 
+    if "time_window" in instance:
+        time_windows = round_func(instance["time_window"])
+    else:
+        # Upper bound based on the maximum route duration
+        max_route_len = capacity // demands[demands > 0].min()
```
wouterkool: Not sure, but there may be instances with some very small demands, how about we cap `max_route_len` at `num_clients`? Maybe also support 0 demand (i.e. TSP) and check for overflow of these values?

wouterkool (commented):
LGTM, one small comment

pyvrp/Population.py
```diff
@@ -146,6 +146,12 @@ def purge(self):
             # population, and remove it.
             self.remove(max(self, key=lambda item: item.fitness).individual)
 
+    def restart(self):
```
N-Wouda, line 149: Suggestion: we can recreate the subpops in `Population.restart`, which makes this function unnecessary.

pyvrp/Population.py
```diff
@@ -351,6 +357,17 @@ def select(self) -> Tuple[Individual, Individual]:
 
         return first, second
 
+    def restart(self):
+        """
+        Restarts the population. All individuals are removed and a new initial
+        population population is generated.
+        """
+        self._feas.restart()
```
N-Wouda, line 365: So here `self._feas = SubPopulation(...)`.

pyvrp/read.py
```diff
@@ -13,6 +13,10 @@
 _INT_MAX = np.iinfo(np.int32).max
 
 
+def round_nearest(vals: np.ndarray):
```
N-Wouda, line 16: Docs: can we add this to the `read()` docstring?

pyvrp/Population.py
```diff
@@ -351,6 +357,17 @@ def select(self) -> Tuple[Individual, Individual]:
 
         return first, second
 
+    def restart(self):
```
N-Wouda, line 360: Can we add a test that after calling `restart()`, the population is of size `min_pop_size` and consists of new individuals?

pyvrp/GeneticAlgorithm.py
```diff
@@ -25,6 +25,7 @@ class GeneticAlgorithmParams:
     repair_probability: float = 0.80
     collect_statistics: bool = False
     should_intensify: bool = True
+    nb_iter_no_improvement: int = 10000
```
N-Wouda, line 28: Parameter value we'll need to tune. It was, I think, 20K for DIMACS? (because I'm getting old now, can we write this as `10_000` rather than `10000`?)

leonlan: 20K is for DIMACS yes. The 10K is what we used in the competition. Since we'll benchmark on GH1000 instances anyway I think putting it on 20K is a better choice.

pyvrp/read.py
leonlan: I have removed the bounds from this PR but I'll add your comments to my other branch for this.

pyvrp/Population.py
leonlan: Can you check the new test?

N-Wouda: This works for me. Could perhaps be made a bit nicer if we add a `__contains__` dunder to `SubPopulation`, but that's not really important right now.

N-Wouda (approved):

pyvrp/Population.py
```diff
@@ -351,6 +351,17 @@ def select(self) -> Tuple[Individual, Individual]:
...
```
wouterkool, line 354: We could use restart() also in `__init__` to avoid duplicate code but since we're seperating initializing the population anyway in #88 I propose to address it there (i.e. have both `__init__` and `restart` use `self.initialize()`.

wouterkool (approved):
We could use a test for the round function but that's a todo anyway

PR #91, opened 2023-02-17
Performance tweaks

N-Wouda:
Mostly on the C++/LS side. This PR:
- [x] Implements a few performance improvements identified via profiling;
- [x] Adds a compiler flag to shortcut TW computations. This effectively enables one to compile a CVRP solver from the current codebase.


N-Wouda:
Current main on all HG instances:
```
      Avg. objective: 335895
     Avg. iterations: 75019
   Avg. run-time (s): 3600.03
        Total not OK: 0
```
This PR (272c0a53577acfee8387d8bd3d3eaed8114a466f):
```
TODO
```

pyvrp/cpp/educate/Route.h
```diff
@@ -111,7 +111,14 @@ bool Route::isFeasible() const
 
 bool Route::hasExcessCapacity() const { return !isLoadFeasible_; }
 
-bool Route::hasTimeWarp() const { return isTimeWarpFeasible_; }
```
N-Wouda, line 114: This actually had a bug: should have been `!isTimeWarpFeasible_`. Fixed with the proposed changes.

N-Wouda:
I need to check which places actually matter for the compiler flag, and which are not doing much. The latter we can revert.

leonlan (approved):
Looks good to me. I've merged this branch into my CVRP branch and will run a benchmark overnight.

meson.build
```diff
@@ -49,6 +49,12 @@ endif
 
 assert(pybind11.found(), 'Could not find pybind11!')
 
+if get_option('problem') == 'cvrp'
+    # CVRP does not have time windows, so we set a flag that compiles time 
+    # window stuff out of the extension modules.
+    add_project_arguments('-DVRP_NO_TIME_WINDOWS', language: 'cpp')
+endif
+
```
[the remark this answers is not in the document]
N-Wouda: Yes, I'll have a look at this. My first look will be at "why didn't gcc give me this overview" :).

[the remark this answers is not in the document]
N-Wouda: This is a clang-specific warning, so gcc doesn't have it (yet). I'm disabling it when compiling for CVRP, since those unused fields are all TW-related and it's OK they're still there.

N-Wouda:
@wouterkool can I merge this? :-)

pyvrp/cpp/educate/Route.h
wouterkool: Nice catch! This makes`Route::isFeasible()` incorrect as well, which is used in prechecks in exchange and 2OPT, so, in the old situation:
* We skip potentially improving moves if there is a timewarp but `deltaCost >= 0`
* We waste some time evaluating non-improving moves. I wonder if it changes performance?
For the second point the bugfix is a guaranteed improvement, but for the first point it is not as it may be better to skip these moves anyway if they rarely yield improvement. Therefore this bug may actually have been a feature, so it may make sense to test this bugfix independently of the other (guaranteed) performance improvements.

pyvrp/educate/tests/test_MoveTwoClientsReversed.py
```diff
@@ -33,13 +33,14 @@ def test_single_route_OkSmall():
     assert_equal(individual.num_routes(), 1)
     assert_(individual.cost() < copy.cost())
 
-    # (2, 3) was inserted after the depot as (3, 2)
-    assert_equal(individual.get_routes(), [[3, 2, 1, 4], [], []])
+    # (2, 3) was inserted after 1 as 1 -> 3 -> 2 -> 4. Then (1, 3) got inserted
+    # after 2 as 2 -> 3 -> 1 -> 4.
+    assert_equal(individual.get_routes(), [[2, 3, 1, 4], [], []])
```
wouterkool, line 38: Was this changed as a result of the bugfix? Do we still need to have the tests below as we test for this exact solution?

wouterkool (approved):
Given the `ifdefs` in `TimeWindowSegment::merge` I assume the compiler won't optimize it by itself? Then would it make sense to also add the `ifdef` in `Route::setupRouteTimeWindows` as this still runs a loop and sets `Node::twAfter`?

N-Wouda:
> Given the `ifdefs` in `TimeWindowSegment::merge` I assume the compiler won't optimize it by itself? Then would it make sense to also add the `ifdef` in `Route::setupRouteTimeWindows` as this still runs a loop and sets `Node::twAfter`?

Yes, but: when I profiled this last on the largest X instance, that particular function took less than 0.3% of the total time, so I figured it wasn't worth tinkering with any further (for now).

pyvrp/educate/tests/test_MoveTwoClientsReversed.py
N-Wouda: Yes to the first part. What do you mean by the second question?

pyvrp/cpp/educate/Route.h
N-Wouda: I think I introduced this bug last month when I started caching the route feasibilities. It wasn't there during Euro/NeurIPS, or DIMACS.

A very significant part of the performance improvement came from fixing the bug (like 80% or so, but I do not have the exact numbers). The old (wrong) behaviour of `Route::isFeasible()` might have helped incidentally, but almost 2x more iterations seems to more than make up for it!

PR #92, opened 2023-02-17
Python neighbourhood

### LINKED ISSUE #64: LocalSearch should take an (optional) neighbourhood structure
Right now, the LS object creates its own granular neighbourhood structure. In many cases, another neighbourhood structure might well perform better, and it would help a lot if we can pass our own structure into the LS object.

We should add another constructor to LS that takes such a structure (e.g., a `std::vector<std::vector<int>>` - a vector of near clients for each client, like the current neighbourhood structure maintained by LS).

wouterkool:
This pull requests implements computing the neighbourhood in Python exactly as previously on the C++ side, and allows to set (and get) the neighbourhood from Python which can be used to implement growing neighbourhoods later. This fixes #64.


pyvrp/cpp/educate/LocalSearch.cpp
```diff
@@ -20,6 +20,12 @@ void LocalSearch::search(Individual &indiv)
     if (nodeOps.empty())
         throw std::logic_error("No known node operators.");
 
+    int neighbourhood_size = 0;
+    for (auto const client : orderNodes)
+        neighbourhood_size += neighbours[client].size();
+    if (neighbourhood_size == 0)
+        throw std::runtime_error("Granular neighbourhood is empty.");
```
N-Wouda, line 27: I'm thinking of what the right approach here is. Do we only want to raise when the entire neighbourhood is empty? It already feels somewhat weird to me if any neighbourhood is empty, for any client. Should we raise in that case? Or are there valid use cases where one might have a zero-size neighbourhood for a particular client?

pyvrp/cpp/educate/LocalSearch.h
```diff
@@ -15,25 +15,17 @@
...
+        : postProcessPathLength(postProcessPathLength)
     {
-        if (nbGranular == 0)
-            throw std::invalid_argument("Expected nbGranular > 0.");
+        
     }
 };
 
+// TODO define within LocalSearch
+using Neighbours = std::vector<std::vector<int>>;
```
N-Wouda, line 28: Is this TODO still relevant?

pyvrp/educate/tests/test_neighborhood.py
```diff
@@ -0,0 +1,227 @@
...
+        nb_granular,
+        symmetric_proximity,
+        symmetric_neighbours,
+    )
+    neighbours = compute_neighbours(data, params)
+
+    assert_equal(len(neighbours), data.num_clients + 1)
+    assert_equal(len(neighbours[0]), 0)
+
+    assert neighbours[idx_check] == expected_neighbours_check
```
N-Wouda, line 158: In this file: can we use `assert_()`, rather than a naked `assert`?

pyvrp/educate/neighbourhood.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # Mask self from proximities by adding a large constant
+    # and add the index with a very small weight as tie-breaker (to mimic
+    # original c++ version)
+    n = len(proximity)
+    k = min(params.nb_granular, n - 1)
+    rng = np.arange(n)
+    mask = np.eye(n, dtype=bool)
+    masked_proximity = np.where(mask, 1e9, proximity + 1e-6 * rng[None, :])
+    idx_topk = np.argpartition(masked_proximity, k, axis=-1)[:, :k]
```
N-Wouda, line 106: This took me a second to reason through, but I think I get it. I think this is perhaps a bit clearer:
```python
...

jitter = 1e-6 * np.arange(n)  # add a bit of jitter to break ties
proximity = proximity + jitter[None, :]
np.fill_diagonal(proximity, np.inf)  # exclude self from neighbourhood
idx_topk = np.argpartition(proximity, k, axis=-1)[:, :k]
```

N-Wouda:
I think the CI can be satisfied by a `from __future__ imports annotations` at the top of the offending file.

N-Wouda:
Overall this looks OK to me.

pyvrp/educate/neighbourhood.py
wouterkool: I tend to prefer declarative over imperative but I agree with the clarity. I do define the `rng` separately as we reuse it but hope you're ok with it!

pyvrp/cpp/educate/LocalSearch.cpp
wouterkool: It's a bit subtle, I can imagine situations where e.g. you want to focus the search on a subset of the solution, so then, yes it's valid so let's use the most general option.

pyvrp/educate/neighbourhood.py
N-Wouda: OK, but `rng` to me reads like 'random number generator', which it is not. Does it have anything to do with randomness?

N-Wouda (approved):

N-Wouda:
We can merge this once the CI completes.

PR #93, opened 2023-02-18
Remove subpath enumeration

### LINKED ISSUE #64: LocalSearch should take an (optional) neighbourhood structure

### LINKED ISSUE #90: There's something wrong with subpath enumeration
I noticed this writing tests in #86: sometimes subpath enumeration allows a solution to degrade. That should not be possible, so something is wrong. This ties into #65, and can probably be investigated while we tackle that issue.

For now, I have disabled this operator in #86.

N-Wouda:
Closes #64, closes #90.


N-Wouda:
Current main (w/o subpath enumeration):
```
      Avg. objective: 335895
     Avg. iterations: 75019
   Avg. run-time (s): 3600.03
        Total not OK: 0
```
so it does not seem to add much. Since it's broken, we might be better off just getting rid of it entirely.

PR #95, opened 2023-02-19
A few more test cases

N-Wouda:


N-Wouda:
I'm merging this without review because all this PR does is introduce two new tests.

PR #97, opened 2023-02-19
Move penalty updates/feasibility registration to PenaltyManager

N-Wouda:
This moves feasibility registration fully into the PenaltyManager class, and has that class update the penalties whenever sufficiently many results have been registered. That further concentrates all penalty-related shenanigans into the PenaltyManager.

This slightly modifies the meaning of the parameter controlling the updating. I have increased it a little (from 47 to 50) to reflect this, but will also run a benchmark to be sure we are not losing performance here.

Benchmark (ff55f03225eac59e6fa04187c819a30201a3f0d8):
```
      Avg. objective: 336032
     Avg. iterations: 75038
   Avg. run-time (s): 3600.03
        Total not OK: 0
```
which is largely unchanged from current main.


wouterkool (approved):
I understand the logic was scattered between python/c++ so this is better, however wouldn't it be better to have the PenaltyManager on the Python side (this does not seem performance critical) and just pass a Penalties (or more generally, Objective/ObjectiveParams or something like that) object to the local search?

leonlan (approved):

N-Wouda:
@wouterkool yes, I think splitting off the penalty values from the PenaltyManager class makes sense to do sometime soon-ish. I'll make an issue for it.

PR #106, opened 2023-02-20
Finish PyPI deployment steps

### LINKED ISSUE #63: Make wheel builds work for Windows, OSX
See also #58. There, we set up cibuildwheel to precompile wheels for distribution. That works well for Linux and 64-bit systems (as expected), but does not yet support OSX or Windows. This issue tracks what needs to be done.

### LINKED ISSUE #67: PyPI: steps
To deploy the package, we need to do a few things:

- Figure out the naming issue. Maybe revisit #4, if we cannot get pyvrp on PyPI?
- Complete building wheels in the CI: #63.
- Finish the CD workflow to include uploading to PyPI. This is part of a bigger task that's basically "give some love to the CI situation".
- Strip the resulting wheels from debug data, etc. Make sure the wheels are properly packaged, and that the source releases contain everything to reproduce them.

This issue serves to track all that.

N-Wouda:
This PR closes #63 and closes #67:
- [x] We have the `PyVRP` name.
- [x] The CI can build binaries on:
  - [x] Windows;
  - [x] Linux;
  - [x] OS X.
- [x] The CI can build binaries for Python 3.8 - 3.11.
- [x] The workflows are somewhat streamlined.
- [x] The wheels are properly packaged.
- [x] The PyPI upload works correctly, using the right credentials.

I also looked into building on 32-bit systems, but decided against that because (a) it's a bit of a pain to set up all the tooling for in the CI, and (b) scipy and the like also do not provide binaries on those platforms.


N-Wouda:
@leonlan @wouterkool can you have a look at this? I've moved our build system over to Python, and use the `build_extensions.py` script everywhere now. This was needed because a shell script doesn't work under Windows, and I didn't want to maintain both a Unix and a Windows version of the same thing.

leonlan (approved):

build_extensions.py
```diff
@@ -0,0 +1,118 @@
...
+    check_call(["rm", "-rf", str(build_dir)])
+
+    for extension in install_dir.rglob("*.so"):
+        extension.unlink()
+
+    for extension in install_dir.rglob("*.pyd"):
+        extension.unlink()
+
+
+def regen_stubs(install_dir: pathlib.Path):
```
wouterkool, line 61: Maybe just call it `regenerate_stubs`? Also, it feels a bit weird to have this in here as it changes files under version control, which I would not per se expect in a build script, but I think it is not that important.

pyproject.toml
```diff
@@ -33,24 +35,30 @@ classifiers = [
...
+
 [tool.poetry.urls]
-"Tracker" = "https://github.com/N-Wouda/pyvrp/issues"
+"Tracker" = "https://github.com/N-Wouda/PyVRP/issues"
+
 
 [tool.poetry.dependencies]
 python = "^3.8,<4.0"
 numpy = ">=1.15.2"
 matplotlib = ">=2.2.0"
```
wouterkool, line 46: In general related to dependencies, given that dependencies for benchmark, dev and examples are separated, does it make sense to have matplotlib a required dependency or should this be part of examples (for notebooks) and dev (for tests) groups?

docs/source/setup/benchmarking.rst
```diff
@@ -0,0 +1,33 @@
+Benchmarking
+============
+
+To run a benchmark, one can use the command line interface ``PyVRP`` provides.
+After installation, it is available as ``pyvrp`` from the command line.
```
wouterkool, line 5: Does this hold both for building from sources and installing through PyPI? In the first case, would it be helpful to mention Poetry environment must be activated or you must run `poetry run pyvrp`? Finally, this requires `benchmark` group dependencies, which are optional, will they be installed by default via PyPI?

docs/source/setup/benchmarking.rst
```diff
@@ -0,0 +1,33 @@
...
+
+   pyvrp --help
+
+To benchmark ``pyvrp``, we use the following `CRVPLIB <http://vrp.atd-lab.inf.puc-rio.br/index.php/en/>`_ instances:
+
+* For the VRPTW, we use `the Gehring and Homberger instances <http://vrp.atd-lab.inf.puc-rio.br/media/com_vrp/instances/Vrp-Set-HG.tgz>`_, particularly those with 1000 customers.
+
+* For the CVRP, we use `the X-instances of Uchoa <http://vrp.atd-lab.inf.puc-rio.br/media/com_vrp/instances/Vrp-Set-X.tgz>`_.
+
+To get both instance sets quickly, the following script can be used to download and unpack them:
```
wouterkool, line 18: Don't think this works on Windows, maybe add a comment?

wouterkool (approved):

.github/workflows/CI.yml
```diff
@@ -73,7 +73,7 @@ jobs:
         with:
           version: ${{ matrix.compiler-version }}
       - name: Build pyvrp
-        run: poetry run scripts/install.sh
+        run: poetry run python build_extensions.py --build_type debug
```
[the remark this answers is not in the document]
N-Wouda: Because `poetry build` does not pass additional arguments to the script.

README.md
```diff
@@ -30,26 +31,26 @@ poetry run pytest
 ```
 If all tests pass without errors, you have a working installation of the codebase.
 
-From this point onwards, recompilation of the C++ extensions can best be done using the `scripts/install.sh` script.
+From this point onwards, recompilation of the C++ extensions can best be done using the `build_extensions.py` script.
 It can be called as
 ```shell
-poetry run scripts/install.sh <buildtype>
+poetry run python build_extensions.py
```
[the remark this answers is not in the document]
N-Wouda: Yeah but that'll only work if you (only) want the default build, which is a release build. That's probably not the case when e.g. developing locally, so I think it's good to forget about `poetry build` altogether.

pyproject.toml
N-Wouda: Since we have code that directly depends on matplotlib being installed, we should probably consider it a core dependency (e.g., when a user wants to use `pyvrp.plotting` they'll need matplotlib).

N-Wouda: Since we provide `pyvrp` as a CLI command, we should perhaps consider the benchmark group as core dependencies as well.

docs/source/setup/benchmarking.rst
N-Wouda: For developing locally, this needs to run within the poetry environment, but we'll make that clear in the contributing/developing sections. The command is provided when the package is installed, and that requires the benchmark group which I will add to the core dependencies now.

pyproject.toml
wouterkool: Agree.

Issue #59, opened 2023-02-02
Persist Result or Statistics data to filesystem after a run


N-Wouda:
The Euro/NeurIPS repository had a `Statistics::toCsv` method to persist Statistics data as a CSV file. That method has not (yet) been reimplemented. 

We do still have an `Individual::toFile` method that persists a solution, however. This too persists some sort of state to yet another file.

Can we streamline writing to/from files in some way? Perhaps to/from a JSON object, and then the whole Result object, not just the Individual or the Statistics?

wouterkool:
Maybe we could do it somewhat like PyTorch:
* Have `get_state_dict`/`load_state_dict` methods that return a dictionary/load state from a dictionary
* Have `pyvrp.save(dict, filename)` and `pyvrp.load(filename)` to save to file, either use proprietary format (e.g. `pkl`) or let user specify format (e.g. `json`)
* We can implement this function on individual, population, but also on genetic algorithm itself so we could also use it for checkpointing the state of an algorithm to resume later (e.g. for use on spot compute instances)

N-Wouda:
The biggest issue to persisting an Individual is that it takes a `PenaltyManager` instance, which we should really get rid of. @wouterkool your suggestion of passing the relevant penalties to `Individual::cost()` is probably what we should go for.

N-Wouda:
>We do still have an `Individual::toFile` method that persists a solution, however. This too persists some sort of state to yet another file.

I have removed this in #58, and instead print that information when calling `str(individual)` from Python. So if one wants to write a individual to a file, one could (for now) do
```python
with open(file, 'w') as fh:
    fh.write(str(individual))
```

N-Wouda:
Statistics has for some time now supported writing to a CSV, and reading a CSV back into a Statistics object. So that should be OK in terms of human-readable output.

Serialising a Result object is a bit harder, because of the underlying `Solution`. I think we should go towards pickling the whole thing, because [pybind has great support for that](https://pybind11.readthedocs.io/en/latest/advanced/classes.html#pickling-support). If we make `Solution` pickleable we should be done in terms of serialisation. The other objects in Result and Statistics should already be pickleable out of the box (but we should write tests that confirm this in `test_Result.py` and `test_Solution.py`).

N-Wouda:
I'm assigning myself and will work on this tomorrow. Should be pretty easy to do.

[...]

PR #110, opened 2023-02-21
Serialise statistics

N-Wouda:
This PR:

- [x] Is part of #59.

### DIFF
--- a/pyvrp/Population.py
+++ b/pyvrp/Population.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from statistics import fmean
 from typing import Callable, Iterator, List, NamedTuple, Tuple
 
 import numpy as np
@@ -189,7 +190,7 @@ class SubPopulation:
         """
         _, _, prox = self[individual_idx]
         closest = prox[: self._params.nb_close]
-        return np.mean([div for _, div in closest]) if closest else 0.0
+        return fmean(div for _, div in closest) if closest else 0.0
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


pyvrp/cli.py
```diff
@@ -87,6 +88,9 @@ def solve(
     max_iterations
         Maximum number of iterations for solving. Either ``max_runtime`` or
         ``max_iterations`` must be specified.
+    stats_dir, optional
+        The directory to write runtime statistics to. Enables statistics
+        collection when passed.
```
leonlan, line 93: Suggestion: could we also add a `solutions_dir` to save best solutions in VRPLIB format?

leonlan (approved):

wouterkool (approved):
I think the setup is good in general, but I'm strongly in favor of an option to collect statistics only every x iterations, to keep file sizes managable (when we do many runs of 100K+ iterations). We therefore can consider to add an explicit column 'iteration' to the CSV.

pyvrp/cli.py
N-Wouda: Yeah, I'll add that.

N-Wouda:
> I think the setup is good in general, but I'm strongly in favor of an option to collect statistics only every x iterations, to keep file sizes managable (when we do many runs of 100K+ iterations). We therefore can consider to add an explicit column 'iteration' to the CSV.

I experimented with this a bit, and by my estimates a full stats run on the HG instances will result in around 1GB data (15-20MB per instance). When zipped or read into a dataframe that significantly reduces, since the csv file has a lot of overhead.

Issue #112, opened 2023-02-22
Automatic testing of (example) notebooks


wouterkool:
Currently the notebooks are not tested. Once we make updates to the library, we should make sure example notebooks still work as expected.

Simplest solution: add `jupyter nbconver --execute {notebook}` for all notebooks as a step to the CI.

If we need more advanced functionality, we could use https://github.com/treebeardtech/nbmake.

PR #114, opened 2023-02-22
Fix example notebook

wouterkool:
This PR:

- [x] Fixes the example notebook where compute_neighbours was missing (no issue)
- [ ] Adds and passes relevant tests. (TODO see #112)


N-Wouda:
Good that it works again! Can we do `fig.tight_layout()` for the `plot_result` calls as well? The figure labels overlap a bit, and that looks somewhat sloppy.

(this call used to be in `plot_result` already, but that didn't play nice with the plot tests so it was removed a few PRs ago)

N-Wouda (approved):

PR #115, opened 2023-02-22
Add pyvrp.educate.neighbourhood to the docs, add neighbourhood constructor to local search

### LINKED ISSUE #105: Document pyvrp.educate.neighbourhood

N-Wouda:
This PR:

- [x] Closes #105.

<details>

**Notes**:

</details>


pyvrp/cpp/educate/LocalSearch_bindings.cpp
```diff
@@ -9,10 +9,14 @@ namespace py = pybind11;
 PYBIND11_MODULE(LocalSearch, m)
 {
     py::class_<LocalSearch>(m, "LocalSearch")
-        .def(py::init<ProblemData &, PenaltyManager &, XorShift128 &>(),
+        .def(py::init<ProblemData &,
+                      PenaltyManager &, 
+                      XorShift128 &,
+                      std::vector<std::vector<int>>>(),
```
wouterkool, line 15: Not sure, but can't you use the `LocalSearch::Neighbours` type here?

pyvrp/educate/tests/test_LocalSearch.py
```diff
@@ -61,7 +83,10 @@ def test_local_search_set_get_neighbours(
     seed = 42
     rng = XorShift128(seed=seed)
     pen_manager = PenaltyManager(data.vehicle_capacity)
-    ls = LocalSearch(data, pen_manager, rng)
+
+    prev_neighbours = [[client] for client in range(data.num_clients)]
```
wouterkool, line 87: This makes me think: shouldn't we validate that client can't be a neighbour of client itself? I think u = v gives undefined behavior in the local search. (I understand that will not be the case here if you prepend the depot afterwards.)

wouterkool (approved):
LGTM

pyvrp/cpp/educate/LocalSearch_bindings.cpp
N-Wouda:
```
allocator<int> >, std::allocator<std::vector<int, std::allocator<int> > > >’ is private within this context
   12 |         .def(py::init<ProblemData &,
      |                  ^~~~~~~~~~~~~~~~~~~
   13 |                       PenaltyManager &,
      |                       ~~~~~~~~~~~~~~~~~
   14 |                       XorShift128 &,
      |                       ~~~~~~~~~~~~~~
   15 |                       LocalSearch::Neighbours>(),
      |                       ~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../pyvrp/cpp/educate/LocalSearch_bindings.cpp:1:
../pyvrp/cpp/educate/LocalSearch.h:20:11: note: declared private here
   20 |     using Neighbours = std::vector<std::vector<int>>;
      |           ^~~~~~~~~~
ninja: build stopped: subcommand failed.
```
So I don't think so? We could make it public, but I'm not sure if that helps.

pyvrp/educate/tests/test_LocalSearch.py
N-Wouda: I've added some checks for this (depot and the client itself).

PR #119, opened 2023-02-23
Store only best feasible solutions

leonlan:
This PR fixes a bug where infeasible best solutions were collected.

### DIFF
--- a/pyvrp/GeneticAlgorithm.py
+++ b/pyvrp/GeneticAlgorithm.py
@@ -162,8 +162,9 @@ class GeneticAlgorithm:
                         self._best = individual
 
                     self._pop.add(individual)
-
-                    # We already know the individual is feasible, so load and
-                    # time constraints are both satisfied.
-                    self._pm.register_load_feasible(True)
-                    self._pm.register_time_feasible(True)
+                    self._pm.register_load_feasible(
+                        not individual.has_excess_capacity()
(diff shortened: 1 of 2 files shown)


pyvrp/GeneticAlgorithm.py
```diff
@@ -155,7 +155,10 @@ def _educate(self, individual: Individual):
                     ):
                         self._ls.intensify(individual)
 
-                    if individual.cost() < self._best.cost():
+                    if (
+                        individual.is_feasible()
+                        and individual.cost() < self._best.cost()
+                    ):
                         self._best = individual
```
N-Wouda, line 162: It's not needed here because this already happens inside `if individual.is_feasible()`

N-Wouda (commented):
Thanks! Can you update the test in `test_GeneticAlgorithm.py` to explicitly check that the returned solutions are feasible?

pyvrp/GeneticAlgorithm.py
leonlan: This might still be needed because an intensified solution can become infeasible I think.

N-Wouda: Yeah that's fair. But then the register calls below this are also wrong. Can you update those?

N-Wouda (approved):

PR #120, opened 2023-02-23
SubPopulation in C++

N-Wouda:
This PR:
- [x] Moves the current SubPopulation to C++, for a substantial performance boost.
- [x] Removes the publicness of the subpopulations. Instead, we can iterate over the full population now.
- [x] Copies a bunch of tests from #88.  

### Benchmarks

Main (0d102e4e242e26177ea8807fa74516acecb060f8) [job ID 28700256, seeds 1-10]:
```
 Avg. objective: 334303.6 (min: 334093, max: 334668)
Avg. iterations: 123359.6 (min:  99403, max: 143166)
```

This branch (2c54391cdc468fb82e40fa469d0d1aaf1220cfcc) [job ID 28700467, seeds 1-10]:
```
 Avg. objective: 334094.9 (min: 338840, max: 334300)
Avg. iterations: 152968.2 (min: 139171, max: 174017)
```

Old benchmark runs:

<details>

Benchmark (7e3bb3613c232efba0067933e49b900e57dbef58) [job ID 28700054, seed 0]:
```
      Avg. objective: 334922
     Avg. iterations: 208683
   Avg. run-time (s): 3600.01
        Total not OK: 0
```
 
Benchmark (5a966aa3099ded9e835befc86da108ceca9f4349) [job ID 28696529, seed 1]:
```
      Avg. objective: 335776
     Avg. iterations: 209256
   Avg. run-time (s): 3600.01
        Total not OK: 0
```

Benchmark (5a966aa3099ded9e835befc86da108ceca9f4349) [job ID 28699377, seed 2]:
```
      Avg. objective: 334811
     Avg. iterations: 190627
   Avg. run-time (s): 3600.01
        Total not OK: 0
```

---

Main (7705accf0bde2efab8714584a91a5cffa412d5f8) [job ID 28696067, seed 1]:
```
      Avg. objective: 334963
     Avg. iterations: 146988
   Avg. run-time (s): 3600.01
        Total not OK: 0
```

</details>


N-Wouda:
I've mostly focused on getting this to work. Tomorrow I'm going to run the old and new subpopulations side by side, and then make sure everything's really as it should be.

N-Wouda:
If we compare the 'old' SubPopulation (the one currently on main) with the new SubPopulation in this branch, the following test passes:

```python
def test_are_same():
    data = read("data/RC208.txt", "solomon", "dimacs")
    div_op = broken_pairs_distance
    pm = PenaltyManager(data.num_vehicles)
    rng = XorShift128(seed=42)
    params = PopulationParams(min_pop_size=2, generation_size=10)

    old = SubPopulationOld(data, div_op, params)
    new = SubPopulation(data, div_op, params)

    assert_equal(len(old), len(new))
    assert_equal(len(old), 0)

    indiv1 = Individual(data, pm, rng)

    assert_(not indiv1.is_feasible())

    old.add(indiv1)
    new.add(indiv1)

    assert_equal(len(old), len(new))
    assert_equal(len(old), 1)

    old_item1 = old[0]
    new_item1 = new[0]

    assert_equal(old_item1.individual, new_item1.individual)
    assert_equal(old_item1.individual, indiv1)

    assert_equal(old_item1.fitness, new_item1.fitness)

    indiv2 = Individual(data, pm, rng)
    assert_(not indiv2.is_feasible())

    old.add(indiv2)
    new.add(indiv2)

    assert_equal(len(old), len(new))
    assert_equal(len(old), 2)

    assert_(indiv1.cost() > indiv2.cost())

    assert_equal(old[0].individual, indiv1)
    assert_equal(new[0].individual, indiv1)

    assert_equal(old[1].individual, indiv2)
    assert_equal(new[1].individual, indiv2)

    assert_allclose(old[0].fitness, new[0].fitness)
    assert_allclose(old[1].fitness, new[1].fitness)
    assert_(old[0].fitness > 0.0)

    for step in range(10):
        indiv = Individual(data, pm, rng)
        assert_(not indiv.is_feasible())

        assert_equal(len(old), 2 + step)
        assert_equal(len(new), 2 + step)

        old.add(indiv)
        new.add(indiv)

        assert_allclose(old[-1].fitness, new[-1].fitness)

    for idx, (item1, item2) in enumerate(zip(old, new)):
        assert_equal(item1.individual, item2.individual)
        assert_allclose(item1.fitness, item2.fitness)
        assert_allclose(old.avg_distance_closest(idx), new.avg_distance_closest(idx))

    indiv = Individual(data, pm, rng)
    assert_(not indiv.is_feasible())

    old.add(indiv)
    new.add(indiv)

    # Previous insert should have caused a purge
    assert_equal(len(old), params.min_pop_size)
    assert_equal(len(new), params.min_pop_size)

    for _ in range(1_000):
        indiv = Individual(data, pm, rng)

        print(_, len(old), len(new))

        old.add(indiv)
        new.add(indiv)

        assert_equal(len(old), len(new))

        for idx, (item1, item2) in enumerate(zip(old, new)):
            assert_equal(item1.individual, item2.individual)
            assert_allclose(
                old.avg_distance_closest(idx),
                new.avg_distance_closest(idx)
            )

            old_prox = [div for _, div in item1.proximity]
            new_prox = [div for div, _ in item2._proximity]
            assert_allclose(old_prox, new_prox)
```

N-Wouda:
@wouterkool @leonlan can you have a look at this? Mostly for structure and stuff; I'm running a benchmark right now to make sure this is actually an improvement (it should be, but you never quite know).

pyvrp/cpp/SubPopulation.cpp
```diff
@@ -0,0 +1,161 @@
...
+        byCost.end(),
+        [&](size_t a, size_t b)
+        { return items[a].individual->cost() < items[b].individual->cost(); });
+
+    std::vector<std::pair<double, size_t>> diversity;
+    for (size_t rank = 0; rank != size(); rank++)
+    {
+        auto const dist = avgDistanceClosest(byCost[rank]);
+        diversity.emplace_back(dist, rank);
+    }
```
leonlan, line 131: Suggestion: can we rename `rank` here to `costRank`?

pyvrp/cpp/SubPopulation.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+double SubPopulation::avgDistanceClosest(size_t idx) const
+{
+    auto const &item = items[idx];
+    auto const maxSize = std::min(item.proximity.size(), params.nbClose);
+
+    if (maxSize == 0)
+        return 0.0;  // TODO 0.0 or 1.0?
```
leonlan, line 153: I think 0.0 is fine, right? Either an item has no proximity, so it will always have the best diversity rank. Or when `nbClose=0`, then all individuals have the same diversity.

leonlan (approved):

pyvrp/cpp/SubPopulation_bindings.cpp
```diff
@@ -0,0 +1,67 @@
...
+             py::arg("ub_diversity") = 0.5)
+        .def_readwrite("min_pop_size", &PopulationParams::minPopSize)
+        .def_readwrite("generation_size", &PopulationParams::generationSize)
+        .def_property_readonly("max_pop_size", &PopulationParams::maxPopSize)
+        .def_readwrite("nb_elite", &PopulationParams::nbElite)
+        .def_readwrite("nb_close", &PopulationParams::nbClose)
+        .def_readwrite("lb_diversity", &PopulationParams::lbDiversity)
+        .def_readwrite("ub_diversity", &PopulationParams::ubDiversity);
+
+    py::class_<SubPopulation::Item>(m, "SubPopulationItem")
```
wouterkool, line 27: Can we not expose the SubPopulationItem through the bindings? I think this is an implementation detail. I suggest `SubPopulation::__getitem__` returns an Individual and `SubPopulation::__iter__` iterates over Individuals. Preferably also on the c++ code but at least on the bindings. To facility the internal tests and sampling the binary tournament, you could make a `get_diversity(idx)` method (not pretty to use idx but same `SubPopulation::avg_distance_closest`. For the test, you can expose an `_get_proximity` method if you really need to. This keeps implementation details abstracted from the Python interface.

wouterkool (changes requested):
* I think only reason SubPopulation has a ProblemData argument is for the diversity op. I suggest we change diversity op to bind the data and just take two individuals as arguments, then we can remove the data dependency and make the implementation simpler/more general.
* Can we not expose the SubPopulationItem to Python?


In Population you can use:
<details>
```

def get_binary_tournament(self) -> Individual:

        def select():
            num_feas = len(self._feas)
            idx = self._rand_int_func(len(self))

            if idx < num_feas:
                return self._feas[idx], self._feas.get_biased_fitness(idx)

            return self._infeas[
                idx - num_feas
            ], self._infeas.get_biased_fitness(idx - num_feas)

        indiv1, fitness1 = select()
        indiv2, fitness2 = select()

        if fitness1 < fitness2:
            return indiv1

        return indiv2
```
</details>

pyvrp/cpp/SubPopulation_bindings.cpp
N-Wouda: I think the entirety of `SubPopulation` is an implementation detail. The only reason it might not be is that we have `feasible_subpopulation` and `infeasible_subpopulation` properties on the `Population`. But those feel like a step in the wrong direction (why does a user need to know the Population has subpopulations by feasibility?), and should not be part of the public interface.

> To facility the internal tests and sampling the binary tournament, you could make a `get_diversity(idx)` method (not pretty to use idx but same `SubPopulation::avg_distance_closest`.

I think this is a mistake, for the same reason having `avgDistanceClosest` with an index was a mistake. That method should be a property of the item, not a standalone thing on `SubPopulation`.

N-Wouda:
>I think only reason SubPopulation has a ProblemData argument is for the diversity op. I suggest we change diversity op to bind the data and just take two individuals as arguments, then we can remove the data dependency and make the implementation simpler/more general.

Yes, but that means any C++ call to that diversity operator is going to route through Python first, before returning to C++ (at least, if we bind using Python's `partial`). That's somewhat slow, but it might not matter. I'll experiment with this a bit.

N-Wouda:
@wouterkool:

- I've removed the dependency on `ProblemData` from the `SubPopulation`. It does not really matter for performance that we pass a Python function, and I agree this is a lot simpler on the C++ side.
- I've moved `avgDistanceClosest` to the `Item`, thus removing the only function on `SubPopulation` that required an index argument.

I am not sure about removing `Item` from the picture altogether, and using getters instead. I personally prefer the thin wrapper of `Item` because it keeps the SubPopulation interface rather small.

N-Wouda:
Using Python lambda:

image

After revert:

image

These numbers are from just one run, but I've checked this over several, and the percentages are fairly indicative: not having the Python lambda is much faster.

wouterkool:
> @wouterkool:
> 
> * I've tried removing the dependency on `ProblemData` from the `SubPopulation`, and using a Python lambda instead. That comes at a significant performance hit: it's about 5x slower than keeping everything in C++. On a 30s run it now takes about 4.5% of the time, vs 0.6-0.7% before. I have reverted this change, since that C++ -> Python -> C++ round trip per function call is clearly not insignificant.
> * I've moved `avgDistanceClosest` to the `Item`, thus removing the only function on `SubPopulation` that required an index argument.
> 
> I am not sure about removing `Item` from the picture altogether, and using getters instead. I personally prefer the thin wrapper of `Item` because it keeps the SubPopulation interface rather small.

* I am surprised it takes that long, haven't looked in detail but do we not recompute diversity too often? But good to know this roundtrip is slow and fine to keep it then for performance.
* I agree not having the function with index as argument is an improvement and I don't mind keeping the Item in the bindings as long as we can keep it internal to the population. That means either not exposing the subpopulations or making the subpopulation `__index__` and `__iter__` return individuals and bind an explicit `_get_item` to get the item to be used only within `Population`.

N-Wouda:
I've removed the subpopulation properties from Population, and updated the tests to use public interfaces where possible. Only for the binary tournament that was impossible: we really need the fitness attribute there. The Statistics object also uses the (private) subpopulations directly, but I think that is OK: it needs to collect that stuff.

This effectively separates the internal implementation with `SubPopulation` and `SubPopulation::Item` from the external, Python-side `Population`. The latter now only knows and cares about `Individual`.

N-Wouda:
@leonlan can you also review this again? It's changed quite a bit since your last approval.

pyvrp/cpp/Individual.h
```diff
@@ -110,4 +112,21 @@ class Individual
...
+{
+    std::size_t operator()(Individual const &individual) const
+    {
+        size_t res = 17;
+        res = res * 31 + std::hash<size_t>()(individual.nbRoutes);
+        res = res * 31 + std::hash<size_t>()(individual.distance);
+        res = res * 31 + std::hash<size_t>()(individual.capacityExcess);
+        res = res * 31 + std::hash<size_t>()(individual.timeWarp);
+
+        return res;
```
leonlan, line 127: Question: what happens if the two hash values of an individual happen to be the same? Individuals that have the same but permuted routes all have the hash value right?

pyvrp/cpp/SubPopulation.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+void SubPopulation::purge()
+{
+    while (size() > params.minPopSize)
+    {
+        // Remove duplicates from the subpopulation (if they exist)
+        auto const pred = [&](auto &iterator)
+        {
+            return !iterator.proximity.empty()
+                   && *iterator.proximity[1].second == *iterator.individual;
```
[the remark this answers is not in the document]
N-Wouda: Cause it was late yesterday and I'm not the sharpest 😅. Let me fix that!

pyvrp/cpp/Individual.h
N-Wouda: Objects with the same hash value end up in the same bucket in a hash table (basically: they're put under the same index), and a hash table then has to linearly search through all items in the bucket using `operator==` to compare the items stored there. So typically you want your hash values to spread out, which is the most efficient way to use a hash table (spreading over all the buckets, which limits the amount of linear searches needed). 

For permuted individuals, that's fine. They are equal when compared using `operator==`, so the search stops at the first duplicate.

N-Wouda:
Since there's quite a bit of variability in the results, I'm running two benchmarks with ten seeds each for current main and this branch. Results should be in in a few hours.

Main (0d102e4e242e26177ea8807fa74516acecb060f8) [job ID 28700256, seeds 1-10]:
```
TODO
```

This branch (494bc4bea4b51920a271755f0fb494552b90162b) [job ID 28700246, seeds 1-10]:
```
TODO
```

pyvrp/cpp/Individual.h
wouterkool: I have two additional questions:
* Why do we need the hash function? I think this is to efficiently identify duplicates among different objects, correct? Where do we do that?
* The hash function changes if the individual changes, which is currently possible (e.g. `LocalSearch::exportIndividual`). What does that mean if this individual is inside some collection using the hash? Is there an implicit contract that individuals should not change once they're in some collection? (which makes sense but is tricky, we already have challenges with cost functions changing with penalties, which you luckily did not use in the hash ;)).

pyvrp/cpp/SubPopulation.h
```diff
@@ -0,0 +1,103 @@
...
+                     double lbDiversity = 0.1,
+                     double ubDiversity = 0.5)
+        : minPopSize(minPopSize),
+          generationSize(generationSize),
+          nbElite(nbElite),
+          nbClose(nbClose),
+          lbDiversity(lbDiversity),
+          ubDiversity(ubDiversity)
+    {
+        if (0 > lbDiversity || 1 < lbDiversity)
```
wouterkool, line 35: Are 0 and 1 not allowed? If not the exception should mention the open interval (0, 1) but I think we can do `>=` and `<=` here as well as below?

wouterkool: I get that 0 may get you two equal parents but I'm not sure if we need to forbid, even though offspring is the same potentially the local search is randomized (e.g. intensification) so this can still be useful. And I think at least 1 should be allowed)

pyvrp/cpp/SubPopulation.cpp
```diff
@@ -0,0 +1,156 @@
...
+        // Remove the worst individual (worst in terms of biased fitness)
+        updateFitness();
+
+        auto const worstFitness = std::max_element(
+            items.begin(),
+            items.end(),
+            [](auto const &a, auto const &b) { return a.fitness < b.fitness; });
+
+        remove(worstFitness);
+    }
```
wouterkool, line 109: Shouldn't we have an `updateFitness()` here after we remove the last individual? If this is indeed missing, I think we should add a test that fails for the current implementation.

pyvrp/cpp/SubPopulation.cpp
```diff
@@ -0,0 +1,156 @@
...
+        return;
+
+    std::vector<size_t> byCost(size());
+    std::iota(byCost.begin(), byCost.end(), 0);
+
+    std::stable_sort(
+        byCost.begin(),
+        byCost.end(),
+        [&](size_t a, size_t b)
+        { return items[a].individual->cost() < items[b].individual->cost(); });
```
wouterkool, line 124: Note that this will compute cost per individual multiple times. I think it's probably fast enough currently but good to be aware of if that were to change.

pyvrp/cpp/SubPopulation.cpp
```diff
@@ -0,0 +1,156 @@
...
+        auto const worstFitness = std::max_element(
+            items.begin(),
+            items.end(),
+            [](auto const &a, auto const &b) { return a.fitness < b.fitness; });
+
+        remove(worstFitness);
+    }
+}
+
+void SubPopulation::updateFitness()
```
wouterkool, line 112: Maybe add some comment on when this is expected to be called. I think it's 'eager' for now rather than 'lazy' in the sense that it should always be up to date after a change, rather than before a use of the fitness.

If indeed the policy is 'eager', shouldn't we trigger this function directly once penalties change? Now it relies on an individual being added, which means we probably do a binary tournament first with cost based on old penalties. Now, I think that makes negligible difference in practice, but it is not what I'd expect and rather subtle.

pyvrp/tests/test_Population.py
```diff
@@ -189,56 +185,115 @@ def test_select_returns_same_parents_if_no_other_option():
...
+
+    # Now we compare the observed ranking from the binary tournament selection
+    # against what we would expect from the actual fitness ranking. We compute
+    # the percentage of times we're incorrect, and test that that number is not
+    # too high.
+    actual_rank = np.argsort(-infeas_count)  # higher is better
+    expected_rank = np.arange(len(infeas))
+    pct_off = np.abs((actual_rank - expected_rank) / len(infeas)).mean()
+
+    assert_(pct_off < 0.05)
```
wouterkool, line 299: Would it be possible to bring back the biased fitness computation test as well (I know it tests internals but so do some other tests here). https://github.com/N-Wouda/PyVRP/blob/population-generics/pyvrp/tests/test_Population.py#L345

While less convenient, we could look up the index for a random individual to define a custom diversity function so it should be possible. I am curious if that test would currently pass given that I think we miss an `updateFitness` call in `purge` (but I may have missed something here).

wouterkool (commented):
I'm fine if we keep it this way for now although I see it as a compromise that we need to keep the Individuals on the c++ side, especially given that sorting depends on cost(), which depends on penalty manager, which can change at random times which may give undefined behavior, while this is not clear at all. Admittedly, the problem is probably not that big and not introduced in this PR, so I think it's not worth the time but I would have liked if we could have fixed it. For now, let's keep it but please check if we call `updateFitness` where needed (after `purge`?).

pyvrp/cpp/SubPopulation.cpp
N-Wouda: The current implementation does
image
Which this function exactly replicates. We could move the call to update the biased fitness to the `remove` member altogether and make it a private method. I think that makes more sense. What do you think?

pyvrp/cpp/SubPopulation.h
N-Wouda: But we do not forbid either zero or one?

N-Wouda: Although `lbDiversity = 1` will not pass because a condition below this (`lbDiversity <= ubDiversity`) fails.

pyvrp/cpp/SubPopulation.cpp
wouterkool: I agree so the current implementation seems incorrect? Downside of moving it to remove (which is already private I think?) is that you do not need to update the biased fitness in between of removing multiple exact duplicates so maybe just keep it but add the one update? Or let remove do it by default but make it optional to not do it for efficiency? I don't care that much.

N-Wouda: I'm not really worried about performance here tbh. Everything on the C++ side is already (mostly) a rounding error; the whole subpopulation takes <1% of our current runtimes.

So I think we can make it simpler for ourselves and just always update after removing. That way we know it's correct without having to think about exceptional cases. We just update when adding and update when removing.

N-Wouda: >so the current implementation seems incorrect? 

So yes, I think so.

pyvrp/cpp/SubPopulation.cpp
N-Wouda: >Maybe add some comment on when this is expected to be called. I think it's 'eager' for now rather than 'lazy' in the sense that it should always be up to date after a change, rather than before a use of the fitness.

That's a good idea, I'll do this in a sec.

>If indeed the policy is 'eager', shouldn't we trigger this function directly once penalties change?

I think so, but... that updating is entirely internal to the `PenaltyManager` right now. It'd be quite a bit of work to make it possible to listen to that event, and then update our fitness vector based on it. Since this has been an issue going back to HGS-CVRP (I checked just yet, to be sure), I'm inclined to leave this as "wontfix".

I'll add a TODO at the top of the biasedFitness function about this.

pyvrp/cpp/SubPopulation.h
N-Wouda: I've flipped the inequalities to hopefully be clearer now.

pyvrp/cpp/SubPopulation.cpp
N-Wouda: Changing this hasn't impacted the individuals in the subpopulation when compared to the implementation on current main, so in that regard we're OK. I also ran once more with the profiler, and `Population::add` is still around 0.2s on 1000 iteration runs of `RC1_10_1`. So I think we're good here!

pyvrp/tests/test_Population.py
N-Wouda: I'll try to do this in a new `test_SubPopulation` file, but when I tried to do that earlier today it turned into a right mess.

N-Wouda: I've tried doing this, but my attempts so far have boiled down to basically testing whether the fitness values are computed exactly the way they're computed. That's a pretty poor test. I've so far added a few tests that checks the average distance of the closest few individuals, which is an important component in those calculations.

N-Wouda:
@wouterkool what further changes would you like to see?

pyvrp/cpp/Individual.h
N-Wouda: Sorry, I had missed this comment yesterday because the conversation was already marked resolved.

>Why do we need the hash function? I think this is to efficiently identify duplicates among different objects, correct? Where do we do that?

In a number of tests, both for mappings (dict keys) and set membership.

>What does that mean if this individual is inside some collection using the hash? Is there an implicit contract that individuals should not change once they're in some collection? 

As a C++ programmer: yeah, that'd a user issue. As a Python programmer: yeah, we should probably make sure that does not happen. I do really like the idea of being able to hash individuals, however, since it's pretty convenient when you need to do a reverse lookup of individuals to their index in tests. 

In #137 and #138 I fix this problem by making the Individual immutable altogether. That also simplifies some of the complications in the GeneticAlgorithm needing to copy the new best solution - it's fairly straightforward now.

N-Wouda:
Blocked by #138.

pyvrp/tests/test_Population.py
wouterkool: I think it's not a poor test given that the implementation computes it differently (e.g. by caching closest indivs). Checking that this is consistent with a simple but inefficient implementation of the definition makes sense to me. This argument was stronger with the numpy implementation but I think it is still useful to have a good test on fitness values. I do think the fact that we can't test it easily without making a mess suggests things are too much entangled but given that we have other things to do as well I suggest we leave it for now.

wouterkool (approved):

N-Wouda:
@wouterkool @leonlan thanks for the reviews - this PR has really gotten a lot better as a result!

Issue #107, opened 2023-02-20
Wrap C++ objects in Python


N-Wouda:
In particular the LocalSearch and Individual objects would benefit from this, and it would obviate the need for the pyi files for these items. The relevant binary modules could then be prefixed by an underscore (e.g., as `_Individual.cpython.*.so`).

N-Wouda:
@wouterkool I could not find an issue for this, so I made a new one.

N-Wouda:
This is kind of important, because the (unwrapped) C++ modules no longer show up in the documentation due to the `_` prefix.

N-Wouda:
I'm pretty happy with the current set-up, so I don't think we need to keep this issue around for now.

PR #121, opened 2023-02-24
Wrap local search and intensify at each iteration

### LINKED ISSUE #98: Use local search intensification at every iteration with some probability
The intensification step of local search is currently only applied when a new best solution is found. The original implementation of HGS uses the intensification step at every iteration after the node ops are exhausted. We will try to bring this back to see if we can get better performance.

Related commit from Euro-NeurIPS: https://github.com/N-Wouda/Euro-NeurIPS-2022/commit/1031b8a90742a2756f85f3f415d5189e1fc1fae0

leonlan:
This PR:
- [x] Closes #98.
- [x] Partially addresses #107 by wrapping the `LocalSearch` object and prepending all extension module names with an underscore.


meson.build
```diff
@@ -114,13 +114,25 @@ foreach extension : extensions  # extension[0] = name, extension[1] = subdir
...
+        py.extension_module(
+            extension[0],
+            source_dir / extension[0] + '_bindings.cpp',
+            dependencies: [py.dependency(), pybind11],
+            link_with: libcommon,
+            install: true,
+            subdir: install_dir,
+            include_directories: [include_directories(SRC_DIR)],
+        )
+    endif
```
N-Wouda, line 137: Can't we make the extension name start with an underscore? Or does that go wrong, and do we need one more field?

pyvrp/educate/LocalSearch.py
```diff
@@ -0,0 +1,52 @@
...
+        Parameters
+        ----------
+        data
+            Data object describing the problem to be solved.
+        penalty_manager
+            Penalty manager to use.
+        rng
+            Random number generator.
+        neighbours
+            Matrix that defines the local search neighbourhood.
```
N-Wouda, line 32: `neighbours` need not be a matrix, it's just a list of (possibly variable length) lists.

wouterkool (approved):
Looks good. In general, would it be an idea to make neighbours optional in the constructor, and compute it by default using `compute_neighbours`. Same for operators, I think we can just take params node_operator and route_operators in the constructor, and by default use the default operators, then not make the `add_*_operator` functions part of the public API? (depends a bit on where we want to go in general)

pyvrp/educate/LocalSearch.py
```diff
@@ -0,0 +1,69 @@
...
+    def intensify(self, individual: Individual) -> bool:
+        # Runs intensification on the individual and returns whether an
+        # improvement was found (TODO let c++ return this)
+        cost = individual.cost()
+        self._ls.intensify(individual)
+        return individual.cost() < cost
+
+    def search(self, individual: Individual):
+        # TODO return whether improvement was found
+        self._ls.search(individual)
```
N-Wouda, line 69: We can freely speed this up further if we move import/export of individuals out of search() and intensify() in the C++ implementation. I don't have numbers on how much this matters, but I could profile and find out?

leonlan: Sure thing

pyvrp/GeneticAlgorithm.py
```diff
@@ -146,11 +152,14 @@ def _educate(self, individual: Individual):
             and self._rng.rand() < self._params.repair_probability
         ):
             with self._pm.get_penalty_booster() as booster:  # noqa
-                self._ls.search(individual)
+
+                # TODO shouldn't we resample the 'intensify' variable here?
```
wouterkool, line 156: Can you fix this TODO?

pyvrp/GeneticAlgorithm.py
```diff
@@ -146,11 +152,14 @@ def _educate(self, individual: Individual):
...
+                # this is what HGS used to do
+                self._ls.run(individual, intensify)
 
                 if individual.is_feasible():
                     if (
-                        self._params.should_intensify
+                        self._params.intensify_on_best
```
wouterkool, line 162: We do the intensify before before saving the new best, whereas the individual in theory may become infeasible by intensification so shouldn't we check and store the new best both before and after doing this intensifcation?

pyvrp/educate/LocalSearch.py
N-Wouda: About 0.1% of the total time for each, so probably not worth changing :).

N-Wouda: (or at least, I think so. I'm not entirely sure how `perf` from the commandline works yet)

N-Wouda:
@leonlan let us know when you want this to be reviewed again!

pyvrp/GeneticAlgorithm.py
```diff
@@ -121,23 +125,31 @@ def run(self, stop: StoppingCriterion):
...
+
+        intensification_prob = self._params.intensification_probability
+        intensify = self._rng.rand() < intensification_prob
+
+        self._ls.run(individual, intensify)
 
-        if individual.is_feasible() and individual.cost() < self._best.cost():
-            self._best = individual
+        if _is_new_best(individual):
+            self._best = Individual(individual)
```
leonlan, line 143: GA collecting best individuals is possibly still bugged on main. Since we are not copying the individual, any subsequent intensification or repair step can possibly change the feasibility of the best individual.

pyvrp/GeneticAlgorithm.py
```diff
@@ -23,13 +23,17 @@
 class GeneticAlgorithmParams:
     repair_probability: float = 0.80
     collect_statistics: bool = False
-    should_intensify: bool = True
+    intensification_probability: float = 0.15
```
N-Wouda, line 26: What about `intensify_probability`? That meshes with `intensify_on_best`, and is quite a bit shorter.

pyvrp/GeneticAlgorithm.py
```diff
@@ -121,23 +125,31 @@ def run(self, stop: StoppingCriterion):
...
 
-        # Only intensify feasible, new best solutions. See also the repair
-        # step below.
-        if (
-            self._params.should_intensify
-            and individual.is_feasible()
-            and individual.cost() < self._best.cost()
-        ):
-            self._ls.intensify(individual)
+        def _add_and_register(indiv):
```
N-Wouda, line 131: The privateness on these helpers is a bit much: they're already internal to the `_educate` scope :).

pyvrp/GeneticAlgorithm.py
```diff
@@ -121,23 +125,31 @@ def run(self, stop: StoppingCriterion):
...
+            self._pop.add(indiv)
+            self._pm.register_load_feasible(not indiv.has_excess_capacity())
+            self._pm.register_time_feasible(not indiv.has_time_warp())
+
+        intensification_prob = self._params.intensification_probability
+        intensify = self._rng.rand() < intensification_prob
```
N-Wouda, line 137: `should_intensify`? To me, just "intensify" does not really imply this is a flag.

pyvrp/cpp/educate/CircleSector.h
```diff
@@ -0,0 +1,82 @@
+#ifndef CIRCLESECTOR_H
```
N-Wouda, line 1: We really cannot avoid this thing, can we? 😢

N-Wouda (commented):
Looks good to me. I think we should also prefix the pyi files with an underscore, to match the new extension names. Could you add that in as well?

N-Wouda:
I'm going to make a new issue about the "overlap degrees" thing. Instead of a degree in [0, 360], I think it makes much more sense to use a percentage [0, 1.0]. We can translate that back to a degree/CircleSector integer internally, but then users don't have to understand that there's a circle sector thing around at all. It's just "how much should routes overlap before we consider them in the intensification routine".

pyvrp/cpp/educate/CircleSector.h
leonlan: Nope, because it's currently the best thing we have to identify two neighbouring route pairs. We could at some point think of a "RouteNeighbours" object. That could generalize the circle sectors thing, but also implement other route neighbor selection schemes such as % overlapping nodes or some other heuristic.

PR #129, opened 2023-02-25
Support copying Individual

leonlan:
This PR adds support for (deep)copying individuals.


pyvrp/cpp/Individual_bindings.cpp
```diff
@@ -29,6 +29,15 @@ PYBIND11_MODULE(_Individual, m)
         .def("has_excess_capacity", &Individual::hasExcessCapacity)
         .def("has_time_warp", &Individual::hasTimeWarp)
         .def("__eq__", &Individual::operator==)
+        .def(
```
N-Wouda, line 32: Can we now also remove the copy constructor binding above? (the last of the `py::init`'s).

N-Wouda (commented):
LGTM, except that we still bind the copy constructor as an overloaded `__init__`. That should go.

pyvrp/cpp/Individual_bindings.cpp
```diff
@@ -29,6 +29,15 @@ PYBIND11_MODULE(_Individual, m)
...
+            "__copy__",
```
N-Wouda, line 33: I forgot this in my earlier review, but can we add a test that checks copy.copy and copy.deepcopy actually return a copy (that is, == to original, but not when using `is`})?

PR #134, opened 2023-02-26
Add Valgrind to CI

### LINKED ISSUE #124: Valgrind action
I caused a memory leak in #120 (fixed now). A bit embarrassing, especially since tools exist that can help us discover these. One such tool is Valgrind. We can run that in the CI just to be sure, using Python's valgrind suppression file: https://svn.python.org/projects/python/trunk/Misc/valgrind-python.supp.

N-Wouda:
This PR closes #124.

<details>

**Notes**:

</details>


N-Wouda:
I think this is a good addition, but Valgrind is a bit.. flaky: on Py3.8 it works fine (no errors), but for Py3.11 it shows a lot of false positives. At least, I think they are false positives. I'll keep it in the CI for now, so we can periodically inspect the results, but I wouldn't consider a run 'failed' just because Valgrind is complainy.

Issue #132, opened 2023-02-26
High-level interfaces `solve_cvrp` and `solve_vrptw`


leonlan:
From https://github.com/N-Wouda/PyVRP/issues/48#issuecomment-1406684143

> I think we should do the following.
> 
> First, we offer two high-level solver routines in the pyvrp package: `solve_cvrp()` and `solver_vrptw()`. These take only the bare minimum of arguments (perhaps only a ProblemData instance) and are intended for users that only want to solve the problem, and do not care at all about _how_. This is intentionally a black box. These routines return a Result object that contains some statistics, and the best solution, and everything else is basically a hidden implementation detail.
> 
> For these functions, we compile two versions of our codebase: one where we use the TWS data, and another where we do not, using compile time flags. The compiler will optimise for each case. In the high-level solver routines, we then use the appropriate binary to implement the solver function.
> 
> For people's own use cases, we expose only the full/complete components, not the faster, problem-specific implementations.

N-Wouda:
Purely realistically, this is something we're going to do after the initial submission. For the immediate future I think the different compilation modes are sufficient.

N-Wouda:
The interface for this can be something like `solve_<variant>(data: ProblemData)`. Together with the high-level interface of #217 it should then be pretty easy to model and solve each variant.

N-Wouda:
We promised this in the paper, but I just tinkered with it a bit and would like not to do this (yet). It's particularly tricky because we have objects that are wrapped in Python - those have hardcoded dependencies on the location of a particular native extension. If we want this, we need to think about a way to do that differently.

For now it's enough for me that we support this via compilation flags.

N-Wouda:
OK, I just checked the paper again, and I don't think we explicitly promise this. We just talk about PyVRP being easy to use. Then I think #217 and the associated PR are enough for now.

N-Wouda:
@leonlan @wouterkool shall we close this?

leonlan:
Sure #236 closes this for me

PR #136, opened 2023-02-26
CVRP example notebook 

### LINKED ISSUE #102: Switch assert_almost_equal over to assert_allclose in tests
See [here](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html):
image
Using [`assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) instead avoids any of these issues.

### LINKED ISSUE #112: Automatic testing of (example) notebooks

### LINKED ISSUE #130: Example notebook CVRP
I'll try to make an example notebook for CVRP.

leonlan:
This PR
- [x] Closes #102.
- [x] Closes #130.
- [x] Refactors the VRPTW notebook.
- [x] Improves some plotting and diagnostics functions.
- [x] Closes  #112: run notebooks in the CI.


leonlan:
This PR is ready to be reviewed. 

I haven't been able to fix the plotting tests - could you let me know how to reproduce the images to make it pass the tests? I tried saving the images using the `plt.figure` from each function and then save it with `plt.savefig`, but that did not work.

docs/source/index.rst
```diff
@@ -23,6 +23,7 @@ TODO
    :caption: Examples
 
    examples/vrptw
+   examples/cvrp
```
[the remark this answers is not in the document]
N-Wouda: I think 1.5min is OK, but there's no need to run this on all jobs. Maybe limit it to run on only a single job?

pyvrp/plotting/plot_objectives.py
```diff
@@ -58,6 +58,10 @@ def _plot(x, y, *args, **kwargs):
     y = [d.avg_cost for d in result.stats.feas_stats]
     _plot(x, y, label="Feas. avg.", c="tab:green", alpha=0.3, ls="--")
 
+    # Use best objectives to set reasonable y-limits
+    best = np.nanmin([d.best_cost for d in result.stats.feas_stats])
+    ax.set_ylim(best * 0.95, best * 1.15)
```
N-Wouda, line 63: Can we make `ylim_adjust` a tuple argument with defaults `(0.95, 1.15)`?

N-Wouda: I can imagine this is really run-dependent, so 0.95 and 1.15 should work in most cases, but sometimes they might not. It'd be nice to have some control on those cases.

pyvrp/plotting/plot_coordinates.py
```diff
@@ -34,7 +34,7 @@ def plot_coordinates(
     kwargs = dict(c="tab:red", marker="*", zorder=3, s=500)
     ax.scatter(x_coords[0], y_coords[0], label="Depot", **kwargs)
 
-    ax.scatter(x_coords[1:], y_coords[1:], s=75, label="Clients")
+    ax.scatter(x_coords[1:], y_coords[1:], s=50, label="Clients")
```
N-Wouda, line 37: 👍

pyvrp/plotting/plot_solution.py
```diff
@@ -50,7 +50,7 @@ def plot_solution(
 
             # Edges from and to the depot, very thinly dashed.
             kwargs = dict(ls=(0, (5, 15)), linewidth=0.25, color="grey")
-            ax.plot([x_coords[0], x[0]], [y_coords[0], y[1]], **kwargs)
+            ax.plot([x_coords[0], x[0]], [y_coords[0], y[0]], **kwargs)
```
N-Wouda, line 53: 👍

pyvrp/read.py
```diff
@@ -104,16 +104,21 @@ def read(
...
 
+    if "time_window" in instance:
+        time_windows = round_func(instance["time_window"])
+    else:
+        # The default value for the latest time window based on the maximum
+        # route duration. This ensures that the time window constraints are
+        # always satisfied.
+        bound = num_clients * (edge_weight.max() + service_times.max())
+        bound = bound if bound < _INT_MAX else _INT_MAX
+        time_windows = np.repeat([[0, bound]], num_clients, axis=0)
```
N-Wouda, line 120: So this solves our CVRP/VRPTW INT_MAX woes? Good!

Suggestion: `bound = <ternary expression>` is a somewhat non-obvious way of writing `bound = min(bound, _INT_MAX)`.

N-Wouda (commented):
Notebook:

- "This build can be used to solve CVRP faster about 10-20% faster, but we remark that the default PyVRP library can also be used to solve CVRP." in conclusion; double faster.
- Header importance seems a bit weird: there's now "Compiling", "Example", and "Conclusion". I guess everything under "Example" could be at the same level as the other two?
- I get that it's not obvious that one can compile for CVRP, but is that something we want to spend half this notebook on explaining? If yes: this is probably not the place; maybe the contributing section? If no: I'd run with the default VRPTW solver for now, to keep the notebook simple. Then later once we have #132 we can switch out those components and get a free performance improvement.

N-Wouda:
> I haven't been able to fix the plotting tests - could you let me know how to reproduce the images to make it pass the tests? I tried saving the images using the `plt.figure` from each function and then save it with `plt.savefig`, but that did not work.

@leonlan there should now be a folder (I think called `result_images`) that contains the generated image, the expected image, and the diff. I think some of the changes you've made here change the figures, so you should replace the image in `pyvrp/plotting/tests/baseline_images` with the newly generated image in `result_images` (for those images where you'd expect things to change as a result of your code changes).

N-Wouda:
For the CVRP notebook: there's a bunch of text [here](https://alns.readthedocs.io/en/latest/examples/capacitated_vehicle_routing_problem.html#The-capacitated-vehicle-routing-problem) that really explains everything well. We could copy that over?

N-Wouda:
FYI:
```python
    node_ops = [node_op(data, pen_manager) for node_op in NODE_OPERATORS]

    for op in node_ops:
        ls.add_node_operator(op)

    route_ops = [route_op(data, pen_manager) for route_op in ROUTE_OPERATORS]

    for op in route_ops:
        ls.add_route_operator(op)
```
can now be rewritten as
```python
for op in NODE_OPERATORS:
   ls.add_node_operator(op(data, pen_manager))

for op in ROUTE_OPERATORS:
   ls.add_route_operator(op(data, pen_manager))
```
The `LocalSearch` object will keep the operator alive until it itself goes out of scope.

.github/workflows/CI.yml
```diff
@@ -103,3 +103,10 @@ jobs:
         run: |
           VALGRIND_CMD="pyvrp pyvrp/tests/data/ORTEC-VRPTW-ASYM-0bdff870-d1-n458-k35.txt --max_runtime 0.5 --seed 1"
           poetry run env PYTHONMALLOC=malloc valgrind --tool=memcheck --quiet $VALGRIND_CMD
+      - if: matrix.compiler == 'clang' && matrix.python-version == '3.11'
+        # Run this only for one job to reduce workflow runtimes.
+        name: Run example notebooks
+        run: |
+          poetry install --with examples
+          poetry run jupyter nbconvert --to notebook --execute examples/cvrp.ipynb
+          poetry run jupyter nbconvert --to notebook --execute examples/vrptw.ipynb
```
N-Wouda, line 112: This is not covered by the cache. Can we update
```
      - name: Install Python dependencies
        if: steps.cache-python.outputs.cache-hit != 'true'
        run: poetry install --no-root
```
to include ```--with examples```? I'll clear the cached dependencies to make this work, let me know when.

Then,
```
          poetry run jupyter nbconvert --to notebook --execute examples/cvrp.ipynb
          poetry run jupyter nbconvert --to notebook --execute examples/vrptw.ipynb
```
This needs to be updated every time we have a new notebook. Maybe
```
for f in examples/*.ipynb; do poetry run jupyter nbconvert --to notebook --execute $f; done

File: .github/workflows/CI.yml [hunk 7]
```diff
@@ -17,7 +17,7 @@ jobs:
       ${{ matrix.compiler-version }} and Python ${{ matrix.python-version }}
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: true
+      fail-fast: false
```
N-Wouda, line 20: Can we revert this now that we have fixed the plots?

leonlan:
@N-Wouda Could you update the CI cache?

N-Wouda:
@leonlan done; I've restarted the jobs.

PR #138, opened 2023-02-26
Make Individual an immutable value object

### LINKED ISSUE #137: Make Individual a value object
Right now, the Individual is _almost_ immutable: after construction, it can only be queried for values. The only mutability currently supported comes from `operator=`, and that only exists so we can modify the argument given to `LocalSearch::search` and `LocalSearch::intensify`.

That we implement things that way exists only for historical reasons: HGS-CVRP and the like did it this way, and thus we also do. But we don't really have to keep doing it this way. Making the Individual a value object that's immutable after construction should not really hurt performance* (since LS uses its own representation anyway), but makes it easier to reason about an Individual throughout its lifetime.

---

*but we should check to be sure.

N-Wouda:
This PR:

- [x] Closes #137.
- [x] Makes Individual hashable.

Making Individuals immutable doesn't seem to hurt performance (but I'm running a benchmark just to be sure) and makes it much easier to reason about what gets changed when (which is particularly important when registering new best solutions). As a consequence, there's no real need for `deepcopy(individual)` any more.

### Benchmarks

This branch (ecfb6b17abdb70e914c226699b3405c1431a8b3d) [job ID 28701010, seed 1-10]:
```
 Avg. objective: 334211.6 (min: 334097, max: 334364)
Avg. iterations: 135072.2 (min: 122816, max: 145283)
```
Cf. current main (0d102e4e242e26177ea8807fa74516acecb060f8):
```
 Avg. objective: 334303.6 (min: 334093, max: 334668)
Avg. iterations: 123359.6 (min:  99403, max: 143166)
```

So no meaningful difference, as expected.


N-Wouda:
This PR partially duplicates #120, but I'm going to resolve that after this is merged.

N-Wouda:
When I profile this there's no meaningful difference in runtimes between current main and this PR. So I'm reasonably confident this does not come at a performance hit. I hadn't really expected that either, but it's good to make sure.

leonlan (approved):

PR #141, opened 2023-02-26
New neighbourhood parameter values

leonlan:
This PR:
- [x] Adds reading neighbourhood parameters from config files in `cli.py`.
- [x] Changes the default neighbourhood parameter values.

---
I'm running a 10-seed benchmark for VRPTW. 

```
Cost:  333667.4
Iters: 109110.6
```


N-Wouda:
Ah #144 is the same as this PR, and this will provide the benchmark.

leonlan:
This is definitely an improvement, so I'll merge it in.

PR #145, opened 2023-02-27
Test SWAP*

### LINKED ISSUE #139: Test SWAP*
The SWAP* operator is the last big code component that's currently not under test. We should add some tests to ensure its basic functionality is correct.

N-Wouda:
This PR:

- [x] Closes #139.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/educate/tests/test_SwapStar.py
+++ b/pyvrp/educate/tests/test_SwapStar.py
@@ -15,8 +15,8 @@ def test_swap_star_identifies_more_moves_than_regular_swap():
     """
     SWAP* can move two clients to any position in the routes, whereas regular
     swap ((1, 1)-exchange) must reinsert each client in the other's position.
-    Thus, SWAP* should be able to identify more improving moves than the
-    regular (1, 1)-exchange move.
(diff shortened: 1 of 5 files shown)


N-Wouda:
@leonlan can you have a look? It's not a lot, but at least it tests some of the expected behaviours of SWAP*. This is somewhat similar to how we tested the other operators as well (so far), so I think it's good to go.

pyvrp/educate/tests/test_SwapStar.py
```diff
@@ -1 +1,70 @@
...
+
+        # The regular swap operator should have been able to improve the random
+        # individual. After swap gets stuck, SWAP* should still be able to
+        # further improve the individual.
+        assert_(swap_individual.cost() < individual.cost())
+        assert_(swap_star_individual.cost() < swap_individual.cost())
+
+
+@mark.parametrize("seed", [2643, 2742, 2941, 3457, 4299, 4497, 6178, 6434])
+def test_RC208_instance(seed: int):
```
leonlan, line 50: Maybe rename this test? E.g., `test_swap_star_same_number_of_routes`.

leonlan (approved):

Issue #113, opened 2023-02-22
Individual.cost() incorrect occassionally


wouterkool:
#### Describe the bug

When running the example VRPTW notebook, sometimes the `individual.cost()` resulting from the first optimization (on instance RC208) is not correct and gives a large value (e.g. `3246399057`). The routes are correct and when initializing a new individual with the same routes and a new penalty manager. Making a copy via the copy constructor the costs are still incorrect, so I suspect this has to do with the fact that the individual has a reference to the penalty manager, which gets deallocated as it goes out of scope after returning from the `solve()` function, but I cannot reproduce this consistently.

In general, should the PenaltyManager be kept alive as long as the solution is alive? This is an undesired situation I think and it would be better to pass PenaltyManager or a penalties object explicitly to the `Individual::cost()` function (see #100).

#### Code sample

See [examples/vrptw.ipynb](https://github.com/N-Wouda/PyVRP/blob/482bf1a71e2bf150b69206eaa3f8a29669a2d8c2/examples/vrptw.ipynb).

```python
# See vrptw.ipynb first run
# args = {"seed": 42, "max_runtime": 10}
# result = solve(instance, **args)
# solution = result.best

routes = solution.get_routes()
route_stats = diagnostics.get_all_route_statistics(solution, instance)

# Make a copy using individual
solution2 = Individual(solution)
route_stats2 = diagnostics.get_all_route_statistics(solution2, instance)

# Make a copy using routes
pm = PenaltyManager(instance.vehicle_capacity)
solution3 = Individual(instance, pm, routes)
route_stats3 = diagnostics.get_all_route_statistics(solution3, instance)

print(solution.has_excess_capacity(), solution.has_time_warp())
print(solution.cost(), sum([s.distance for s in route_stats]))
print(solution2.cost(), sum([s.distance for s in route_stats2]))
print(solution3.cost(), sum([s.distance for s in route_stats3]))
```
gives:
```
False False
3246399057 7761
3246399057 7761
7761 7761
```

This doesn't reproduce the issue and seems to work fine:
```python
# Create individual with penalties trying to reproduce incorrect cost bug
def test():
    pm = PenaltyManager(instance.vehicle_capacity)
    sol = Individual(instance, pm, [[i + 1 for i in range(100)]])
    # Test result inside function
    print(sol.cost())
    return sol

# Test result outside function
sol = test()
print(sol.cost())

# Test after garbage collection
import gc
gc.collect()
print(sol.cost())
```
outputs
[...]

PR #151, opened 2023-02-27
API documentation

N-Wouda:


N-Wouda:
@leonlan why does the new notebook stuff fail?

leonlan:
> @leonlan why does the new notebook stuff fail?

I have no idea. You can remove it from the CI and then I’ll look into this tomorrow morning.

leonlan (approved):

leonlan:
> @leonlan why does the new notebook stuff fail?

Does it something have to do with https://github.com/N-Wouda/PyVRP/issues/113#issuecomment-1446942862?
The CI now passes wihtout any changes. https://github.com/N-Wouda/PyVRP/pull/152 also passes the CI. The assertion that failed was probably due to `solution.cost()` returning an incorrect value.

N-Wouda:
> Does it something have to do with [#113 (comment)](https://github.com/N-Wouda/PyVRP/issues/113#issuecomment-1446942862)? The CI now passes without any changes. #152 also passes the CI. The assertion that failed was probably due to `solution.cost()` returning an incorrect value.

Could well be! I'll add some docs here for the operators as well, and then merge this in. I'm pretty confident the documentation didn't cause the issue :).

PR #152, opened 2023-02-27
Remove data from diversity functions, SubPopulation, add documentation

N-Wouda:


N-Wouda:
@wouterkool you'll like this, I think

pyvrp/diversity/_broken_pairs_distance.pyi
```diff
@@ -1,5 +1,36 @@
-from pyvrp import Individual, ProblemData
+from pyvrp._Individual import Individual
 
-def broken_pairs_distance(
-    data: ProblemData, first: Individual, second: Individual
-) -> float: ...
+def broken_pairs_distance(first: Individual, second: Individual) -> float:
```
wouterkool, line 3: Can we add that this is directed broken pairs distance? (As we may want to have an undirected broken pairs distance for CVRP)?

pyvrp/_ProblemData.pyi
```diff
@@ -3,73 +3,68 @@ from typing import Any, List, Tuple
...
+    duration_matrix
+        A matrix that gives the travel times between clients (and the depot at
+        index 0). Does not have to be symmetric.
+
+    Notes
+    -----
+    All array data assume that the data at or involving index 0 relates to the
+    depot, and all other indices specify client information.
+    """
+
```
wouterkool, line 67: In general, should we place the docstring for the constructor at the class level rather than in `__init__`? There are a few more places where this applies I think?

wouterkool: Update: I see this follows the standard https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes, so let's do this everywhere! I think we can handle that in https://github.com/N-Wouda/PyVRP/issues/133.

wouterkool (approved):
@N-Wouda I like it! I don't like that it was that simple after all though. ;)

pyvrp/_ProblemData.pyi
N-Wouda: Sounds good to me. I moved this over because the docs won't show `__init__` stuff by default, and we do want to display that kind of information. We'll pick up the rest as we add more docs I think.

PR #155, opened 2023-02-28
Add all operators to NODE_OPERATORS, explain in docs

N-Wouda:
This PR:
- Adds all operators to the default `NODE_OPERATORS` and `ROUTE_OPERATORS` lists;
- Makes these operators configurable from a config file.

### DIFF
--- a/pyvrp/cli.py
+++ b/pyvrp/cli.py
@@ -132,14 +132,18 @@ def solve(
     neighbours = compute_neighbours(data, nb_params)
     ls = LocalSearch(data, pen_manager, rng, neighbours)
 
-    for op in config.get("node_ops", NODE_OPERATORS):
-        if isinstance(op, str):
-            op = getattr(pyvrp.educate, op)
+    node_ops = NODE_OPERATORS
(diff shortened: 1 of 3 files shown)


N-Wouda:
We should also make this configurable from the CLI

pyvrp/cli.py
```diff
@@ -267,6 +272,14 @@ def main():
     """
     parser.add_argument("--config_loc", help=msg)
 
+    msg = "Node operators to apply. Defaults to all."
+    node_ops = [op.__name__ for op in NODE_OPERATORS]
+    parser.add_argument("--node_ops", nargs="+", default=node_ops, help=msg)
+
+    msg = "Route operators to apply. Defaults to all."
+    route_ops = [op.__name__ for op in ROUTE_OPERATORS]
+    parser.add_argument("--route_ops", nargs="+", default=route_ops, help=msg)
```
leonlan, line 281: How about we move the ops to the config? That to me feels like the right place to define the algorithm configuration, which LS operators are part of.

leonlan (approved):

PR #160, opened 2023-03-01
Initial submission configs

leonlan:
This PR adds the configs used for the initial submission.

### DIFF
--- a/configs/cvrp.toml
+++ b/configs/cvrp.toml
@@ -1,0 +1,54 @@
+node_ops = [
+    "Exchange10",
+    "Exchange20",
+    "Exchange30",
+    "Exchange11",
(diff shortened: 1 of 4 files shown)


N-Wouda:
I get that these need to exist somewhere, but must that be on `main`? I can imagine we make a branch `paper` that stores these (similar to how JOSS worked).

leonlan:
We need them to be visible somewhere since we're publishing the benchmarks on the docs https://pyvrp.readthedocs.io/en/latest/setup/benchmarks.html. Specifically for CVRP, how will users now know which parameter values to obtain best performance for that?

leonlan:
Instead of a `paper` branch we could make a `benchmark` branch for this purpose?

PR #165, opened 2023-03-03
Individual.make_random

### LINKED ISSUE #153: Move the random init on Individual into a classmethod `make_random`
Right now we support two `__init__` methods on `Individual`, with the following signatures:
```python
    def __init__(
        self,
        data: ProblemData,
        penalty_manager: PenaltyManager,
        rng: XorShift128,
    ) -> None:

    def __init__(
        self,
        data: ProblemData,
        penalty_manager: PenaltyManager,
        routes: List[List[int]],
    ) -> None:
```

The first is what I call the "random init", the second the "route init". Having multiple overloaded `__init__` methods really screws up typing and generating docs, as well as IDE support. We can do better to express the intent of the random init by having it bound as a classmethod instead.

I propose something like this:
```python
    def __init__(
        self,
        data: ProblemData,
        penalty_manager: PenaltyManager,
        routes: List[List[int]],
    ) -> None:

   @classmethod
   def make_random(
        cls,
        data: ProblemData,
        penalty_manager: PenaltyManager,
        rng: XorShift128,
    ) -> Individual:
``` 
This should not require changes on the C++ side, other than in the bindings. Intended usage on the Python side is `individual = Individual.make_random(data, pm, rng)`.

N-Wouda:
This PR closes #153. It:
- Moves the random constructor into a classmethod `Individual.make_random` on the Python side;
- Runs clang-format on the whole thing because apparently there's some small stuff that it missed.
- Updates Python-side code to use `Individual.make_random`.

### DIFF
--- a/pyvrp/cpp/Individual_bindings.cpp
+++ b/pyvrp/cpp/Individual_bindings.cpp
@@ -25,8 +25,9 @@ PYBIND11_MODULE(_Individual, m)
                 return py::cpp_function(  // See issue 1693 in the pybind repo.
                     [](ProblemData const &data,
                        PenaltyManager const &penaltyManager,
-                       XorShift128 &rng)
-                    { return Individual(data, penaltyManager, rng); },
+                       XorShift128 &rng) {
+                        return Individual(data, penaltyManager, rng);
+                    },
                     py::arg("data"),
                     py::arg("penalty_manager"),
                     py::arg("routes"),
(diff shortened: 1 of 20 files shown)


pyvrp/cpp/Individual_bindings.cpp
```diff
@@ -44,14 +54,13 @@ PYBIND11_MODULE(_Individual, m)
             },
             py::arg("memo"))
         .def("__hash__",
-             [](Individual const &individual)
-             { return std::hash<Individual>()(individual); })
+             [](Individual const &individual) {
+                 return std::hash<Individual>()(individual);
+             })
```
leonlan, line 59: I’ve had some issues like this with the clang formatting. It seems that pre-commit doesn’t catch formatting differences between newer versions clang-format and the one that’s in the pre-commit configuration. 

Should we bump the clang-format version to the newest one?

leonlan (approved):

pyvrp/cpp/Individual_bindings.cpp
N-Wouda: I'll open an issue for this. We definitely need to figure out why the pre-commit's clang-format seems to be letting stuff like this go through.

N-Wouda: See #166.

PR #167, opened 2023-03-04
Remove args from clang-format in pre-commit

### LINKED ISSUE #166: Check why pre-commit's clang-format does not seem to format everything the same way
See #165 for details. Basically: clang-format when run on its own over the codebase catches things the same clang-format in the pre-commit does not. Why?

leonlan:
This PR closes #166.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,4 +31,3 @@ repos:
     rev: v11.1.0
     hooks:
       - id: clang-format
-        args: [ "-style=file", "--dry-run" ]
--- a/pyvrp/cpp/Individual_bindings.cpp
+++ b/pyvrp/cpp/Individual_bindings.cpp
@@ -25,8 +25,9 @@ PYBIND11_MODULE(_Individual, m)
                 return py::cpp_function(  // See issue 1693 in the pybind repo.
                     [](ProblemData const &data,
                        PenaltyManager const &penaltyManager,
-                       XorShift128 &rng)
-                    { return Individual(data, penaltyManager, rng); },
+                       XorShift128 &rng) {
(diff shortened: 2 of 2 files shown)


leonlan:
CI indeed now fails due to clang-formatting issues https://github.com/N-Wouda/PyVRP/actions/runs/4330215575/jobs/7561301668.

N-Wouda:
The downside of this is that it doesn't show _what_ needs to be changed in the CI. Is there a way we can get it to both format, but also show what it did in output? (not sure)

leonlan:
> The downside of this is that it doesn't show _what_ needs to be changed in the CI. Is there a way we can get it to both format, but also show what it did in output? (not sure)

I looked for this, but it seems that we can't have both. Either we have a dry run which shows the changes necessary, or we have the non-dry-run which will make the file changes in place. I prefer the latter, because it's easier for contributors to just run pre-commit to get the formatting right. Otherwise, they will need a separate clang-format installed with the right version to make the formatting changes.

N-Wouda:
> I looked for this, but it seems that we can't have both. Either we have a dry run which shows the changes necessary, or we have the non-dry-run which will make the file changes in place. I prefer the latter, because it's easier for contributors to just run pre-commit to get the formatting right. Otherwise, they will need a separate clang-format installed with the right version to make the formatting changes.

What I don't like about this is that we then force new contributors to also set-up the pre-commit workflow locally. It's not a lot of work if you know what it is, but when you're very new to this stuff everything's hard. The problem also exists for black, which doesn't really say what's wrong either. I'd prefer tools that you could run locally, but if that's not done, then at least the CI should explain what's wrong.

Let's merge this in for now, but keep in mind that we should really improve the CI on this aspect.

Issue #157, opened 2023-03-01
Reduce coupling


N-Wouda:
A lot of things depend on a lot of things right now. Sometimes that's unavoidable, but there are also definitely cases where it kind of grew together that way (e.g., we found a recent case in #152). This issue serves two goals:

1. Determine how each class and function depends on the other classes and functions;
2. Decide which of these are OK/desirable, and which of these we should try to get rid of.

N-Wouda:
`Individual` stores a pointer to a `ProblemData` instance, but that is not actually needed: it's only passed to the constructor for initialisation (similarly to how `XorShift128` is used there). After that it's unnecessary.

N-Wouda:
Similarly, `TimeWindowSegment` stores a pointer to the distance matrix. But that's only needed when it's merged with other TWS, and can be passed in at that point.

N-Wouda:
#162 addresses the above-mentioned two cases. The following is a list of classes that depend on other classes. I do not consider cases where the dependency is transient (e.g., just a function argument but no state), but only those where an internal reference to an external object is kept within the class - the dependency has to be part of the _state_ of the class.

With that criterion, we have the following in `pyvrp`:

- GeneticAlgorithm: depends on ProblemData, PenaltyManager, XorShift128, Population, LocalSearch, GeneticAlgorithmParams
- Individual: depends on PenaltyManager
- Matrix: none
- PenaltyManager: none
- Population: depends on ProblemData, PenaltyManager, XorShift128, PopulationParams
- ProblemData: none
- Result: depends on Individual and Statistics
- Statistics: none
- SubPopulation: depends on PopulationParams, which is owned by Population
- TimeWindowSegment: none
- XorShift128: none

The `pyvrp.crossover` and `pyvrp.diversity` modules define only stateless functions. We might want to have a look at their function signatures at a later point in time as well.

In `pyvrp.educate`:

- All operators: ProblemData, PenaltyManager
- LocalSearch: ProblemData, PenaltyManager, XorShift128
- Node: ~ProblemData (why?)~ EDIT: removed in #162
- Route: ProblemData

And that's about it.

wouterkool:
I am a big fan of improving this. Some additional thoughts:
* I think naming and decoupling of things also relates to the overal philosophy of PyVRP, I have made #164 for this.
* Regarding PenaltyManager, see #100 which will help decouple things. Major question is how to deal with the population if costs change due to changed penalties.
[...]

Issue #156, opened 2023-02-28
Run local search on initial solutions


wouterkool:
The original HGS implementation runs the local search on the initial solutions before inserting them into the population. We don't do this, and as a result, parents involved in the crossover are not locally optimal in the first phase of the algorithm, which may affect the crossover and local search runtime of the offspring. Should we test whether it is an improvement to educate the initial individuals before inserting them in the initial population?

N-Wouda:
I only have Euro/NeurIPS results to back this up with, but at least on those instances there was no meaningful difference between starting with random individuals, or starting with a population of already-educated individuals. That's why we took it out. But I am not sure if that result generalises.

wouterkool:
At least its helpful to know this was an informed decision :). I think it may be worth reconsidering, if not for performance perhaps also for consistency (i.e. making the population always contain locally optimally individuals).

N-Wouda:
From the paper:
![image](https://user-images.githubusercontent.com/16272507/222096123-781d6f68-26ba-4923-bae0-d0d0b1ff2959.png)
This doesn't say everything, but just looking at the diversity on a GH1000 instance suggests that after, say, 200 iterations there isn't much difference compared to later iterations. At that point we've had a few purges, and those should be pretty good at weeding out really bad random solutions.

>I think it may be worth reconsidering, if not for performance perhaps also for consistency (i.e. making the population always contain locally optimally individuals).

How should we support this? It seems strange to me to have the Population know about or depend on the educator/local search, like in HGS-CVRP or -VRPTW. An alternative might be to require the population be seeded with a set of initial solutions constructed outside of the population. That'd also clean up a lot of the other dependencies the population has, so I was already in favour of this, but that would push more work to the user who then also has to think about seeding the population (instead of that happening by default).

leonlan:
#168 introduced an `initial_solutions` argument to the GeneticAlgorithm. This way the user is now responsible for passing the desired initial solutions to the algorithm. The initial solutions are used again when restarting the population, but I don't think that's an issue since the other parts of the algorithm introduce randomness.

N-Wouda:
> #168 introduced an `initial_solutions` argument to the GeneticAlgorithm. This way the user is now responsible for passing the desired initial solutions to the algorithm. The initial solutions are used again when restarting the population, but I don't think that's an issue since the other parts of the algorithm introduce randomness.

[...]

PR #168, opened 2023-03-04
Reduce coupling in Population and GA now takes initial solutions

### LINKED ISSUE #71: Construct Population with an initial pool of solutions
Rather than having the Population initialise its own individuals.

leonlan:
This PR:
- Population no longer depends on problem data, penalty manager and Xorshift.
- Closes #71: initial population solutions are no longer initialised in Population itself.
- GeneticAlgorithm now takes as `initial_solutions` argument, which is required to be non-empty.


leonlan:
@N-Wouda This is a proposal to close #71, to address the discussion in #156, and to address coupling in Population #157. I have not yet changed tests and notebooks, which I'll do if we decide to proceed with these changes.

pyvrp/Population.py
```diff
@@ -15,38 +13,34 @@ class Population:
...
-        data: ProblemData,
-        penalty_manager: PenaltyManager,
         rng: XorShift128,
         diversity_op: Callable[[Individual, Individual], float],
+        initial_solutions: List[Individual],
         params: PopulationParams = PopulationParams(),
     ):
-        self._data = data
-        self._pm = penalty_manager
         self._rng = rng
```
leonlan, line 34: We can also remove the RNG dependency and pass it as argument to `self.select`/`self.binary_tournament`. I don't know if that's desirable.

pyvrp/Population.py
```diff
@@ -138,14 +132,14 @@ def select(self) -> Tuple[Individual, Individual]:
...
+        Restarts the population. All current individuals are removed and the
+        initial individuals used to seed the restarted population.
         """
         self._feas = SubPopulation(self._op, self._params)
         self._infeas = SubPopulation(self._op, self._params)
 
-        for _ in range(self._params.min_pop_size):
-            self.add(Individual.make_random(self._data, self._pm, self._rng))
+        for indiv in self._initial_solutions:
+            self.add(indiv)
```
N-Wouda, line 142: Do you know if this impacts performance? I don't really have a good intuition for this.

pyvrp/Population.py
N-Wouda: I think it's better because there's less internal state on `Population` then.

pyvrp/Population.py
N-Wouda: I'd assume no, but it might not hurt to make sure.

leonlan: I'll run a benchmark just to be sure.

leonlan:
Benchmark: GH1000, 3600s, 1 seed

[3f14e09](https://github.com/N-Wouda/PyVRP/pull/168/commits/3f14e09a63bf8cd7cfcc5070e344a3bba8ac6b1c)
```
Cost:  333780.9
Iters: 102301.9
```

main
```
TODO
```

wouterkool:
Shouldn't we use a different set of initial solutions when restarting?

leonlan:
> Shouldn't we use a different set of initial solutions when restarting?

I don't think it matters. The binary selection, crossovers, and local search introduce randomness. So even if we restart with the same initial population, the resulting trajectory will be different each time.

leonlan:
This PR is ready to be reviewed. Meanwhile, I'll run a 10 seed benchmark to check that there is no performance drop with the new change (restart with the same individuals).

N-Wouda:
I'll try to have a look at this tonight.

pyvrp/Population.py
```diff
@@ -131,26 +126,32 @@ def select(self) -> Tuple[Individual, Individual]:
...
 
         return first, second
 
     def restart(self):
         """
-        Restarts the population. All individuals are removed and a new initial
-        population population is generated.
+        Restarts the population. All current individuals are removed and the
+        original initial individuals are used agains to initialize the
+        restarted population.
```
N-Wouda, line 138: Typo: agains.

Can we rewrite this a bit to really clarify that we're restarting from the initial individuals that were passed when this object was initialised? It's not clear enough to me what "original initial individuals" means.

pyvrp/Population.py
```diff
@@ -15,12 +13,8 @@ class Population:
...
     Parameters
     ----------
-    data
-        Data object describing the problem to be solved.
-    penalty_manager
-        Penalty manager to use.
-    rng
-        Random number generator.
+    initial_solutions
+        List of individuals used to initialize the population.
```
N-Wouda, line 17:
```suggestion
        List of individuals used to initialise the population.
```

pyvrp/Population.py
```diff
@@ -30,23 +24,19 @@ class Population:
...
+        self._initial_solutions = initial_solutions
         self._params = params
 
         self._feas = SubPopulation(diversity_op, params)
         self._infeas = SubPopulation(diversity_op, params)
 
-        for _ in range(params.min_pop_size):
-            self.add(Individual.make_random(data, penalty_manager, rng))
+        for indiv in initial_solutions:
+            self.add(indiv)
```
N-Wouda, line 39:
```suggestion
        for individual in initial_solutions:
            self.add(individual)
```

pyvrp/Population.py
```diff
@@ -131,26 +126,32 @@ def select(self) -> Tuple[Individual, Individual]:
...
+        original initial individuals are used agains to initialize the
+        restarted population.
         """
         self._feas = SubPopulation(self._op, self._params)
         self._infeas = SubPopulation(self._op, self._params)
 
-        for _ in range(self._params.min_pop_size):
-            self.add(Individual.make_random(self._data, self._pm, self._rng))
+        for indiv in self._initial_solutions:
+            self.add(indiv)
```
N-Wouda, line 144:
```suggestion
        for individual in self._initial_solutions:
            self.add(individual)
```

pyvrp/plotting/tests/test_plotting.py
```diff
@@ -52,9 +59,13 @@ def test_plot_result():
...
     bks = read_solution("data/RC208.sol")
+
     pm = PenaltyManager(data.vehicle_capacity)
     rng = XorShift128(seed=42)
-    pop = Population(data, pm, rng, broken_pairs_distance)
+    params = PopulationParams()
+    init = make_random_solutions(params.min_pop_size, pm, rng, data)
+
+    pop = Population(init, broken_pairs_distance, params=params)
     stats = Statistics()
```
N-Wouda, line 69: The spacing here doesn't make much sense to me. I'd expect `params`, `init`, and `pop` to all be together in the same paragraph (since they're all part of "population stuff").

pyvrp/tests/test_Population.py
```diff
@@ -257,12 +277,13 @@ def test_elite_individuals_are_not_purged(nb_elite: int):
 
 
 def test_binary_tournament_ranks_by_fitness():
-    data = read("data/RC208.txt", "solomon", "dimacs")
+    data = read("data/RC208LessVehicles.txt", "solomon", "dimacs")
```
N-Wouda, line 280: Why does this need a different (new) instance?

pyvrp/tests/data/RC208LessVehicles.txt
```diff
@@ -0,0 +1,110 @@
+RC208
```
N-Wouda, line 1: If we can, I'd like to avoid introducing new test data.

pyvrp/tests/test_Population.py
```diff
@@ -177,24 +178,43 @@ def test_select_returns_same_parents_if_no_other_option():
...
 
-    old = {individual for individual in pop}
     pop.restart()
-    new = {individual for individual in pop}
 
-    assert_equal(len(pop), params.min_pop_size)
-    assert_equal(len(old & new), 0)  # no old pops survived the restart
+    # Check again for the restarted population.
+    for indiv in pop:
+        assert_(np.any(indiv == other for other in init))
```
N-Wouda, line 200: I don't really get why this needs to be changed. Would something like:
```python
init = set(...)
current = {individual for individual in pop}
assert_(len(current & init), params.min_pop_size)

pop.restart()

new = {individual for individual in pop}
assert_(len(new & init), params.min_pop_size)
```
not work?

pyvrp/tests/test_Population.py
```diff
@@ -177,24 +178,43 @@ def test_select_returns_same_parents_if_no_other_option():
...
+
+
+@mark.parametrize("num_init", [1, 15, 21, 30, 100])
+def test_num_initial_solutions(num_init):
+    data = read("data/OkSmallFeasible.txt")
```
N-Wouda, line 205: Do we need a new test data instance for this?

pyvrp/tests/helpers.py
```diff
@@ -37,6 +38,13 @@ def sleep(duration, get_now=time.perf_counter):
         now = get_now()
 
 
+def make_random_solutions(num_sols, pm, rng, data):
+    """
+    Returns a list of a ``num_sols`` random solutions.
+    """
+    return [Individual.make_random(data, pm, rng) for _ in range(num_sols)]
```
N-Wouda, line 45: Can we have the order of the arguments here be the same as they are for almost every other thing: `data, pm, rng`? I'm fine with `num_sols` coming first.

pyvrp/Population.py
```diff
@@ -30,23 +24,19 @@ class Population:
 
     def __init__(
         self,
-        data: ProblemData,
-        penalty_manager: PenaltyManager,
-        rng: XorShift128,
+        initial_solutions: List[Individual],
```
N-Wouda, line 27: Can we make the type here a general iterable? It doesn't necessarily have to be a list.

pyvrp/tests/data/OkSmallFeasible.txt
```diff
@@ -0,0 +1,41 @@
+NAME : TEST
```
N-Wouda, line 1: If we can, I'd like to avoid introducing new test data.

pyvrp/tests/helpers.py
```diff
@@ -37,6 +38,13 @@ def sleep(duration, get_now=time.perf_counter):
         now = get_now()
 
 
+def make_random_solutions(num_sols, pm, rng, data):
+    """
+    Returns a list of a ``num_sols`` random solutions.
```
N-Wouda, line 43:
```suggestion
    Returns a list of ``num_sols`` random solutions.
```

pyvrp/tests/test_Population.py
```diff
@@ -177,24 +178,43 @@ def test_select_returns_same_parents_if_no_other_option():
...
+
+    params = PopulationParams(min_pop_size=0, generation_size=10)
+    init = make_random_solutions(num_init, pm, rng, data)
+    pop = Population(init, bpd, params)
+
+    # If there are more initial individuals than the maximal population size
+    # allows, then a purge is triggered. This resulting population size is
+    # equal to ``num_init`` modulo the number individuals that need to
+    # be added to trigger a purge (11).
+    assert_equal(len(pop), num_init % 11)
```
N-Wouda, line 217: This isn't true in general. Suppse we add `15` individuals. Then `15 % 11 == 4`. But what if we've added 7 feasible and 8 infeasible solutions? Then no purge would have been triggered because neither subpopulation would have been of maximal size.

The reason this test passes is that all solutions are feasible in this instance, which is not explicitly mentioned in the test. Can we document that much more explicitly?

pyvrp/Population.py
N-Wouda: I think I'd make sense for `initial_solutions` to be optional. By default we construct an empty population. This'd be possible if we did, for example,
```python
def __init__(
  self,
  diversity_op,
  initial_solutions: Iterable[Individual] = tuple(),
  params: PopulationParams = PopulationParams()
)
```

pyvrp/tests/test_GeneticAlgorithm.py
```diff
@@ -92,17 +93,21 @@ def test_params_constructor_does_not_raise_when_arguments_valid(
...
 
-    node_op = Exchange10(data, pen_manager)
+    pop_params = PopulationParams()
+    init = make_random_solutions(pop_params.min_pop_size, pm, rng, data)
+    pop = Population(init, bpd, params=pop_params)
+
+    ls = LocalSearch(data, pm, rng, compute_neighbours(data))
+
+    node_op = Exchange10(data, pm)
     ls.add_node_operator(node_op)
```
N-Wouda, line 105: While we're here anyway:

```suggestion
    ls = LocalSearch(data, pm, rng, compute_neighbours(data))
    ls.add_node_operator(Exchange10(data, pm))
```

pyvrp/tests/test_Population.py
```diff
@@ -143,15 +144,15 @@ def test_select_returns_same_parents_if_no_other_option():
     rng = XorShift128(seed=2_147_483_647)
 
     params = PopulationParams(min_pop_size=0)
-    pop = Population(data, pm, rng, broken_pairs_distance, params)
+    pop = Population([], bpd, params=params)
```
N-Wouda, line 147: I was thinking about whether we should make `params` a keyword only argument. Perhaps yes, but then not in this PR (since we should then probably enforce it all at once for every Params-like object).

N-Wouda (commented):
A bunch of small things and some questions. On the whole this looks good to me!

pyvrp/tests/test_Population.py
leonlan: Because the test assumes that the randomly generated individuals are all infeasible, which was not the case when I ran the test with the original file.

pyvrp/Population.py
leonlan: I don't think that making the initial solution argument is desirable. Creating an empty Population is not a good default, and it also doesn't work well with the restarting mechanism.

pyvrp/tests/test_Population.py
N-Wouda: But we can just filter the feasible solutions away before passing `init` to `pop`, right?

pyvrp/Population.py
N-Wouda: >Creating an empty Population is not a good default

Why not? That Population should not be empty is a problem for the GeneticAlgorithm, not for the Population container itself.

>and it also doesn't work well with the restarting mechanism.

This I don't understand. Restart would just return to an empty population again; I don't see what'd go wrong.

N-Wouda: Thinking about this, we should probably enforce in the GA that the population has least 1/2 (I think 2?) individuals, right now it can be empty.

N-Wouda: I'll add a PR for that.

pyvrp/tests/test_Population.py
leonlan: No idea why I changed this - I changed it again to what you suggested!

leonlan:
I'll address the remaining comments tomorrow morning.

pyvrp/GeneticAlgorithm.py
```diff
@@ -73,20 +75,27 @@ def __init__(
...
+        self._initial_solutions = initial_solutions
         self._params = params
 
-        self._best = Individual.make_random(data, penalty_manager, rng)
+        # Initialise best using the feasible initial solution with lowest cost,
+        # or the initial solution with lowest cost if none are feasible.
+        self._best = min(
+            self._initial_solutions,
+            key=lambda indiv: (not indiv.is_feasible(), indiv.cost()),
+        )
```
N-Wouda, line 98: Here, and in a test: this is a bit too clever IMHO. I had to think for too long about why this works, so its readability could be improved.

Something like
```python
# Find best feasible initial solution if any exist, else set lower cost
# solution as initial best.
init_feas = [indiv for indiv in initial_solutions if indiv.is_feasible()]
sols = init_feas if init_feas else initial_solutions
self._best = min(sols, key=lambda indiv: indiv.cost())
```
is barely longer, but doesn't make me think as hard.

In the test I think we can add some feasible individual that we know is the best, and then test whether that individual is indeed set as the best. There's no need to reproduce the logic there.

N-Wouda (approved):
LGTM! A few small things about how these changes can be made a bit more legible, but nothing that's blocking merging this in.

leonlan:
> LGTM! A few small things about how these changes can be made a bit more legible, but nothing that's blocking merging this in.

OK, I addressed all comments. Thanks for all the help! I'll run a benchmark before merging this PR.

leonlan:
GH1000, 10 seeds, 36000s
```
min     33341.5
mean    33359.6
max     33374.5
```
All looks fine, so I'm merging this in.

Issue #60, opened 2023-02-03
Rename distance_matrix to (travel_)duration_matrix


N-Wouda:
Hadn't really given this much thought, but everything's of course in time units, so we shouldn't talk about a "distance matrix" so much. We need to update a few places in the code to reflect this.

N-Wouda:
>I haven’t seen this before in any of the existing VRPLIB instances though.

That makes sense to me. The typical VRPLIB instances are quite stylised, so I'd expect this kind of stuff not to be present. On our end, there's nothing stopping us from writing a ProblemData class that can accept two matrices (cost + time), and another 'shortcut' constructor that simply copies one into the other, when only one has been provided.

wouterkool:
* I don't think there is a VRPLIB instance having both distance and time, and if it supports two matrices I don't think its unambiguous. Also note there is distance, time and (in theory) costs, which may be a combination of distance and time or even completely something else (general support of objectives is another topic anyway). You can replace distance by cost except if you want to support constraints like maximum route distance. Anyway, I don't think we should let our support depend on the VRPLIB format.
* For now we can either choose to just support duration, or do both, indeed by default being equal. However we should make sure having two matrices does not affect performance?

Issue #61, opened 2023-02-03
Improve interop between numpy and our own matrix class


N-Wouda:
And maybe also have that replace a few of the from/to Python lists conversions? E.g., return numpy arrays instead of `std::vector` here and there.

N-Wouda:
I think this is quite important to set up soon.

N-Wouda:
We can probably provide methods to make our Matrix class work efficiently with numpy arrays. That'd already bring a lot of the benefit. [This](https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#numpy) is relevant here.

N-Wouda:
(at that point we can also bind the `distance_matrix` and `duration_matrix` to Python again, because the conversion is just to some numpy array)

N-Wouda:
This also seems relevant: https://github.com/tdegeus/pybind11_examples.

PR #169, opened 2023-03-04
Add support for double precision

### LINKED ISSUE #47: Compute using double precision
Currently the code uses integers for accumulating distances/times. To comply with academic standards we want to support full double precision. To keep the code somewhat flexible to be compiled with integer precision if desired, therefore use a (globally defined) general type `TPrecision`/`TTimeDistance` (better suggestions?) that can be either `int` or `double`. For now, we decided against making the code templated to avoid many classes with template parameters. TODO how do we deal with tolerance when doing comparisons using floating points (this should be considered at 1 or 2 places).

N-Wouda:
Closes #47. This PR adds support for floating point values. As part of that, it introduces a few new types: `cost_type` (expressing something's part of the objective), `distance_type` (related to the travel distance), and `duration_type` (pertaining to time window stuff). Distances are used for the objective; durations for feasibility. This is a start to separating distance and travel duration computations as discussed in #5, and also relates to #60.

### Benchmarks

This branch (8cb1d46cf468482707017315844a9cd9c7c851b2) with int precision [job ID 28741902, seeds 1-10]:
```
 Avg. objective: 333634.9 (min: 333469, max: 333860)
Avg. iterations: 106675.2 (min:  97578, max: 113489)
```

This branch (8cb1d46cf468482707017315844a9cd9c7c851b2) with float precision [job ID 28742029, seeds 1-10]:
```
 Avg. objective: 333758.2 (min: 333540, max: 333946)
Avg. iterations:  91390.4 (min:  83910, max: 103511)
```
Note that both these runs are with the default parameters and operators - I have not used the configurations available in `configs/`.


N-Wouda:
I'm going to run two benchmarks of this, one with ints and the other with floats. If performance is the same for both, I suggest we just port everything to floats.

N-Wouda:
> I'm going to run two benchmarks of this, one with ints and the other with floats. If performance is the same for both, I suggest we just port everything to floats.

Benchmarks in first comment. Using integers rather than `double` seems to be some 15% faster. That's a fairly respectable difference, so it might be worthwhile keeping both ints and floats around.

N-Wouda:
I've moved type definitions into a single header and added simple routines for number comparisons, including tests to make sure they work as expected.

pyvrp/cpp/ProblemData.h
```diff
@@ -47,12 +48,27 @@ class ProblemData
...
+
+    /**
+     * Returns the travel duration between the indicated two clients.
+     *
+     * @param first First client.
+     * @param second Second client.
+     * @return duration to travel from the first to the second client.
+     */
+    [[nodiscard]] inline duration_type duration(size_t first,
+                                                size_t second) const;
```
N-Wouda, line 61: Shorten this to `dur`?

N-Wouda: Could also do `time`, or `travel`? Since we call this a lot I'm looking for something less verbose.

pyvrp/cpp/TimeWindowSegment.h
```diff
@@ -61,7 +64,7 @@ TimeWindowSegment TimeWindowSegment::merge(Matrix<int> const &dist,
 }
 
 template <typename... Args>
-TimeWindowSegment TimeWindowSegment::merge(Matrix<int> const &dist,
+TimeWindowSegment TimeWindowSegment::merge(Matrix<distance_type> const &dist,
```
N-Wouda, line 67: Should this be the duration matrix?

pyvrp/tests/test_TimeWindowSegment.py
```diff
@@ -32,7 +32,7 @@ def test_merge_multiple():
     tws2 = TimeWindowSegment(1, 1, 0, 0, 3, 6)
     tws3 = TimeWindowSegment(2, 2, 0, 0, 2, 3)
 
-    mat = IntMatrix([[1, 4, 1], [1, 2, 4], [1, 1, 1]])
+    mat = DoubleMatrix([[1, 4, 1], [1, 2, 4], [1, 1, 1]])
```
N-Wouda, line 35: This does not pass when compiling in integer mode, but I'm OK with that. Once we implement #61, the issue is moot anyway.

pyvrp/cpp/ProblemData.h
wouterkool: I think `duration` is still best as it's clear that it specifies an interval. In that sense, `time` is more ambiguous as it can also refer to a point in `time` (and is used like that in the code). `travel` can mean anything, i.e. `travel_(dist|cost|time)` so I would not use that. If it's that important that it's short, I'd say go with `dur`, it's not immediately clear but also not ambiguous, but I prefer to keep it as is.

pyvrp/cpp/TimeWindowSegment.h
wouterkool: I agree, maybe it was like that since there is no separate distance/duration yet? (Did I implement it? Don't remember)

pyvrp/cpp/ProblemData.h
```diff
@@ -82,22 +98,29 @@ class ProblemData
...
-                std::vector<int> const &servDurs,
-                std::vector<std::vector<int>> const &distMat);
+    ProblemData(
+        std::vector<std::pair<int, int>> const &coords,
+        std::vector<int> const &demands,
+        size_t numVehicles,
+        size_t vehicleCap,
+        std::vector<std::pair<duration_type, duration_type>> const &timeWindows,
+        std::vector<duration_type> const &servDurs,
+        std::vector<std::vector<duration_type>> const &distMat);
```
wouterkool, line 108: Shouldn't `distMat` have `distance_type`? (They're the same now anyway)

pyvrp/cpp/ProblemData.h
```diff
@@ -82,22 +98,29 @@ class ProblemData
...
 };
 
 ProblemData::Client const &ProblemData::client(size_t client) const
 {
     return clients_[client];
 }
 
-int ProblemData::dist(size_t first, size_t second) const
+distance_type ProblemData::dist(size_t first, size_t second) const
```
wouterkool, line 116: Maybe a stupid question and not related to this change but can we `inline` this?  Same for some of the other methods in this class like `client` which get called a lot?

pyvrp/cpp/crossover/crossover.cpp
```diff
@@ -8,35 +9,35 @@ namespace
...
+    auto const clientEarliestFinish
         = clientEarliestArrival + data.client(client).serviceDuration;
-    int distClientNext = data.dist(client, next);
-    int nextLate = data.client(next).twLate;
+    auto const nextLate = data.client(next).twLate;
 
-    if (clientEarliestFinish + distClientNext >= nextLate)
-        return INT_MAX;
+    if (clientEarliestFinish + data.duration(client, next) >= nextLate)
+        return cost_type(INT_MAX);
```
wouterkool, line 37: Wouldn't it be better to define a constant `COST_MAX` of the right type? Now it relies on doubles being able to represent correctly INT_MAX.

pyvrp/cpp/educate/Node.h
```diff
@@ -16,9 +17,9 @@ class Node
     Route *route;     // Pointer towards the associated route
 
     // TODO can these data fields be moved to Route?
-    int cumulatedLoad;              // Load from depot to client (inclusive)
-    int cumulatedDistance;          // Distance from depot to client (inclusive)
-    int cumulatedReversalDistance;  // Distance if (0 .. client) is reversed
+    int cumulatedLoad;                // Load from depot to client (incl)
+    distance_type cumulatedDistance;  // Distance from depot to client (incl)
+    distance_type cumulatedReversalDistance;  // Reverse distance of (0..client)
```
wouterkool, line 22: Does this specify the `delta` of the distance between the forward and backward direction, or the actual distance of the reversed segment? From my head it was the first, but the comment now reads as if it's the second. Also, if this is inclusive, maybe use `[0, ..., client]` to indicate that?

pyvrp/cpp/educate/Route.cpp
```diff
@@ -136,7 +136,7 @@ void Route::update()
     isLoadFeasible_ = static_cast<size_t>(load_) <= data.vehicleCapacity();
 
     timeWarp_ = nodes.back()->twBefore.totalTimeWarp();
-    isTimeWarpFeasible_ = timeWarp_ == 0;
+    isTimeWarpFeasible_ = equal(timeWarp_, duration_type(0));
```
wouterkool, line 139: I'm not sure if we want/need a tolerance here? Doubles can represent exact 0 and the timewarp will be zero if all arrivals are before the end of the time windows. I'm not sure if we need a tolerance there

pyvrp/tests/test_TimeWindowSegment.py
wouterkool: I'm fine with that, however could we somehow make a way to inspect from Python whether PyVRP was compiled in integer or double precision mode? That may be helpful to the user and then we could also easily make this test work.

wouterkool (approved):
Looks good, I think it's very nice that the code is now explicit on what are times, distances or costs rather than all of them being doubles. Some questions/comments
* Will users be able to select double/integer when installing through PyPI? Or will they need to install locally by cloning the repo?
* I think it's good that this code is prepared to support a separate duration/distance matrix, but I'm not sure if it's weird to have it in the public api (the `duration` and `durationMatrix` functions) without actually supporting it? We could easily add it now, however:
    * It makes sense to me to be able to construct a ProblemData either with only a distance or both a distance and a duration matrix. As I think we want only one constructor in the bindings, we should probably implement a `make_problem_data` function, wrap the clas in Python or use `std::optional` in the bindings?
    * Will it hurt performance if we use two separate matrices, since we may look up duration and distance separately? If so, we could consider making that a compiler option also, such that it can be optimized if `duration` and `distance` are inlined and the same.
* Is there a way in which we can make times and distances proper types (Scalar types?) such that we can check that we only add distances to distances, durations/times to times and perhaps need an explicit/automatic cast when converting to costs. This way we could prevent implementation errors where we accidentally mix durations and distances.

leonlan (commented):
I don't have much to say about this - Wouter already covered my questions (and way more) :-).

pyvrp/cpp/ProblemData.h
N-Wouda: These are already inlined - you're looking at the implementation; `inline` is part of the declaration above (in the class).

pyvrp/cpp/educate/Route.cpp
N-Wouda: When both are exactly zero the comparison is exact; see the implementation of `equal()`. 

A problem would be when an arrival happens at about the end of a time window, say at `1.00001` for a time window of `[0, 1]`, we'd have `0.00001` time warp. I think that's just a rounding error, but the route would then not be considered time warp feasible because that value is not zero. To avoid such cases I think we should have a tolerance here.

pyvrp/cpp/TimeWindowSegment.h
N-Wouda: I don't know what caused it; I just noticed it when reading through the PR!

pyvrp/cpp/ProblemData.h
N-Wouda: Works for me. I've mostly kept `duration`, expect one or two places where I needed something slightly shorter and used `dur`.

pyvrp/cpp/Individual.cpp
```diff
@@ -23,11 +23,11 @@ void Individual::evaluate(ProblemData const &data)
...
         numRoutes_++;
 
-        int routeDist = data.dist(0, route[0]);
-        int routeTimeWarp = 0;
+        distance_type routeDist = data.dist(0, route[0]);
+        duration_type routeTimeWarp = 0;
         int routeLoad = data.client(route[0]).demand;
 
-        int time = routeDist;
+        duration_type time = routeDist;
```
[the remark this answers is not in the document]
N-Wouda: Good catch, thanks! I think we do assume we start at $t=0$. I'll update that to `data.depot().twEarly` instead.

pyvrp/cpp/ProblemData.h
N-Wouda: Yes! Funnily enough this was already OK in the bindings, but not yet updated here.

pyvrp/cpp/crossover/crossover.cpp
N-Wouda: I've replaced this with `std::numeric_limits<cost_type>::max()`.

pyvrp/cpp/Individual.cpp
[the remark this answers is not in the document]
N-Wouda: We could require that the depot's opening time window is always zero, but starting from `data.depot().twEarly` is also not too difficult. Should we assume the depot has no service time?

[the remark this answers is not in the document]
N-Wouda: (I think yes, along with that the depot has no demand)

pyvrp/cpp/Individual.cpp
```diff
@@ -62,11 +62,11 @@ void Individual::evaluate(ProblemData const &data)
         // rDist and time
         routeDist += data.dist(route.back(), 0);
         time += data.client(route.back()).serviceDuration
-                + data.dist(route.back(), 0);
+                + data.duration(route.back(), 0);
 
         // For the depot, we only need to check the end of the time window
         // (add possible time warp)
-        routeTimeWarp += std::max(time - data.depot().twLate, 0);
+        routeTimeWarp += std::max(time - data.depot().twLate, duration_type(0));
```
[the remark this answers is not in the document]
N-Wouda: I did some reading on this and we should prefer `static_cast`. I'm fixing this in the few places I used this functional cast instead.

pyvrp/cpp/educate/Node.h
N-Wouda: It's the $\Delta$, and I've updated the comments/names to reflect this.

pyvrp/_ProblemData.pyi
```diff
@@ -23,12 +23,12 @@ class Client:
         client's (x, y) location tuple.
     """
 
+    x: int
```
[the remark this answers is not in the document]
N-Wouda: Only x, y, and demand are currently always integral. We can update that if we add two more types: `location_type`, and `demand_type`. But I'm not sure if that's worth it.

pyvrp/tests/test_TimeWindowSegment.py
N-Wouda: It's already possible via signature inspection at runtime; that's no problem. But the larger question is how to make these different compilation modes available to the user, and I don't really have ideas about that yet.

wouterkool (approved):

pyvrp/cpp/educate/Route.cpp
wouterkool: Ok, I agree that is a problem if we use doubles to represent integer distances and time windows, which is quite likely. If we actually compute distances using double precision, I think it makes more sense to put the boundary at `1.00000` as you will not exactly arrive at this time anyway, but it does not hurt either so let's keep it this way.

pyvrp/cpp/Individual.cpp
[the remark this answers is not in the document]
wouterkool: I think starting at `data.depot().twEarly` supports some valid (future) use cases like supporting multiple depots with different opening times and I see no downside. Hence I am in favor of supporting it, but we must make sure it is consistent throughout the code, i.e. with the `TimeWindowSegment` implementation, plotting/diagnostics code, etc.

Likewise, I think service time at a depot can be seen as time for loading the vehicle and should be valid, as long as we are clear about the definition. While you may also add the 'loading' time to `twEarly` I think it's better to be explicit and this supports another (future) use case like multiple visits to the depot using a single vehicle.

For demand, I see no meaningful interpretation (except possibly a negative value representing the supply/capacity) so I suggest to keep it 0.

wouterkool:
> Re. non-code questions:
> 
> > Will users be able to select double/integer when installing through PyPI? Or will they need to install locally by cloning the repo?
> 
> For now the latter. We haven't really figured out delivery to PyPI yet for nonstandard compilations (neither here nor for CVRP).

Ok, I think we agreed for maximal performance/nonstandard use cases users should compile from sources anyway.

> 
> > I think it's good that this code is prepared to support a separate duration/distance matrix, but I'm not sure if it's weird to have it in the public api (the duration and durationMatrix functions) without actually supporting it? We could easily add it now, however:
> > 
> > * It makes sense to me to be able to construct a ProblemData either with only a distance or both a distance and a duration matrix. As I think we want only one constructor in the bindings, we should probably implement a make_problem_data function, wrap the clas in Python or use std::optional in the bindings?
> 
> This PR adds very rudimentary support for these things, which doesn't yet have to be bound to Python. I'm not opposed to removing the Python bindings for these so we can avoid this discussion in this PR.

I agree, let's add the bindings when we support it or add a note that it's not yet supported (if it's too much to support it as part of this PR, as I think it should not be that hard?).

> 
> > * Will it hurt performance if we use two separate matrices, since we may look up duration and distance separately? If so, we could consider making that a compiler option also, such that it can be optimized if duration and distance are inlined and the same.
> 
> I'd hesitate against even more compilation flags, since they're difficult to maintain and hard to get right. In this particular case I don't think there is much to worry about in terms of performance. Two matrices will consume a bit more memory than just one, but I've never seen a run that needed more than several MB so I think we're good on that end as well.
 
I agree on the compilation flags but I think it would be good to test the performance. I'm not worried about memory but doing two matrix lookups instead of one, whereas I think they lookups constitute a significant part of the runtime (as optimizing the matrix implementation seems to be worth it).

> > Is there a way in which we can make times and distances proper types (Scalar types?) such that we can check that we only add distances to distances, durations/times to times and perhaps need an explicit/automatic cast when converting to costs. This way we could prevent implementation errors where we accidentally mix durations and distances.
> 
> Not with an easy alias like we do now, unfortunately. We could create our own classes or structs that are thin wrappers around doubles and ints, but then we'd have to implement every operator ourselves as well. This might well be worth doing at some point, but it's a lot of work that I don't really want to do right now. I'll open an issue for this instead. EDIT: #179.
> 
> Also somewhat related to Boost's strong typedef (https://www.boost.org/doc/libs/1_42_0/boost/strong_typedef.hpp).

Agree! I am a bit surprised that this is non-trivial, as I would say this is a very common use case (as every double actually represents some unit that you typically don't want to mix).

pyvrp/_ProblemData.pyi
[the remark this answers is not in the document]
wouterkool: I think it makes sense to make x, y distance type as well (i.e. horizontal/vertical distance from origin), especially when using Euclidean distances?

For demand, we can make it a type if it's easy, but if there is no use case for it let's keep it integral.

wouterkool (approved):

wouterkool:
@N-Wouda I tried to update this to be in sync with the new developments in main. I think we should performance test the following cases:
* main, dimacs objective (baseline)
* Integer precision, dimacs objective (scale by 10 and truncate, should be same as main)
* Double precision, dimacs objective (scale by 10 and truncate to integer values represented as doubles)
* Double precision, truncate distances to 1 decimal (don't scale, use actual non-integer values in computation)

I have tested a bit with the last case and fixed some problems by adding `smaller` and `greater` in `precision.h`, to add some tolerance to avoid non-improvements being evaluated as improvements with delta cost something like `-1e-15` causing infinite loops (so you were right a tolerance is needed).

However, there are many more places were a comparison is used, either explicitly or implicitly (through sorting). I think we have some options:
1. leave as is and benchmark performance
2. add smaller/greater and potential greater_equal, smaller_equal everywhere (maybe also custom comparators for sorting)
3. implement a 'strong' type (like [boost](https://www.boost.org/doc/libs/1_42_0/boost/strong_typedef.hpp)) for distance/duration/cost where we overload the comparison operators to have the same effect as option 2 but prettier code (and better type checks)

I like option 3 as it keeps the comparisons in the code nice and clean and we could potentially guarantee equivalent results between integer/double compilation with the same data. However, I'm not sure if it's good practice overloading comparisons with additional tolerances (motivated by the assumption that the represented data is integral)? Also, do you have suggestions how to best implement this?

Finally, I think it makes sense to add both integer and double versions to the CI tests and add the pyvrp cli to the tests to test a simple run (like the notebooks)... but I'm not sure if we want that for all combinations (python/os) since it's starting to grow large?

N-Wouda:
I'm continuing with this and my other open PRs on Thursday.

N-Wouda:
With #210 merged, there's nothing stopping double support now. However, since this branch is really old (pre explicit types, pre duration/distance separation, etc.), I'll probably do work on double support in a new branch, and copy over the relevant bits.

N-Wouda:
PR #234 will complete this.

PR #174, opened 2023-03-05
Make all params arguments keyword-only

### LINKED ISSUE #171: Make parameters a keyword-only argument
>  I was thinking about whether we should make `params` a keyword only argument. Perhaps yes, but then not in this PR (since we should then probably enforce it all at once for every Params-like object).

_Originally posted by @N-Wouda in https://github.com/N-Wouda/PyVRP/pull/168#discussion_r1125506972_

leonlan:
Closes #171.


N-Wouda:
The PR looks good to me, but I'm not sure if we should force this on users. I'd have liked a bit more discussion about this in #171 before committing to it, because normally keyword-only arguments exist when otherwise a contextless value would be passed, like
```python
def join(*args, glue):
    ...
```
called with `join(..., "_")`. That's much less clear than `join(..., glue="_")`. But in our case, `params` is always a labelled object that's pretty clear about what it is already. So I don't know if we should enforce this.

N-Wouda:
@leonlan @wouterkool are there cases where this was not clear?

leonlan:
> @leonlan @wouterkool are there cases where this was not clear?

I think all cases were all clear. I didn't really think this through, it's indeed somewhat unnecessary to enforce `params=params` everywhere.

wouterkool:
For me it's not about the clarity per se but I like that it improves robustness/backwards compatibility of the code when the signature changes, and consistency/ease of use as the user does not need to know if there are any optional arguments preceding params if we always use the keyword. I don't really see a downside either.

N-Wouda:
>I don't really see a downside either.

Well, for one, it breaks unpacking:
```python
def f(x, *, y, z):
    return x + y + z


args = [1, 2, 3]
f(*args)
```
raises a `TypeError` about the number of positional arguments. This'd be very annoying in higher-up code that passes arguments for different things in programmatically.

wouterkool:
Can we reach a conclusion on this?

For the unpacking use case you're describing, I think it's not uncommon to have both `*args` and `**kwargs`, or just treat params separately, and I don't see that we use this much. Is this downside big enough to not merge this change?

Personally, I'm in favor of merging this. @leonlan @N-Wouda?

leonlan:
I'm undecided about this one. The unpacking thing doesn't look like an issue to me. But the advantage of having better robustness is also not clear to me either, since signature changes will most likely lead to breaking changes anyway (e.g., removal of arguments).

N-Wouda:
I'm personally not in favour because keyword-only arguments are somewhat niche, and I don't think this use case fits the niche they're typically used in. But I don't care enough to argue the point much further. 

I'll leave it up to you to decide what to do with your PR, @leonlan.

leonlan:
We have a pretty clear style for parameters (always the last argument in the signature, a clear `params` argument name, and a default). I think that enforcing the keyword-only argument will confuse (new) users when they want to use non-default parameters. So I don't see any direct advantages of having keyword-only arguments right now, so I'm closing this PR and we can always revisit this again when it's becoming relevant again.

PR #175, opened 2023-03-06
Deterministic C++ implementation of SREX

### LINKED ISSUE #79: Split off SREX into seperate function without rng
The current SREX tests written in #76 rely on the RNG to set the internal variables to some specific value. This makes the tests tied to the specific rng and implementation details.

> Perhaps we should split off SREX's implementation into a separate function that takes `startA` and `nMovedRoutes` as arguments, with a shim around it that keeps the original interface.

_Originally posted by @N-Wouda in https://github.com/N-Wouda/PyVRP/pull/76#discussion_r1103801892_

leonlan:
This PR closes #79. 
- C++ implementation of SREX is now deterministic by taking the start indices and number of routes to move as arguments.
- Python-side SREX wraps around the C++ version and still has the randomized implementation.


leonlan:
Something's wrong in the tests: the CI keeps running for >1 hours. But for some reason the job with gcc 10 and Python 3.8 does complete.

N-Wouda:
🤞

pyvrp/cpp/crossover/selective_route_exchange.cpp
```diff
@@ -12,18 +12,22 @@ Individual selectiveRouteExchange(
...
+    size_t startA,
+    size_t startB,
+    size_t const nMovedRoutes)
 {
     size_t nRoutesA = parents.first->numRoutes();
     size_t nRoutesB = parents.second->numRoutes();
 
-    // Picking the start index of routes to replace of parent A. We like to
-    // replace routes with a large overlap of tasks, so we choose adjacent
-    // routes (they are sorted on polar angle by the local search/educate step).
```
N-Wouda, line 22: I hadn't looked at this method in a long long time, but now that I do: this is really dependent on a super non obvious detail of the current LS implementation. I don't think we need to change that in this PR, but can we keep the comment that explains we rely on that polar angle sort somehow?

pyvrp/cpp/crossover/crossover.h
```diff
@@ -19,26 +19,36 @@ void greedyRepair(std::vector<std::vector<int>> &routes,
...
+ * from Nature. Springer. 2010, pp. 536–545.
  */
 Individual selectiveRouteExchange(
     std::pair<Individual const *, Individual const *> const &parents,
     ProblemData const &data,
     PenaltyManager const &penaltyManager,
-    XorShift128 &rng);
+    size_t startA,
+    size_t startB,
+    size_t const nMovedRoutes);
```
N-Wouda, line 52: The `A`, `B` terminology follows from Nagata and Kobayashi, I think. I vaguely recall reading about it when looking at edge assembly. I get that we use that in the implementation to match the paper, but in the interface can we maybe use a tuple like with the parents? Something like
```cpp
std::pair<size_t, size_t> starts,
```
or similar. Then our use of first/second matches that of the parents, and we don't need to use A/B outside of the SREX implementation (which I don't think we do anywhere else - we call things first/second in multiple places, but not A/B).

For `nMovedRoutes`, maybe `numMovedRoutes`?

pyvrp/crossover/tests/test_selective_route_exchange.py
```diff
@@ -1,7 +1,11 @@
-from numpy.testing import assert_equal
+from numpy.testing import assert_equal, assert_raises
+from pytest import mark
 
 from pyvrp import Individual, PenaltyManager, XorShift128
 from pyvrp.crossover import selective_route_exchange as srex
+from pyvrp.crossover._selective_route_exchange import (
+    selective_route_exchange as _srex,
+)
```
N-Wouda, line 8: Maybe `cpp_srex` to clarify in code that this involves the native extension?

pyvrp/crossover/tests/test_selective_route_exchange.py
```diff
@@ -20,6 +24,31 @@ def test_same_parents_same_offspring():
     assert_equal(offspring, individual)
 
 
+# All tests below use the deterministic C++ implementation of SREX, which,
+# instead of a random number generator, takes as arguments the first and
+# second start indices and the number of routes to move.
```
N-Wouda, line 29: I think this will become superfluous if we rename `_srex`.

N-Wouda (commented):
This looks good to me. Only a few comments on the interface and naming, but nothing substantial!

N-Wouda (approved):
LGTM! :shipit:

Issue #170, opened 2023-03-04
Do we need probabilistic intensification?


N-Wouda:
For CVRP, our default is already to always intensify (see `configs/cvrp.toml`). For VRPTW, we intensify 15% of the time, or when we find a new best solution (see `configs/vrptw.toml`). How much worse is it to always intensify for VRPTW as well, and remove the parameter altogether?

### Benchmarks

On VRPTW and current main, since CVRP already does 100% and I do not propose we change that. Current main is eacde31981dfe354274bfd2c973ccbd80e63f631. One hour, average over ten seeds.

| category     | Never | Current (15%) | Always |
|-----|-------|---------------|----------------|
| C1  | **413476.4** | 413508.7 | 413505.4 |
| C2  | 161968.6 | **161899.1** | **161899.1** |
| R1  | 473631.6 | 473244.7 | **473185.8** |
| R2  | **276960.7** | 277035.6 | 277338.1 |
| RC1 | 442861.8 | 442598.2 | **442229.6** |
| RC2 | **232205.0** | 232231.6 | 232349.2 |

_Never_ behaves as if route operators do not exist at all, so there is also no intensification for new best solutions.

<details>

Job IDs:
- Never: 3082452
- Current: 3071365
- Always: 3082871

Never intensify:
```
      inst                              
obj   C1    413476.4  413317.0  413618.9
      C2    161968.6  161940.7  162009.7
      R1    473631.6  473161.8  474195.7
      R2    276960.7  276735.0  277202.2
      RC1   442861.8  442449.7  443491.5
      RC2   232205.0  232028.2  232422.5
iters C1    155327.1  141096.9  161388.9
      C2    186431.3  164638.5  196707.7
      R1     95254.8   90775.0   99113.1
      R2     80365.8   71534.8   85950.6
      RC1   114015.4  105466.9  122114.0
      RC2    90840.8   81770.4   97357.8
```

Current (15%):
```
                mean       min       max
      inst                              
obj   C1    413508.7  413262.9  413690.7
      C2    161899.1  161878.3  161919.7
      R1    473244.7  472820.1  474150.6
      R2    277035.6  276801.2  277295.6
      RC1   442598.2  442179.3  443196.9
      RC2   232231.6  231965.8  232415.0
iters C1    125830.8   98592.9  139190.9
      C2    147227.0  111460.5  167825.8
      R1     70815.6   53593.1   87463.3
      R2     60154.5   45681.5   66854.4
[...]

PR #176, opened 2023-03-06
Allow empty search space in local search

N-Wouda:
Follows somewhat from #170. I don't think we should disallow an empty search space, either due to no operators or an empty granular neighbourhood.


N-Wouda:
This needs work on allowing route/node operators to be passed to the LS's constructor, as well as no neighbourhood.

wouterkool:
Maybe I missed it, but I do not fully see why we want this? I think 9/10 cases having no operators is a mistake and just doing nothing is confusing for the users. If you don't want `search` to do anything, why not just skip the call from Python? This is much more explicit. Otherwise, should we at least consider emitting a warning (which is ugly but may be helpful)?

PR #177, opened 2023-03-06
Remove PenaltyManager state from LS operators

N-Wouda:
This PR is part of #157 and removes the `PenaltyManager` reference from LS operators.


N-Wouda:
@leonlan @wouterkool can you have a look? I debated removing the data piece of the state as well, but in the end I figured it wasn't a problem and having it in the operators simplifies the interface a bit. 

Removing `PenaltyManager` seems more pressing. It's now passed in by the local search, as a step up to removing it there as well in favour of a small penalties object.

leonlan (approved):
LGTM.

wouterkool (approved):
Nice! One thing to consider in the future is how we make sure we invalidate the SWAP* cache if the penalties change. I think now, in theory, we could change the penalties but have the old SWAP* cache (but this doesn't happen as we always load the individual again after changing penalties).

wouterkool:
@N-Wouda can we merge this?

Issue #181, opened 2023-03-09
Various HGS improvements by other EURO-NeurIPS teams


wouterkool:
While OptiML was the clear winner for the static VRPTW, several other teams made improvements to the static HGS solver for the [EURO Meets NeurIPS 2022](https://euro-neurips-vrp-2022.challenges.ortec.com/) competition. We could see if we could benefit from them (or invite them to contribute to this repo?):

Kleopatra (https://github.com/tumBAIS/euro-meets-neurips-2022 not yet available):
* Persist unchanged routes (`LocalSearch::lastModified`) across local searches as SREX does keep some routes unchanged, checking these again after crossover is useless
* Don't shuffle correlated neighbours as they introduce bad arcs with limited runtimes (I think we do the same in keeping the order fixed, but they are ordered by index rather than proximity, while this team suggests ordering them by proximity yields faster improvement)?
* Reduce iterations until restart because less shuffling causes less randomness in the search

Team SB (https://github.com/yd-kwon/NeurIPS-2022-VRPTW-Competition):
* Generalize binary to k-ary tournament to focus more on higher fitness individuals (this gave most performance improvement)
* Always remove individuals after number of offspring to encourage diversity (small but significant improvement)
* Restart mechanism uses warm start from previous population (rather than completely random)

MTGBWS (https://github.com/steveharenberg/sas-euro-neurips-vrp-2022):
* Corrected problematic bugs, integer overflow, division by zero, looping order-sensitive over unordered set (I assume these have all been fixed here as well?)
* Rewrite recursive calls that were not inlined (5-10% improvement)
* Swap data structures, remove o(n) search by o(1) lookup (not sure what is meant exactly, maybe searching for the first empty route in local search?)
* Matrix in 2D array implementation with contiguous memory -> 10% speedup

N-Wouda:
I do not know about the functional enhancements, but re. fixing existing implementation:

>Corrected problematic bugs, integer overflow, division by zero, looping order-sensitive over unordered set (I assume these have all been fixed here as well?)

I'm pretty sure we have fixed most of these along the way as well. To really make sure we need a bit more detail, but Valgrind is happy (other than the case in #113). We use an unordered set only in SREX, and then just for membership tests - not for iteration. So I think we do not have these problems any more.

>Rewrite recursive calls that were not inlined (5-10% improvement)

I cannot quickly think of a place where we use recursion, other than in `TWS::merge` at compile-time. @wouterkool @leonlan do you know other cases?

[...]

PR #183, opened 2023-03-10
Generalise binary tournament

N-Wouda:
This PR follows from #181. It:

- [x] Generalises binary tournament to k-ary tournament.

<details>

**Notes**:

</details>


pyvrp/Population.py
```diff
@@ -156,10 +160,9 @@ def select():
...
-        item1 = select()
-        item2 = select()
+        if k <= 0:
+            raise ValueError(f"Expected k > 0; got k = {k}.")
 
-        if item1.fitness < item2.fitness:
-            return item1.individual
-
-        return item2.individual
+        items = [select() for _ in range(k)]
```
wouterkool, line 166: We could pass in a generator to min directly?

pyvrp/tests/test_Population.py
```diff
@@ -239,7 +239,8 @@ def test_elite_individuals_are_not_purged(nb_elite: int):
         assert_(id(elite_individual) in new_individuals)
 
 
-def test_binary_tournament_ranks_by_fitness():
+@mark.parametrize("k", [2, 3])
```
wouterkool, line 242: I think the test for `k=3` also succeeds if `k=2` would be used internally. Should we add another test that tests the actual frequency that individuals get selected rather than their rank? I think the expected frequency for individual $i$ with rank $r_i$ is something like $(1-\frac{r_i}{n_{pop}})^{k-1} * \frac{k \cdot n_{samples}}{n_{pop}}$.

wouterkool (approved):

pyvrp/Population.py
N-Wouda: Yes, but I'm reasonably sure that is slower. There also isn't a real memory concern here (because a reasonable value for $k$ is fairly small, say, <10).

pyvrp/tests/test_Population.py
N-Wouda: I've added an assert of this form, thanks!

Issue #100, opened 2023-02-20
Rethink PenaltyManager


N-Wouda:
Rather than having many classes depend on the `PenaltyManager` directly, we can ask the `PenaltyManager` for a value object that can provide the current load and time warp penalties.

wouterkool:
And let's implement the `PenaltyManager` in Python while we're at it? :)

wouterkool:
I have given this some thought. Using the penalties can be seen as using a proxy objective for the real objective (which has hard constraints), and this proxy objective changes during optimization as the penalties change. That makes the definition of `Individual.cost()` ambiguous and dependent on the state of the algorithm.

How about we make this explicit and define an `Objective` class (somewhat more general than Penalties) that takes an individual and returns the cost, and remove `Individual.cost` at all? The Objective class could also be used inside local search, providing the `twPenalty` and `loadPenalty` methods.

We can then implement the PenaltyManager class in Python and pass `PenaltyManager.get_objective()` (or perhaps `PenaltyManager.get_booster_objective()`) to `LocalSearch.run`.

The Objective class could also be used to define different objectives (e.g. taking into account waiting time) in the future, although this would require changes to the LocalSearch operators as well.

Note that the fitness computation in the `SubPopulation` depends on the cost, and it is rather dependent on subtle implementation details that the fitness is now recomputed correctly when the penalties change. I think we should somehow trigger the population explicitly when the Objective changes, or always pass the `Objective` as a parameter such that the implementation can detect itself when the objective changes.

wouterkool:
Regarding the `SubPopulation` that depends on the cost... we could also choose to let the 'managed penalties' only apply to the local search to encourage resulting in feasbile solutions, but keep a fixed penalty (i.e. a 'static' relaxed objective) when computing cost for an individual in the population. This makes things less coupled as population no longer depends on changing penalties, but it may affect the performance as well.

N-Wouda:
> How about we make this explicit and define an `Objective` class (somewhat more general than Penalties) that takes an individual and returns the cost, and remove `Individual.cost` at all? The Objective class could also be used inside local search, providing the `twPenalty` and `loadPenalty` methods.

[...]

PR #189, opened 2023-03-17
Remove penalty manager reference from individual

wouterkool:
This PR:

- [x] Removes the penalty manager reference from the individual. Related to #100 and #157.


wouterkool:
Some important comments:

The `OptimisationTarget` has been removed as the individual does not have a parameterless `.cost()` method. Instead, the stopping criterions now directly take the cost as argument. In any way, we should consider these implementation details of GeneticAlgorithm, so this is not our core API.

The statistics component now relies on the `PenaltyManager` being passed in when collecting statistics, so the objective changes (this is the current behaviour). We could also choose to let statistics have it's own constant `PenaltyManager` instance (later `CostEvaluator`), instead of reporting cost using changing definitions.

`Result.cost()` takes an optional `PenaltyManager` argument but has a default to support the parameterless `__str__` method.

`[Individual](std::ostream &operator<<` now prints the distance rather than cost, as the cost relies on the penalty manager. Not sure what we want here. Some options:
* Make an explicit `Individual::toString(PenaltyManager pm)`?
* Print default `cost` (distance) only if individual is feasible
* Remove printing the cost at all

pyvrp/cpp/Individual.cpp
```diff
@@ -118,15 +118,12 @@ bool Individual::operator==(Individual const &other) const
 {
     // First compare costs, since that's a quick and cheap check. Only when
     // the costs are the same do we test if the neighbours are all equal.
-    return cost() == other.cost() && neighbours == other.neighbours;
+    return distance == other.distance && capacityExcess == other.capacityExcess
+           && timeWarp == other.timeWarp && neighbours == other.neighbours;
```
N-Wouda, line 122: Should this also test the `nbRoutes` item? In terms of how, something like this requires less scanning, I think:
```cpp
return distance == ...
    && capacityExces = ...
    && timeWarp == ...
    && neighbours = ...;
```
This is something I explicitly disable `clang-format` for (with `// clang-format off` and `on`) since the default doesn't really make a lot of sense to me.

pyvrp/cpp/Individual.cpp
```diff
@@ -186,6 +179,6 @@ std::ostream &operator<<(std::ostream &out, Individual const &indiv)
         out << '\n';
     }
 
-    out << "Cost: " << indiv.cost() << '\n';
+    out << "Distance: " << indiv.distance << '\n';
```
N-Wouda, line 182: @leonlan do you know if this is allowed? (I'm not sure if VRPLIB requires a particular word / value here, or if we can do whatever we want)

pyvrp/cpp/Individual.h
```diff
@@ -85,28 +83,22 @@ class Individual
...
-    Individual(ProblemData const &data,
-               PenaltyManager const &penaltyManager,
-               Routes routes);
-};
+    Individual(ProblemData const &data, Routes routes);
 
-// Outputs an individual into a given ostream in VRPLIB format
-std::ostream &operator<<(std::ostream &out, Individual const &indiv);
+    // Outputs an individual into a given ostream in VRPLIB format
+    friend std::ostream &operator<<(std::ostream &out, Individual const &indiv);
```
N-Wouda, line 100: Suggestion: can we move this friend declaration up to the `private` section at the start of the class declaration? I think that works, and then the friendness is just an implementation detail again.

pyvrp/cpp/Individual_bindings.cpp
```diff
@@ -11,29 +11,21 @@ namespace py = pybind11;
...
+                        return Individual(data, rng);
                     },
                     py::arg("data"),
-                    py::arg("penalty_manager"),
-                    py::arg("routes"),
-                    py::keep_alive<0, 2>());  // keep penalty_manager alive
+                    py::arg("routes"));
             })
-        .def("cost", &Individual::cost)
+        .def("cost", &Individual::cost, py::arg("penalty_manager"))
```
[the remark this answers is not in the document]
N-Wouda: Or call it `pm` here as well, but I like to be a bit more descriptive in the interface than in the implementation.

pyvrp/cpp/SubPopulation_bindings.cpp
```diff
@@ -36,7 +36,10 @@ PYBIND11_MODULE(_SubPopulation, m)
              py::arg("diversity_op"),
              py::arg("params"),
              py::keep_alive<1, 3>())  // keep params alive
-        .def("add", &SubPopulation::add, py::arg("individual"))
+        .def("add",
+             &SubPopulation::add,
+             py::arg("individual"),
+             py::arg("penalty_manager"))
```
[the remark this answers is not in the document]
N-Wouda: Issue: the `purge()` methode below now also takes an argument that's not yet named.

pyvrp/educate/LocalSearch.py
```diff
@@ -124,7 +124,9 @@ def run(
 
             new_individual = self.intensify(individual, penalty_manager)
 
-            if new_individual.cost() < individual.cost():
+            if new_individual.cost(penalty_manager) < individual.cost(
+                penalty_manager
+            ):
```
N-Wouda, line 129: To avoid these awkward `black`-related expansions I typically assign to a variable first:
```python
new_individual = ...
new_cost = new_individual.cost(penalty_manager)

if new_cost < individual.cost(penalty_manager):
   ...
```

pyvrp/stop/MaxIterations.py
```diff
@@ -1,4 +1,4 @@
-from pyvrp.OptimisationTarget import OptimisationTarget
+from typing import Union
```
N-Wouda, line 1: Here, and in the other cases: just use `float`. MyPy treats `int` as a special case of `float` for type checking, so there's no real need for this amount of formality.

pyvrp/tests/test_Individual.py
```diff
@@ -269,16 +261,17 @@ def test_str_contains_essential_information():
                 assert_(str(client) in str_route)
 
         # Last line should contain the cost
-        assert_(str(individual.cost()) in str_representation[-1])
+        # TODO what do we want with cost in str representation?
+        # assert_(str(individual.cost(pm)) in str_representation[-1])
```
N-Wouda, line 265: Presumably keep it in there as-is, so leave this test intact. If desired we can open an issue to discuss it, but I think VRPLIB requires the cost to be in the (VRPLIB) format that we currently print the Individual as. But @leonlan knows that better than I do.

pyvrp/GeneticAlgorithm.py
```diff
@@ -117,43 +117,45 @@ def run(self, stop: StoppingCriterion):
...
 
         end = time.perf_counter() - start
         return Result(self._best, stats, iters, end)
 
     def _educate(self, individual: Individual):
         def is_new_best(indiv):
-            return indiv.is_feasible() and indiv.cost() < self._best.cost()
+            return indiv.is_feasible() and indiv.cost(
+                self._pm
+            ) < self._best.cost(self._pm)
```
N-Wouda, line 155: Can we assign the indiv's cost to a variable before doing this check? I think that's much easier to read over two lines than what we have currently.

Something like:
```python
cost = indiv.cost(self._pm)
return indiv.is_feasible() and cost < self._best.cost(self._pm)
```

pyvrp/Population.py
```diff
@@ -88,11 +89,17 @@ def add(self, individual: Individual):
         ----------
         individual
             Individual to add to the population.
+        penalty_manager
+            PenaltyManager to use to compute the cost. Required here
+            since adding an individual may trigger a purge which needs to
+            compute the biased fitness which requires computing the cost.
```
N-Wouda, line 95: Part of this is an implementation detail that does not have to be explained here, I think. I would move the part about "Required here because ..." into a comment.

pyvrp/Result.py
```diff
@@ -40,16 +42,23 @@ def __post_init__(self):
         if self.runtime < 0:
             raise ValueError("Negative runtime not understood.")
 
-    def cost(self) -> float:
+    def cost(self, penalty_manager: Optional[PenaltyManager] = None) -> float:
```
N-Wouda, line 45: @leonlan @wouterkool: does it make sense to have a default here? I'm not sure right now. I think I'd rather have that we force a user to be explicit about this, since I cannot really answer clearly what the default here should be.

A middle ground that make sense to me would be the following:
- If the individual is feasible, return the cost with a default PenaltyManager. The penalty terms are zero, so this is just the distance.
- If the individual is infeasible, refuse the temptation to guess what the cost should be. Just raise a ValueError and explicitly force a user to provide a PenaltyManager instance.

pyvrp/_SubPopulation.pyi
```diff
@@ -49,6 +52,10 @@ class SubPopulation:
         ----------
         individual
             Individual to add to the subpopulation.
+        penalty_manager
+            PenaltyManager to use to compute the cost. Required here
+            since adding an individual may trigger a purge which needs to
+            compute the biased fitness which requires computing the cost.
```
N-Wouda, line 58: Same as in the implementation. This leaks a bit of an implementation detail on _why_ we require it. A user shouldn't need to know this, or care. I think a comment in the implementation is sufficient.

N-Wouda (commented):
LGTM, a few comments. One think I noticed is that we currently use `pm` and `penaltyManager / penalty_manager` in multiple places. I'd prefer `pm` in the implementation of methods (where we can be brief - the implementation is ours), but the long form in the interface declarations (where we need to make sure the user understands what's going on was well). Several comments relate to this.

N-Wouda:
This merges into #187, but I'd rather merge that PR first and then merge this PR into main directly. They're fairly decoupled and that way we do not stack things.

pyvrp/cpp/Individual.cpp
N-Wouda: I don't mean the package, I mean the format specification, if any exists

N-Wouda: Ah OK, then we can do whatever. Thanks!

pyvrp/cpp/Individual.cpp
wouterkool: Yes! It's a bit weird as `numRoutes_` has a `_` whereas the other properties don't but I think that is a separate issue.

N-Wouda: That's because it's data but also available as a method, and they cannot be named the same. I suspect we'll want to ahve the same for e.g. distance at some point.

N-Wouda: The comment above this is out of date now too. Can we update that to something along the lines of "first check cheap attributes, and if those all agree we check the expensive thing to really be sure"?

pyvrp/cpp/Individual_bindings.cpp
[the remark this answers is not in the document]
wouterkool: I agree, but it will be replaced by CostEvaluator anyway so I'm not going to do this now ;)

N-Wouda:
Can this PR (and #187) be merged into main directly?

pyvrp/cpp/SubPopulation_bindings.cpp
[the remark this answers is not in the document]
wouterkool: Added parameter to the `purge()` method, other one will be fixed when we have `CostEvaluator`

pyvrp/Population.py
wouterkool: Done

pyvrp/GeneticAlgorithm.py
wouterkool: I agree it's not pretty (did not check the auto formatting). However it seems arbitrary to me to assign `cost` to a variable but have `best_cost` computed inline, so I propose either:
```
cost = indiv.cost(self._pm)
best_cost = self._best.cost(self._pm)
return indiv.is_feasible() and cost < best_cost
```
or
```
return (
    indiv.is_feasible()
    and indiv.cost(self._pm) < self._best.cost(self._pm)
)
```
I prefer declarative over imperative so I choose the second option. Let me know if you have strong objections.

wouterkool: Nevermind, apparently black doesn't allow me the second option so I implemented the first.

pyvrp/Result.py
wouterkool: I agree with this middle ground, although I want to note that it means that the Result uses implicit knowledge on how the cost is computed for feasible individuals.

If we are ok with this in general, I propose to also have a default `.cost()` method in individual, perhaps by separating `.cost()` and `.penalizedCost(PenaltyManager pm)` where the first returns an exception for `infeasible` individuals.

We plan to remove `cost` completely (to `CostEvaluator`) (see #191), but in this case we could also keep the default `cost` and only move the `penalizedCost`. This would also allow us to print the cost in the representation for feasible individuals.

wouterkool: Added comment to #100 to discuss this.

pyvrp/tests/test_Individual.py
wouterkool: I agree we should ideally have it in here but that is not simply possible as the `__str__` function cannot rely on the penalty manager, so I'll leave this for now. We can discuss it further in #100.

pyvrp/educate/LocalSearch.py
wouterkool: This will be fixed with #191 anyway so I'm not changing it now.

pyvrp/stop/MaxIterations.py
wouterkool: The `OptimisationTarget` was using the `Union` which is why I kept it like this but I'll change it.

pyvrp/cpp/Individual.cpp
wouterkool: > That's because it's data but also available as a method, and they cannot be named the same. I suspect we'll want to ahve the same for e.g. distance at some point.

I will fix this in #191 

> The comment above this is out of date now too. Can we update that to something along the lines of "first check cheap attributes, and if those all agree we check the expensive thing to really be sure"?

Yes, thanks!

PR #191, opened 2023-03-17
Move Individual::cost to CostEvaluator::operator()

wouterkool:
This PR:

- [x] Moves Individual::cost to CostEvaluator::operator(), see #100.


pyvrp/tests/test_Individual.py
```diff
@@ -262,8 +262,7 @@ def test_str_contains_essential_information():
 
         # Last line should contain the cost
         # TODO what do we want with cost in str representation?
-        # assert_(
-        #     str(individual.cost(cost_evaluator)) in str_representation[-1])
+        # assert_(str(cost_evaluator(individual)) in str_representation[-1])
```
N-Wouda, line 265: Maybe we can check that the distance is in the representation? Then we can add other attributes later, as wanted/needed.

examples/cvrp.ipynb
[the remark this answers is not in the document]
wouterkool: Yes, my bad, the code was correct but the output was from an incorrect version without scaling (I override the output to not change the figures but this one slipped through).

pyvrp/tests/test_Individual.py
wouterkool: Done

PR #192, opened 2023-03-21
Pass CostEvaluator to Population.select() and .get_binary_tournament()

### LINKED ISSUE #100: Rethink PenaltyManager

wouterkool:
This PR:
* [Pass CostEvaluator to select() and get_binary_tournament()](https://github.com/N-Wouda/PyVRP/commit/858fdf4c4ae382f3fb0140f7b2e20bec4d02ecab)
* [Cache CostEvaluator in PenaltyManager](https://github.com/N-Wouda/PyVRP/commit/bec9cd2f34aafd3b389a561bca80257cd63c400a)

- [x] Closes #100.


pyvrp/Population.py
```diff
@@ -139,21 +150,40 @@ def clear(self):
...
         -------
         Individual
             The selected individual.
         """
 
+        if update_fitness:
+            # Note: even though the cost does not change, for the feasible
+            # subpopulation, we must always call compute_fitness as it is not
+            # updated when adding individuals
+            self._feas.update_fitness(cost_evaluator)
```
leonlan, line 184: Question: Is it needed to update the fitness of the feasible population? The costs (and thus fitness) don't change because feasible individuals don't have penalty.

wouterkool: Yes, since with this change we move to lazy evaluation (compute the fitness when you need it) so the fitness will be outdated after a new individual has been hadded (even though the cost function did not change). See also the comment above the line.

leonlan: Oh I completely missed the comment 😅. Thanks, makes sense now.

pyvrp/PenaltyManager.py
```diff
@@ -124,6 +124,23 @@ def __init__(
...
+
+    def _update_cost_evaluators(self):
+        # Updates the cost evaluators given new penalty values
+        self._cost_evaluator = CostEvaluator(
+            self._capacity_penalty, self._tw_penalty
+        )
+        self._booster_cost_evaluator = CostEvaluator(
+            self._capacity_penalty * self._params.repair_booster,
+            self._tw_penalty * self._params.repair_booster,
+        )
```
N-Wouda, line 143: Do these need to be cached? They're simple objects to create and destroy, is that really costing us?

pyvrp/_SubPopulation.pyi
```diff
@@ -67,11 +67,14 @@ class SubPopulation:
...
         """
-    def update_fitness(self) -> None:
+    def update_fitness(self, cost_evaluator: CostEvaluator) -> None:
         """
         Updates the biased fitness scores of individuals in the subpopulation.
         This fitness depends on the quality of the solution (based on its cost)
         and the diversity w.r.t. to other individuals in the subpopulation.
+
+        Important: this function must be called before accessing the
+        SubPopulationItem.fitness attribute.
```
N-Wouda, line 77: This should be in a sphinx warning environment, something like

```
.. warning::

   This method must be called [..]
```

pyvrp/Population.py
```diff
@@ -140,21 +151,40 @@ def clear(self):
...
+            self._infeas.update_fitness(cost_evaluator)
```
N-Wouda, line 186: How much does it cost us if we do this every time, instead of having a flag that determines whether to update yes/no?

pyvrp/Population.py
```diff
@@ -111,14 +116,20 @@ def select(self, rng: XorShift128) -> Tuple[Individual, Individual]:
         ----------
         rng
             Random number generator.
+        cost_evaluator
+            Cost evaluator to use when computing the fitness.
+        update_fitness
+            Boolean whether to update the fitness values before selecting the
+            parents. Can be set to False for efficiency if it is known that the
+            fitness has not changed, e.g. when calling select repeatedly.
```
N-Wouda, line 124: I'd add a kwargs here and say something like "any additional keyword arguments are passed to get_binary_tournament". That'd then also work well with #183 for `k`.

pyvrp/Population.py
```diff
@@ -140,21 +151,40 @@ def clear(self):
         self._feas = SubPopulation(self._op, self._params)
         self._infeas = SubPopulation(self._op, self._params)
 
-    def get_binary_tournament(self, rng: XorShift128) -> Individual:
+    def get_binary_tournament(
+        self,
+        rng: XorShift128,
+        cost_evaluator: CostEvaluator,
+        update_fitness: bool = True,
+    ) -> Individual:
```
N-Wouda, line 159: For `update_fitness` a keyword-only argument makes sense to me. Just a True/False when calling this does not say much.

pyvrp/tests/test_Population.py
```diff
@@ -166,8 +166,8 @@ def test_select_returns_same_parents_if_no_other_option():
     # small, but not zero. So let's do an experiment where we do 1000 selects,
     # and collect the number of times the parents are different.
     different_parents = 0
-    for _ in range(1_000):
-        parents = pop.select(rng)
+    for i in range(1_000):
+        parents = pop.select(rng, cost_evaluator, i == 0)
```
N-Wouda, line 170: Do we need the really clever `i == 0` stuff? Just updating every iteration should also be OK for tests right? If we somehow do need it, can we be more explicit with an if/else?

pyvrp/tests/test_SubPopulation.py
```diff
@@ -126,6 +128,8 @@ def test_fitness_is_average_of_cost_and_diversity_when_no_elites():
     ranks[cost_rank[div_rank], 1] = np.arange(len(subpop))
 
     expected_fitness = ranks.sum(axis=1) / (2 * len(subpop))
+    # We need to call update_fitness before accessing the fitness
+    subpop.update_fitness(cost_evaluator)
```
N-Wouda, line 132: Same here

N-Wouda (changes requested):

N-Wouda:
I'm not sure about the "maybe update" flag, and would like not to have it if it doesn't cost us too much. In general, a lot of the caching feels a bit premature. Do we really need that already?

wouterkool:
I think removing the flag in favour of a public and private method is a good suggestion! I was thinking to support if a user wants to do multiple selects without a change in population and requires efficiency but in that case we should probably implement this as a function of population so the caching would remain internal (or consider alternatives at that point).

Regarding the caching, I don't know if it's a bottleneck but this way I know for sure it won't be, and I think the added complexity is very small and contained in the class. Repeatedly instantiating the same object for single use (e.g. in `GeneticAlgorithm` where it's accessible as a property) doesn't feel as the right way to me (but I may be wrong). Do you really think not caching is better? (p.s. I looked into a general method for defining a cached property with dependencies but this is too complex).

pyvrp/Population.py
wouterkool: Good idea, done

pyvrp/Population.py
wouterkool: Yes, although I followed your suggestion to remove it completely.

pyvrp/tests/test_Population.py
wouterkool: Didn't do this since we removed the flag completely. For tests it is ok indeed.

pyvrp/tests/test_SubPopulation.py
```diff
@@ -95,6 +95,8 @@ def test_fitness_is_purely_based_on_cost_when_only_elites():
     rank[by_cost] = np.arange(len(subpop))
 
     expected_fitness = rank / (2 * len(subpop))
+    # We need to call update_fitness before accessing the fitness
+    subpop.update_fitness(cost_evaluator)
```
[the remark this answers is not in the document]
wouterkool: I think it's more explicit to call `update_fitness` directly before accessing the fitness and more robust (w.r.t. code changes) as the code implicitly relies on the population not changing after the `update_fitness` which is guaranteed if nothing happens in between.

However, I agree it's less pretty and it's not that important so I've changed it following your suggestion.

pyvrp/Population.py
wouterkool: Good idea but I've removed it completely like you suggested.

pyvrp/PenaltyManager.py
wouterkool: See my other comment, are you ok with keeping it or do you really think it should be changed?

BTW independently of the question whether it's needed I agree this code is not very pretty as it duplicates the initialization code. However I could not see (in general) the best solution? Do you have a suggestion?
* Not defining the properties in the constructor is bad practice.
* Initializing to `None` gets `mypy` complaining. Typing as optional seems overkill but maybe disable the check?
* We do not (yet) have a default constructor for `CostEvaluator` and using this would also not be pretty.

N-Wouda: I don't see a clean way of avoiding this, other than calling `self._update_cost_evaluators()` from the initialiser. Let's leave it as-is.

I ran some experiments with instantiating `CostEvaluator` a million times or getting it from a variable. The variable is some 6x faster. I doubt it matters a lot, but we can keep it in.

N-Wouda (approved):

pyvrp/PenaltyManager.py
wouterkool: Thanks for the experiments! I think its good that we have the discussion and aim to avoid pre-mature optimization in general but here it feels appropriate, now we won't have to worry it could become a bottleneck either. :)

PR #194, opened 2023-03-22
Default cost evaluator

wouterkool:
This PR:
- Separate `CostEvaluator.cost()` and `CostEvaluator.penalizedCost()` (currently using `penalizedCost()` basically everywhere).
- For tests, explicitly instantiate `CostEvaluator` with penalty values rather than getting them via PenaltyManager
- Renames `PenaltyParams.init_tw_penalty` to `PenaltyParams.init_time_warp_penalty` since this is the definition used in the configuration files.

- [x] Partially implements #100.


pyvrp/cpp/CostEvaluator_bindings.cpp
```diff
@@ -10,10 +10,20 @@ PYBIND11_MODULE(_CostEvaluator, m)
         .def(py::init<unsigned int, unsigned int>(),
              py::arg("capacity_penalty"),
              py::arg("tw_penalty"))
+        .def_property_readonly_static(
+            "get_default",                // this is a bit of a workaround for
+            [](py::object)                // classmethods, because pybind does
+            {                             // not yet support those natively.
+                return py::cpp_function(  // See issue 1693 in the pybind repo.
+                    []() { return CostEvaluator(0, 0); });
+            })
```
N-Wouda, line 19: What's the intended use case for this method?

pyvrp/GeneticAlgorithm.py
leonlan, line 1: I think `penalized_cost` should be `cost` in GA, since we use it to evaluate the best individual. 

(Related: https://github.com/N-Wouda/PyVRP/pull/194/files#r1145855589 for Result.)

pyvrp/cpp/CostEvaluator_bindings.cpp
wouterkool: We agreed on 0 as defaults, but I thought we could not have multiple constructor overloads or default values using the bindings so I implemented this as a helper function. We could also make it a singleton instance actually (but not sure if we can implement this in the bindings... suggestions welcome!

pyvrp/cpp/CostEvaluator.h
```diff
@@ -35,7 +35,14 @@ class CostEvaluator
     /**
      * Computes the objective (penalized cost) for a given individual.
      */
-    [[nodiscard]] unsigned int operator()(Individual const &individual) const;
+    [[nodiscard]] unsigned int
+    penalizedCost(Individual const &individual) const;
+
+    /**
+     * Computes the objective for a given individual. Returns infinity (or some
+     * large value if the individual is infeasible.
```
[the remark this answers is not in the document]
wouterkool: Done

pyvrp/GeneticAlgorithm.py
wouterkool: #198

pyvrp/cpp/CostEvaluator.cpp
```diff
@@ -6,10 +6,15 @@ CostEvaluator::CostEvaluator(unsigned int capacityPenalty,
...
 {
     auto const loadPen = loadPenaltyExcess(individual.excessLoad());
     auto const twPen = twPenalty(individual.timeWarp());
 
     return individual.distance() + loadPen + twPen;
 }
+
+unsigned int CostEvaluator::cost(Individual const &individual) const
+{
+    return individual.isFeasible() ? individual.distance() : UINT32_MAX;
```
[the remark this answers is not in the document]
wouterkool: Done. This should then also be done for XorShift128? See #197.

pyvrp/Result.py
```diff
@@ -67,10 +66,10 @@ def cost(self, cost_evaluator: Optional[CostEvaluator] = None) -> float:
             if not self.best.is_feasible():
                 raise ValueError(
                     "Best found solution is infeasible! Provide the "
-                    "penalty_manager argument to compute the cost."
+                    "cost_evaluator argument to compute the cost."
                 )
-            cost_evaluator = PenaltyManager().get_cost_evaluator()
-        return cost_evaluator(self.best)
+            cost_evaluator = CostEvaluator.get_default()
+        return cost_evaluator.penalized_cost(self.best)
```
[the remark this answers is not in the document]
wouterkool: Since the return type is float and Python has actual `math.inf` I implemented the special case for infeasible.

pyvrp/tests/test_Result.py
```diff
@@ -93,7 +95,7 @@ def test_str_contains_essential_information():
 
         # Test that feasibility status and solution cost are presented.
         if individual.is_feasible():
-            assert_(str(cost_evaluator(individual)) in str_representation)
+            assert_(str(cost_evaluator.cost(individual)) in str_representation)
         else:
             assert_("INFEASIBLE" in str_representation)
```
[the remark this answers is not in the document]
wouterkool: Done

pyvrp/cpp/CostEvaluator.h
```diff
@@ -35,7 +35,14 @@ class CostEvaluator
     /**
      * Computes the objective (penalized cost) for a given individual.
      */
-    [[nodiscard]] unsigned int operator()(Individual const &individual) const;
+    [[nodiscard]] unsigned int
+    penalizedCost(Individual const &individual) const;
```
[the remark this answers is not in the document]
wouterkool: I like the Z :( but ok.

pyvrp/cpp/CostEvaluator_bindings.cpp
N-Wouda: I think we can have one constructor that has `0` as default. In the constructor bindings the arguments are assignable:
```
py::arg("val") = 0
```
That should work!

pyvrp/GeneticAlgorithm.py
N-Wouda: I don't really understand how this is a change in functionality. The costs are logged in `Statistics`, not `GeneticAlgorithm`. The `GeneticAlgorithm` can directly use `cost()` to determine the new best solution without having to check feasibility - if the solution's infeasible, it has cost +inf, and never becomes the new incumbent. I think that's pretty neat.

pyvrp/GeneticAlgorithm.py
```diff
@@ -95,7 +95,7 @@ def __init__(
         # infeasible solution as the initial best.
         feas = [indiv for indiv in initial_solutions if indiv.is_feasible()]
         sols = feas if feas else initial_solutions
-        self._best = min(sols, key=self._cost_evaluator)
+        self._best = min(sols, key=self._cost_evaluator.penalized_cost)
```
N-Wouda: If we use `cost()` this whole filtering is not needed: just find the minimum in `initial_solutions`.

pyvrp/Result.py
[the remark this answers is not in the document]
N-Wouda: OK, but then why can this not return +inf when infeasible and actual `cost()` when feasible?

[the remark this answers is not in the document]
wouterkool: Yes that's what I did, this code is outdated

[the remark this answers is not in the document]
N-Wouda: Ah, yes, sorry.

pyvrp/GeneticAlgorithm.py
wouterkool: Sorry, you're right, I missed that genetic only uses it on feasible solutions. I will change it.

For `Statistics`, I think the discussion in #198 remains relevant.

pyvrp/cpp/CostEvaluator_bindings.cpp
wouterkool: I miss the 'mind-is-blown' emoticon to reply. But thanks, that's much better :)

N-Wouda (approved):
LGTM! :shipit:

PR #199, opened 2023-03-24
Improve CostEvaluator tests

wouterkool:
* Remove CostEvaluator from Individual tests
* Add CostEvaluator tests for cost and penalized_cost

This PR:

- [x] Implements CostEvaluator tests for #100.

### DIFF
--- a/pyvrp/tests/test_CostEvaluator.py
+++ b/pyvrp/tests/test_CostEvaluator.py
@@ -62,36 +62,55 @@ def test_tw_penalty():
     assert_equal(cost_evaluator.tw_penalty(2), 8)
 
 
-@mark.parametrize(
-    "routes",
-    [[[1, 2], [3], [4]], [[1, 2, 3, 4]]],
-)
-def test_cost_and_penalized_cost(routes):
+def test_cost():
+    data = read("data/OkSmall.txt")
+    default_cost_evaluator = CostEvaluator()
+    cost_evaluator = CostEvaluator(20, 6)
+
+    # Feasible individual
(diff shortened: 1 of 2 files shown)


pyvrp/tests/test_CostEvaluator.py
```diff
@@ -58,3 +60,38 @@ def test_tw_penalty():
...
+
+    # Test normal cost, should be distance or infinite if infeasible
+    if individual.is_feasible():
+        assert_equal(cost_evaluator.cost(individual), distance)
+        assert_equal(default_cost_evaluator.cost(individual), distance)
+    else:
+        # C++ code represents infinite as max value for unsigned integer
+        INFEAS_COST = np.iinfo(np.uint32).max
+        assert_equal(cost_evaluator.cost(individual), INFEAS_COST)
+        assert_equal(default_cost_evaluator.cost(individual), INFEAS_COST)
```
leonlan, line 97: Can we separate this test the feasible and infeasible individual? It's not clear to me which of the individuals are feasible (and whether we hit all the assertions in the conditionals).

leonlan: I think it'd also be easier to read if the test is divided into two sections, where one section uses the default cost evaluator, and one uses the non-default one.

wouterkool: I separated into two tests (for `cost` and `penalised_cost`) with both individuals instantiated inside the test.

N-Wouda (approved):

PR #205, opened 2023-04-05
Import DoubleMatrix and IntMatrix

### LINKED ISSUE #204: ProblemData.distance_matrix requires importing IntMatrix / test_read.py fails
#### Describe the bug

`test_read.py` fails when running independently (it succeeds when running all tests), since it uses `ProblemData.distance_matrix` which requires `IntMatrix` to be imported.

wouterkool:
This PR adds importing `DoubleMatrix` and `IntMatrix` in `pyvrp.__init__`. This ensures that these classes are known to the interpreter so that `ProblemData.distance_matrix` can be accessed without raising a type error.

- [x] Closes #204.

### DIFF
--- a/pyvrp/_ProblemData.pyi
+++ b/pyvrp/_ProblemData.pyi
@@ -116,6 +116,16 @@ class ProblemData:
         int
             Travel duration between the given clients.
         """
+    def distance_matrix(self) -> Any:
+        """
+        Returns the travel duration matrix used for distance/duration
+        computations.
+
+        Returns
+        -------
+        Any
+            Travel duration matrix.
+        """
     @property
     def num_clients(self) -> int:
         """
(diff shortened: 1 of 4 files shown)


leonlan:
Removing the binding "fixes" the problem, but we probably want to keep duration_matrix exposed so that users can access it. Since we define our own Matrix types, this means that `IntMatrix` or `DoubleMatrix` must always be imported before accessing this attribute to avoid a `TypeError`.

I think that addressing #61 is a better way to address this issue. @N-Wouda what do you think?

wouterkool:
I don't see a strong need for exporting the `distance_matrix` as we already have the `dist` function? If we have it, it also creates problems with ownership if the ProblemData goes out of scope, or it should return a copy? In any case, we could bring it back as a numpy array with #61 but have this as fix for now? It's not used anywhere except in a test.

leonlan:
Running `test_read.py` individually also passes without errors now.

PR #206, opened 2023-04-05
Initialize ProblemData with list of Clients

wouterkool:
This PR:
* Implements the constructor of ProblemData to initialize with a list of Clients rather than arrays.


pyvrp/cpp/ProblemData.cpp
```diff
@@ -16,26 +16,14 @@ size_t ProblemData::numVehicles() const { return numVehicles_; }
...
                          size_t numVehicles,
                          size_t vehicleCap,
-                         std::vector<std::pair<int, int>> const &timeWindows,
-                         std::vector<int> const &servDurs,
                          std::vector<std::vector<int>> const &distMat)
     : dist_(distMat),
-      clients_(coords.size()),
-      numClients_(static_cast<int>(coords.size()) - 1),
+      clients_(clients),
+      numClients_(clients.size() - 1),
```
N-Wouda, line 25: I only notice this now that we're changing it, but this could potentially be a really large number if size == 0. Can we do something like
```cpp
numClients_(std::max(clients.size(), 1) - 1),
```
to ensure this works fine when creating an empty instance?

pyvrp/cpp/ProblemData_bindings.cpp
```diff
@@ -8,6 +8,13 @@ namespace py = pybind11;
 PYBIND11_MODULE(_ProblemData, m)
 {
     py::class_<ProblemData::Client>(m, "Client")
+        .def(py::init<int, int, int, int, int, int>(),
+             py::arg("x"),
+             py::arg("y"),
+             py::arg("demand") = 0,
+             py::arg("service_duration") = 0,
+             py::arg("tw_early") = 0,
+             py::arg("tw_late") = 0)
```
N-Wouda, line 17: Can we not default these? I don't know what appropriate defaults for e.g. the time windows should be, but [0, 0] does not seem to me to be 'it'.

@leonlan what do you think?

N-Wouda: I'm going to try `tw_late = 0` and an all-zero duration matrix in a minute. If that works without issues I think "everything zero" is a reasonable default.

N-Wouda: Seems to work well!

N-Wouda (approved):

pyvrp/_ProblemData.pyi
```diff
@@ -39,45 +54,25 @@ class ProblemData:
...
     Parameters
     ----------
-    coords
-        Array of (x, y) coordinates. The first coordinate at index 0 is assumed
-        to be the depot.
-    demands
-        Array of client demands. The demand at index 0 is assumed to be the
-        depot's demand, and should be zero.
+    clients
+        The first client (at index 0) is assumed to be the depot. The time
```
wouterkool, line 58: Start description with 'List of clients'?

pyvrp/_ProblemData.pyi
```diff
@@ -8,29 +8,44 @@ class Client:
...
         Horizontal coordinate of this client, that is, the 'x' part of the
         client's (x, y) location tuple.
     y
         Vertical coordinate of this client, that is, the 'y' part of the
         client's (x, y) location tuple.
+    demand
+        The amount this client's demanding. Default 0.
+    service_duration
+        This client's service duration, that is, the amount of time we need to
+        visit the client for. Default 0.
```
wouterkool, line 21: Should we mention somewhere servicing should start in the interval (but not end per se)?

pyvrp/read.py
```diff
@@ -83,41 +84,43 @@ def read(
...
+    depots: np.ndarray = instance.get("depot", np.array([0]))
+    num_vehicles: int = instance.get("vehicles", dimension - 1)
+    capacity: int = instance.get("capacity", _INT_MAX)
 
-    depots = instance.get("depot", np.array([0]))
-    num_vehicles = instance.get("vehicles", num_clients - 1)
-    capacity = instance.get("capacity", _INT_MAX)
-    edge_weight = round_func(instance["edge_weight"])
+    edge_weight: np.ndarray = round_func(instance["edge_weight"])
+    distances = edge_weight  # distance and duration are assumed to be the same
```
wouterkool, line 98: Technically this comment does not hold for CVRP (I don't mind)

wouterkool (commented):
GitHub doesn't let me approve it but I still do ;)

N-Wouda:
@wouterkool thanks, I'll update the phrasing a bit and merge this in.

pyvrp/read.py
N-Wouda: Good catch, I've updated the code a bit to make this more obvious, by moving the durations assignment down to where we check time windows.

PR #209, opened 2023-04-19
Separate distance and duration matrices

### LINKED ISSUE #60: Rename distance_matrix to (travel_)duration_matrix

### LINKED ISSUE #184: Can we directly pass distance/time matrix as input to the ProblemData?
I always wonder can we directly pass distance/time matrix as input to the ProblamData instead of current co-ordinates informations. I guess it will results in different set of routes. Any suggestions! Thank you!

N-Wouda:
Adds support for separate distance and duration matrices. This PR:
- [x] Closes #184, closes #60;

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/_ProblemData.pyi
+++ b/pyvrp/_ProblemData.pyi
@@ -122,7 +122,7 @@ class ProblemData:
         int
             Travel distance between the given clients.
         """
-    def dur(self, first: int, second: int) -> int:
+    def duration(self, first: int, second: int) -> int:
         """
         Returns the travel duration between the first and second argument,
         according to this instance's travel duration matrix.
--- a/pyvrp/_ProblemData.pyi
+++ b/pyvrp/_ProblemData.pyi
@@ -1,4 +1,4 @@
-from typing import Any, List, Tuple
+from typing import List, Tuple
 
 from pyvrp._Matrix import Matrix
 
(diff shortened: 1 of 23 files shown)


N-Wouda:
A bit of a work in progress.

N-Wouda:
OK, I think content-wise we're mostly there. I still need to write tests, but I'll do that tomorrow I think.

N-Wouda:
@wouterkool @leonlan I think this is ready for review. The number of tests is somewhat small, but they do cover the main points. Let me know what you think!

leonlan (approved):
LGTM!

wouterkool (commented):
This looks great to me!

Only comment I have regards naming: compared to `dist` vs `distance`, I think `dur` is less self-explanatory than `duration`, which was what we previously implemented and discussed about [here](https://github.com/PyVRP/PyVRP/pull/169/files#diff-31568c78ee510de7ac71543baaee754f3bc38bcd323727c4c10561f290ae0b47). Also, to avoid things like `legDuration = duration()` would it be an idea to refer to the duration matrix as `durationMatrix`/`durationMat`/`durMat` rather than `duration` (same for distance matrix?)?

N-Wouda:
> Only comment I have regards naming: compared to `dist` vs `distance`, I think `dur` is less self-explanatory than `duration`, which was what we previously implemented and discussed about [here](https://github.com/PyVRP/PyVRP/pull/169/files#diff-31568c78ee510de7ac71543baaee754f3bc38bcd323727c4c10561f290ae0b47). 

Done. Also resolved several cases where we were already using `time`, `dur` for this.

> Also, to avoid things like `legDuration = duration()` would it be an idea to refer to the duration matrix as `durationMatrix`/`durationMat`/`durMat` rather than `duration` (same for distance matrix?)? It's just a suggestion, I'm fine keeping it as is.

Can't really be helped; `duration` is a data member of `TimeWindowSegment`, so we need a different name regardless. I have renamed the `dur` argument to `durationMatrix`.

PR #211, opened 2023-04-26
BugFix: loadIndividual sets depot TWdata to zero

wouterkool:
`LocalSearch::loadIndividual` previously resetted `startDepot->twAfter` and `endDepot->twBefore`. The idea is that these would be overwritten anyway in `route->update()`. However it is not if the route was unchanged (i.e. if we load the same route as in the previous result).

This causes the time warp of the current solution to be evaluated as 0, which is (sometimes) incorrect. This causes some improving moves with non-zero time warp to be incorrectly evaluated as non-improving.

It is fixed by not setting these attributes in loadIndividual, this keeps them at their previous values which is in sync with the cache used by route->update(). This PR also includes a test that reproduces the bug in main.

This PR:

### DIFF
--- a/pyvrp/cpp/educate/LocalSearch.cpp
+++ b/pyvrp/cpp/educate/LocalSearch.cpp
@@ -212,10 +212,8 @@ void LocalSearch::loadIndividual(Individual const &individual)
 
         startDepot->tw = clients[0].tw;
         startDepot->twBefore = clients[0].tw;
-        startDepot->twAfter = clients[0].tw;
 
         endDepot->tw = clients[0].tw;
-        endDepot->twBefore = clients[0].tw;
         endDepot->twAfter = clients[0].tw;
 
         Route *route = &routes[r];
--- a/pyvrp/educate/tests/test_Exchange.py
+++ b/pyvrp/educate/tests/test_Exchange.py
@@ -2,7 +2,12 @@ from numpy.testing import assert_, assert_equal
 from pytest import mark
 
 from pyvrp import CostEvaluator, Individual, XorShift128
-from pyvrp.educate import LocalSearch, NeighbourhoodParams, compute_neighbours
+from pyvrp.educate import (
+    LocalSearch,
+    NeighbourhoodParams,
+    Neighbours,
+    compute_neighbours,
+)
 from pyvrp.educate._Exchange import (
     Exchange10,
     Exchange11,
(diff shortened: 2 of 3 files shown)


pyvrp/cpp/educate/LocalSearch.cpp
```diff
@@ -212,10 +212,8 @@ void LocalSearch::loadIndividual(Individual const &individual)
 
         startDepot->tw = clients[0].tw;
         startDepot->twBefore = clients[0].tw;
-        startDepot->twAfter = clients[0].tw;
 
         endDepot->tw = clients[0].tw;
-        endDepot->twBefore = clients[0].tw;
         endDepot->twAfter = clients[0].tw;
```
N-Wouda, line 217: Question: do we need `startDepot->twBefore` and `endDepot->twAfter`?

pyvrp/educate/tests/test_Exchange.py
```diff
@@ -206,3 +211,41 @@ def test_relocate_after_depot_should_work():
...
+    assert_equal(individual, improved_indiv)
+
+    # Now doing it again with a large TW penalty, we must find the alternative
+    # solution
+    # (previously this was not the case since due to caching the current TW was
+    # computed as being zero, causing the move to be evaluated as worse)
+    cost_evaluator_tw = CostEvaluator(0, 1000)
+    improved_indiv = ls.search(individual, cost_evaluator_tw)
+    # expected_indiv = Individual(data, [[2, 1, 3, 4]])
+    # assert_equal(improved_indiv, expected_indiv)
```
N-Wouda, line 249:
```suggestion
```
(or add them in somehow?)

pyvrp/educate/tests/test_Exchange.py
```diff
@@ -206,3 +211,41 @@ def test_relocate_after_depot_should_work():
...
+    improved_cost = cost_evaluator_tw.penalised_cost(improved_indiv)
+    assert improved_cost < cost_evaluator_tw.penalised_cost(individual)
```
N-Wouda, line 251:
```suggestion
    assert_(improved_cost < cost_evaluator_tw.penalised_cost(individual))
```

pyvrp/educate/tests/test_Exchange.py
```diff
@@ -206,3 +211,41 @@ def test_relocate_after_depot_should_work():
     individual = Individual(data, [[1, 2, 3], [4]])
     expected = Individual(data, [[1, 2], [3], [4]])
     assert_equal(ls.search(individual, cost_evaluator), expected)
+
+
+def test_reoptimize_changed_objective_timewarp_OkSmall():
```
N-Wouda, line 216: Refactor: this test relates to LS, not Exchange. I think it belongs in `test_LocalSearch.py` instead.

pyvrp/cpp/educate/LocalSearch.cpp
wouterkool: Those values should always be equal to `depot->tw` and never change, and are independent of the individual or the route, so I think the best place would be in the constructor of local search (as the instance cannot change), not `loadIndividual` or `Route::update`. However we can treat that separate of this bugfix (if you agree)?

N-Wouda (approved):

Issue #56, opened 2023-01-30
How to deal with triangle inequality?


wouterkool:
Assuming the triangle inequality $d_{ij} <= d_{ik} + d_{kj} \ \forall i,j,k$ may allow for some optimizations, e.g.:
- assuming that removing one or more customers from a route will decrease cost, timewarp etc. this way we can avoid computing `TWS::merge` in some cases, e.g. if the current timewarp is 0.

The triangle inequality will hold for Euclidean cases (GH benchmarks) but in theory it can be violated for non-euclidean instances. Do we want to support 100% correctness of the local search operators (in that they never miss an improving move in the neighbourhood) or do we want to accept some edge cases will be skipped in favour of performance?

One idea could be a compiler switch ;).

N-Wouda:
My 2¢: I don't think it's worth introducing a benchmark-specific flag just for this, butI would rather focus efforts on things that improve performance in the general case.

N-Wouda:
(this seems pretty stale to me)

PR #213, opened 2023-05-02
Prize-collecting

### LINKED ISSUE #180: Prize collecting VRP?
We currently support the CVRP and VRPTW. For a paper I am working on, I need some form of prize collecting: in my problem, we do not have to visit every client, but instead get to choose for each client if the value of visiting is worth the expensive in travel time/distance. 

I am not sure if PyVRP should support this directly, but opening an issue to track it costs nothing. I will work on prize collecting in a branch or fork first (can I fork my own repo? 😄). If it's easy we might consider merging those changes in, if not, perhaps not.

I will use this issue to comment on that progress a bit.

N-Wouda:
This PR:

- [x] Closes #180.

## Benchmarks

VRPTW (e7c03a765c37b4c47b053926d5e85e625c1dc7ac; job ID 1444256; 10x seeds/1h runtime with `configs/vrptw.toml`):
```
 Avg. objective: 333466.6 (min: 333352, max: 333669)
Avg. iterations:  99344.8 (min: 71264, max: 107006)
```

PC-VRPTW (e7c03a765c37b4c47b053926d5e85e625c1dc7ac; job ID 1450185; 10x seeds/1h runtime with `configs/vrptw.toml`):
```
 Avg. objective: 212006.4 (min: 211981, max: 212097)
Avg. iterations: 353232   (min: 239154, max: 370375)
```

PC-VRPTW (fc78bce36efe2e07fc5fae06903f80c144a37062; job ID 1475226; 10x seeds/1h runtime with `configs/vrptw.toml`):
```
 Avg. objective: 211984.5 (min: 211908, max: 212040)
Avg. iterations: 375385   (min: 267300, max: 417939)
```

<details>

**Notes**:

</details>


N-Wouda:
> If it's helpful, some thoughts:

Sure, that's why I opened this draft so early.

> * Would it be an idea to implement prizes as penalties for unvisited nodes? This is equivalent (only differs by a constant) but would keep the objective non-negative and (for me) also easier to reason about when minimizing a cost objective. If we really want the 'correct' objective we could have `CostEvaluator` add the constant.

I don't think I understand how this'd work. Right now I'm going for an objective like (with binary $x_{ij}$ indicating if edge $(i, j)$ is used or not):

image

but then in min form, so minimize distance - prizes. How would your proposal achieve this?

> * What is your idea behind virtual routes? Can't we just unassign a node altogether (using a boolean property or making `Node->route` optional)? I think we still need to treat unassigned customers specially when computing deltas?

Keeping a "virtual" (not sure on terminology yet, do you have ideas?) route around lets us keep all the existing machinery for operators with very limited need for change (I think, I'm still working on this). We can move stuff in and out of virtual routes, etc. I wasn't sure on how to do, say, general $(N, M)$-exchange when unassigned customers are just unassigned, and not part of such a virtual route.

wouterkool:
> > If it's helpful, some thoughts:
> 
> Sure, that's why I opened this draft so early.
> 
> > * Would it be an idea to implement prizes as penalties for unvisited nodes? This is equivalent (only differs by a constant) but would keep the objective non-negative and (for me) also easier to reason about when minimizing a cost objective. If we really want the 'correct' objective we could have `CostEvaluator` add the constant.
> 
> I don't think I understand how this'd work. Right now I'm going for an objective like (with binary xij indicating if edge (i,j) is used or not):
> 
> image
> 
> but then in min form, so minimize distance - prizes. How would your propose to model this goal?

Without the fancy math notation, I'd propose
`min_x [distance objective] + sum_i p_i (1 - sum_j x_ij)`
i.e. flip your objective (to minimize rather than maximize) and add `sum_i p_i` to make the objective always non-negative. It can be interpreted as incurring a penalty `p_i` if the node is not visited (which, up to a constant, is equivalent to getting a prize if the node is visited)


> 
> > * What is your idea behind virtual routes? Can't we just unassign a node altogether (using a boolean property or making `Node->route` optional)? I think we still need to treat unassigned customers specially when computing deltas?
> 
> Keeping a "virtual" (not sure on terminology yet, do you have ideas?) route around lets us keep all the existing machinery for operators with very limited need for change (I think. I'm still working on this. This PR contains a partial implementation for Exchange). We can move stuff in and out of virtual routes, etc. I wasn't sure on how to do, say, general (N,M)-exchange when unassigned customers are just unassigned, and not part of such a virtual route.
> 
> Do you have ideas for this?

The (N,M) exchange considers sequences  of N and M adjacent nodes. For unassigned nodes, I think there is no meaningful definition of a sequence of adjacent nodes, so we should only consider exchanging N assigned nodes with M = 0 or M = 1 unassigned nodes (or vice versa). Additionally, when M = 0, it is irrelevant which node is V, so the same move will be checked repeatedly (so it could be considered to make unassign a special operator).

I don't think the current implementation of virtual route solves the ambiguity of which M unassigned nodes to consider adjacent. Therefore, we still cannot sensibly move more than 1 node from a virtual route (as the prev/next for each node is the depot?), so I see no benefit of the virtual route? Or am I missing something?

Additionally, I think having one virtual route with different properties (i.e. multiple nodes connected to the depot) creates more edge cases/trouble than it solves. One alternative could be having `node->prev = node->next = node`, which would make the delta distance for removal 0 automatically (so no need to handle it seperately I think), but still I think it would be better to make it explicit. Accounting for one extra route by replacing the `<` condition by `<=` in for loops is, in my opinion implicit and unclear. 

Overall, I think making the assign/unassign functionality explicit without virtual route and just special casing the operators where necessary would make things much cleaner (I think this is also roughly what Kleopatra did but I'm not 100% sure).

N-Wouda:
OK, so this is roughly +400 - 75, or a little over 300 lines. Not too bad, but we need a bit more. So say 400, with little regard for density. I'm going to try simply inserting/removing next.

N-Wouda:
I think this is now ready for a review. For benchmarking:
1. I don't know what to benchmark against;
2. We need to agree on https://github.com/PyVRP/Instances/pull/1.

leonlan:
>     1. I don't know what to benchmark against;

Should we benchmark against https://github.com/tumBAIS/euro-meets-neurips-2022/tree/main/pchgs?
 
>     2. We need to agree on [Prize-collecting instances Instances#1](https://github.com/PyVRP/Instances/pull/1).

What kind of agreement are you looking for here? PC-CVRP is from the literature, so that's fine by me. PC-VRPTW follows a generation procedure from the literature, which also sounds OK to me. If needed, the latter can still be changed. 

The benchmarking will probably take multiple iterations anyway, so it's good if we have some results already for reference.

N-Wouda:
> I'll probably review this tomorrow.
> 
> > ```
> > 1. I don't know what to benchmark against;
> > ```
> 
> Should we benchmark against https://github.com/tumBAIS/euro-meets-neurips-2022/tree/main/pchgs?
> 
> > ```
> > 2. We need to agree on [Prize-collecting instances Instances#1](https://github.com/PyVRP/Instances/pull/1).
> > ```
> 
> What kind of agreement are you looking for here? PC-CVRP is from the literature, so that's fine by me. PC-VRPTW follows a generation procedure from the literature, which also sounds OK to me. If needed, the latter can still be changed.
> 
> The benchmarking will probably take multiple iterations anyway, so it's good if we have some results already for reference.
> 
> Also, running a benchmark on VRPTW would be helpful to check if there are any regressions on that side!

N-Wouda:
> Should we benchmark against https://github.com/tumBAIS/euro-meets-neurips-2022/tree/main/pchgs?

Perhaps, but.. that implementation can only deal with explicit edge weight matrices, and uses a somewhat different input and output format. 

> What kind of agreement are you looking for here? PC-CVRP is from the literature, so that's fine by me. PC-VRPTW follows a generation procedure from the literature, which also sounds OK to me. If needed, the latter can still be changed.
> 
> The benchmarking will probably take multiple iterations anyway, so it's good if we have some results already for reference.

Yeah, I was looking for something like this. 

> Also, running a benchmark on VRPTW would be helpful to check if there are any regressions on that side!

That one's already underway!

N-Wouda:
VRPTW benchmark is done and looks to be in line with current main. That's more or less what I expected, because the changes here should not have much of a performance impact (and indeed do not appear to have any).

I'm now running a benchmark with the PC-VRPTW instances of https://github.com/PyVRP/Instances/pull/1 to generate BKS.

pyvrp/cpp/Individual.cpp
```diff
@@ -74,11 +82,15 @@ void Individual::evaluate(ProblemData const &data)
 
         if (static_cast<size_t>(routeLoad) > data.vehicleCapacity())
             excessLoad_ += routeLoad - data.vehicleCapacity();
+
+        uncollected_ -= routePrize;
```
wouterkool, line 86: Question: should we also check that all required customers are visited? (Related: we also don't check currently that we visit each client once).

pyvrp/_Individual.pyi
```diff
@@ -123,6 +123,16 @@ class Individual:
         int
             Total time warp over all routes.
         """
+    def uncollected(self) -> int:
```
wouterkool, line 126: Maybe a more descriptive name? This name could as well represent the list of unvisited clients. How about `total_uncollected_prizes`?

pyvrp/cpp/Individual.h
```diff
@@ -37,6 +39,11 @@ class Individual
      */
     [[nodiscard]] size_t numRoutes() const;
 
+    /**
+     * Number of clients in the solution.
+     */
+    [[nodiscard]] size_t size() const;
```
wouterkool, line 45: Question: wouldn't `numClients` be a better name and consistent with `numRoutes`? Size is ambiguous, why not measure the size of a solution by the number of routes? (not sure)

pyvrp/cpp/educate/LocalSearch.cpp
```diff
@@ -190,16 +268,20 @@ void LocalSearch::update(Route *U, Route *V)
...
                               data.client(client).twEarly,
                               data.client(client).twLate};
 
+        clients[client].route = nullptr;  // nullptr implies "not in solution"
+    }
+
     auto const &routesIndiv = individual.getRoutes();
 
-    for (size_t r = 0; r < data.numVehicles(); r++)
+    for (size_t r = 0; r != data.numVehicles(); r++)
```
wouterkool, line 284: Just curious, is there a reason you prefer `!=` over `<`?

pyvrp/cpp/educate/Node.cpp
```diff
@@ -34,3 +40,16 @@ void Node::swapWith(Node *other)
     route = routeV;
     other->route = routeU;
 }
+
+void Node::remove()
+{
+    auto *UPred = prev;
```
wouterkool, line 46: Do you need the intermediate variables?

wouterkool (approved):
This looks very neat to me and much simpler than the virtual routes, nice job!

Two suggestions/future nice to haves, just to have them out of my head:
- Add template parameter (like exchange) to remove 2 or 3 nodes at once
- SREX does not need to reinsert optional customers

pyvrp/cpp/educate/LocalSearch.cpp
```diff
@@ -172,6 +186,70 @@ bool LocalSearch::applyRouteOps(Route *U,
...
+    int const proposed = data.dist(V->client, U->client)
+                         + data.dist(U->client, n(V)->client) - uClient.prize;
+
+    int deltaCost = proposed - current;
+
+    deltaCost += costEvaluator.loadPenalty(V->route->load() + uClient.demand,
+                                           data.vehicleCapacity());
+    deltaCost
+        -= costEvaluator.loadPenalty(V->route->load(), data.vehicleCapacity());
+
```
wouterkool, line 207: We could add a precheck here
```suggestion
    if (deltaCost - V->route->timeWarp() >= 0)
        return deltaCost
```

pyvrp/cpp/educate/LocalSearch.cpp
```diff
@@ -172,6 +186,70 @@ bool LocalSearch::applyRouteOps(Route *U,
...
+    int const proposed = data.dist(p(U)->client, n(U)->client);
+
+    int deltaCost = proposed - current;
+
+    deltaCost += costEvaluator.loadPenalty(U->route->load() - uClient.demand,
+                                           data.vehicleCapacity());
+    deltaCost
+        -= costEvaluator.loadPenalty(U->route->load(), data.vehicleCapacity());
+
```
wouterkool, line 238: We could add a precheck here
```suggestion
    if (deltaCost - U->route->timeWarp() >= 0)
        return deltaCost
```

pyvrp/cpp/educate/LocalSearch.cpp
wouterkool: Maybe even
```suggestion
    if (deltaCost >= 0)
        return deltaCost
```
if we want to assume the timewarp can only increase by adding a customer, but this relies on subtle assumptions (like triangle inequality, see https://github.com/PyVRP/PyVRP/issues/56).

pyvrp/_Individual.pyi
N-Wouda: I've updated this to `prizes()` and `uncollected_prizes()`, which seems fairly clear to me. I'd rather not add `total_` as a prefix, because we also don't do that for the other data attributes on an `Individual`.

N-Wouda:
>SREX does not need to reinsert optional customers

That's a good one. I've added a check to `greedyRepair` to only insert if a customer is required or cost < prize.

pyvrp/cpp/Individual.cpp
N-Wouda: I wasn't sure about this. On the one hand, yes, because else the solution is nonsensical. On the other hand, no, because having to check that every time could be time consuming. I'll add it in and run another benchmark later.

pyvrp/cpp/Individual.h
N-Wouda: Yes, that works better. I was thinking about how to detect empty solutions earlier and then figured "oh what about implementing `__len__`", but `individual.num_clients()` works just as well.

pyvrp/cpp/educate/LocalSearch.cpp
N-Wouda: I've opted for the second form. We probably won't lose much if we assume time warp does not go down, but I'd rather investigate that a bit later in a separate issue/PR. Strengthening these checks is probably worth it, but is probably a topic in and of itself.

N-Wouda: OK, no, I've tinkered with this a bit more and it does seem to be worthwhile doing the check involving time warp. I'm doing that, and will then run another benchmark to see if it helps.

pyvrp/cpp/educate/LocalSearch.cpp
```diff
@@ -35,15 +39,25 @@ Individual LocalSearch::search(Individual &individual,
...
             auto const lastTestedNode = lastTestedNodes[uClient];
             lastTestedNodes[uClient] = numMoves;
 
+            if (U->route && !data.client(uClient).required)  // test removing U
+                maybeRemove(U, costEvaluator);
+
             // Shuffling the neighbours in this loop should not matter much as
             // we are already randomizing the nodes U.
             for (auto const vClient : neighbours[uClient])
```
leonlan, line 51: In another PR, we should probably incorporate prizes in the granular neighborhood. I recall you mentioned this earlier somewhere but can't find it any longer. Should we make an issue of this?

More general: to what extent are you planning to add features to improve the performance on PC instances? Or is this enough?

leonlan (approved):
Very neat implementation, nice job!

pyvrp/cpp/educate/LocalSearch.cpp
N-Wouda: I'm going to make an issue collecting some of the things we discussed here. For prize collecting, my main improvement will through the neighbourhoods (for PyVRP I'm aiming for something simple, but also for another unrelated paper I'll also work on something like this). I'll try to also benchmark against Kleopatra's implementation, but that will take a few more days (and need not be part of this PR).

So basically only the neighbourhoods, unless we find that Kleopatra's implementation is significantly better than this simple thing.

N-Wouda:
I tried to change `crossover::greedyRepair` to take prizes into account, but that didn't quite work all that well (got some erros when running it). I've reverted that for now, and added it as future work to #222.

N-Wouda:
I'll wait until the checks complete and then merge this in. The benchmark against Kleopatra's implementation will go into #222, and the neighbourhood stuff into #215 and a new PR (soon).

pyvrp/cpp/Individual.cpp
N-Wouda: This does not appear to be too expensive, so I think we can leave it in. It's a good check to have!

PR #214, opened 2023-05-02
Sort granular neighbourhood by proximity

### LINKED ISSUE #201: Order neighbours by proximity?
In LocalSearch, we shuffle the order of the nodes, but not the order of the neighbours since arguably this does not change much if we already shuffle the node. However, since the neighbours are sorted by index this makes lower index nodes much more often checked.

I suggest to try the suggestion by Team Kleopatra (see #181) to sort the neighbours by proximity, which they claim yields faster convergence as the best moves are checked and thus applied first.

N-Wouda:
This PR:

- [x] Closes #201, part of #181.

### Benchmarks

Current main (45e30f9bbc0cf6adaa1e7d5c95b0a9b43157664a, job ID 959880)
```
 Avg. objective: 333444.8 (min: 333296, max: 333585)
Avg. iterations: 102187.1 (min: 75036, max: 111424)
```
This branch (cdaf1f70c35947d1392aae326ec3cccd2fe3480f, job ID 960271)
```
 Avg. objective: 333434.3 (min: 333352, max: 333642)
Avg. iterations: 102605.7 (min: 83489, max: 108594)
```
This branch (30f468584e7bec313576fd33df18efb035674fcb, job ID 1041857)
```
 Avg. objective: 333556.2 (min: 333387, max: 333768)
Avg. iterations:  88005.9 (min: 73832, max: 104752)
```

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/educate/neighbourhood.py
+++ b/pyvrp/educate/neighbourhood.py
@@ -89,16 +89,16 @@ def compute_neighbours(
     proximity[0, :] = np.inf  # depot has no neighbours
     proximity[:, 0] = np.inf  # clients do not neighbour depot
 
-    topk = np.argsort(proximity, axis=1, kind="stable")[1:, :k]  # excl. depot
+    top_k = np.argsort(proximity, axis=1, kind="stable")[1:, :k]  # excl. depot
 
     if not params.symmetric_neighbours:
-        return [[]] + topk.tolist()
+        return [[]] + top_k.tolist()
 
     # Construct a symmetric adjacency matrix and return the adjacent clients
     # as the neighbourhood structure.
     adj = np.zeros_like(proximity, dtype=bool)
     rows = np.expand_dims(np.arange(1, n), axis=1)
-    adj[rows, topk] = True
+    adj[rows, top_k] = True
     adj = adj | adj.transpose()
 
     return [np.flatnonzero(row).tolist() for row in adj]
@@ -109,7 +109,7 @@ def _compute_proximity(
 ) -> np.ndarray[float]:
     """
     Computes proximity for neighborhood. Proximity is based on Vidal et al.
-    (2012).
+    (2013).
 
     Parameters
     ----------
--- a/pyvrp/educate/neighbourhood.py
+++ b/pyvrp/educate/neighbourhood.py
@@ -18,17 +18,21 @@ class NeighbourhoodParams:
     Attributes
     ----------
     weight_wait_time
-        TODO
+        Weight given to the minimum wait time aspect of the proximity
+        calculation. A large wait time indicates the clients are far apart
+        in duration/time.
     weight_time_warp
-        TODO
+        Weight given to the minimum time warp aspect of the proximity
+        calculation. A large time warp indicates the clients are far apart in
+        duration/time.
     nb_granular
         Number of other clients that are in each client's granular
         neighbourhood. This parameter determines the size of the overall
         neighbourhood.
     symmetric_proximity
-        TODO
-    symmetric_neighbours
-        TODO
+        Whether to calculate a symmetric proximity matrix. A symmetric matrix
+        ensures the granular neighbourhood is symmetric as well: if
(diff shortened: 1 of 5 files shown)


pyvrp/educate/neighbourhood.py
```diff
@@ -66,39 +69,32 @@ def compute_neighbours(
...
+    k = min(params.nb_granular, n - 2)  # excl. depot and self
 
-        # Add correlated vertex if correlated in one of two directions
-        adj = adj | adj.transpose()
+    np.fill_diagonal(proximity, np.inf)  # cannot be in own neighbourhood
+    proximity[0, :] = np.inf  # depot has no neighbours
+    proximity[:, 0] = np.inf  # clients do not neighbour depot
 
-        # Append empty neighbours for depot and add 1 to offset depot indexing
-        return [[]] + [(np.flatnonzero(row) + 1).tolist() for row in adj]
```
N-Wouda, line 92: I don't think we need this. I think the stable sort already covers this.

N-Wouda:
I'll run a benchmark overnight to check whether this is better or not.

N-Wouda:
Runs are a lot more consistent with this change in (variation between seeds has gone down quite a bit). That could indicate early convergence, so I'm also testing a run where we lower the number of iterations before restart from 20K to 10K.

N-Wouda:
10K is definitely not better. I'm reverting to 20K.

N-Wouda:
@leonlan @wouterkool this does not appear to make much of a difference either way. I'm weakly in favour of merging this in, because intuitively it makes sense to me to explore the most promising edges first, rather than just those on the lower indices. The performance of both approaches is fairly similar though.

I also took this moment to add some documentation to `NeighbourhoodParams` and the various functions in `neighbourhood.py`, and add a test for the symmetry bit.

wouterkool:
It looks good and very neat to me! I agree on the conclusion that ordering by proximity is less arbitrary so I also think we should merge it. Should we keep an issue arround to 'not forget' this may be relevant? I can imagine always considering best edges first can limit the search although this is quite a research topic in itself.

wouterkool (approved):

wouterkool:
BTW as a bit of historical context, the tie-braker was there to exactly mimic the previous c++ implementation (helpful for debugging). Given that it is no longer the case I don't think it is strictly necessary to use stable sort, although I suggest to keep it to have a fully defined ordering.

pyvrp/educate/neighbourhood.py
```diff
@@ -66,39 +73,43 @@ def compute_neighbours(
...
 
 
 def _compute_proximity(
-    data: ProblemData, params: NeighbourhoodParams = NeighbourhoodParams()
+    data: ProblemData, weight_wait_time: float, weight_time_warp: float
 ) -> np.ndarray[float]:
     """
-    Computes proximity for neighborhood.
+    Computes proximity for neighborhood. Proximity is based on Vidal et al.
+    (2012).
```
leonlan, line 112:
```suggestion
    (2013).
```

pyvrp/educate/neighbourhood.py
```diff
@@ -66,39 +73,43 @@ def compute_neighbours(
...
+
+    np.fill_diagonal(proximity, np.inf)  # cannot be in own neighbourhood
+    proximity[0, :] = np.inf  # depot has no neighbours
+    proximity[:, 0] = np.inf  # clients do not neighbour depot
 
-    if params.symmetric_neighbours:
-        # Convert into adjacency matrix that can be symmetrized
-        adj = np.zeros_like(proximity, dtype=bool)
-        adj[idcs[:, None], idx_topk] = True
+    topk = np.argsort(proximity, axis=1, kind="stable")[1:, :k]  # excl. depot
```
leonlan, line 92: Nit: could we rename this to `top_k`?

N-Wouda: Sure!

PR #216, opened 2023-05-10
Add PyVRP/Instances submodule

N-Wouda:
This PR adds a submodule for self-hosted instances, rather than assuming the user can download these. This is not that relevant for CVRPLIB instances, but will become more important when we need benchmarks for problem variants where no good external defaults exist (e.g., prize collecting in https://github.com/PyVRP/Instances/pull/1).

<details>

**Notes**:

</details>

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "instances"]
+	path = instances
+	url = https://github.com/PyVRP/Instances
--- a/docs/source/dev/benchmarking.rst
+++ b/docs/source/dev/benchmarking.rst
@@ -10,8 +10,7 @@ To find out about available options, run
    pyvrp --help
 
 To benchmark ``pyvrp``, we mostly use instances from `CRVPLIB <http://vrp.atd-lab.inf.puc-rio.br/index.php/en/>`_ instances.
-Each instance is run on a single core.
-We use the following instances:
+Particularly:
 
 * For the VRPTW, we use `the Gehring and Homberger instances <http://vrp.atd-lab.inf.puc-rio.br/media/com_vrp/instances/Vrp-Set-HG.tgz>`_, particularly those with 1000 customers.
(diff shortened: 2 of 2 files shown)


N-Wouda:
@wouterkool @leonlan I think we can also add the 100 final Euro/NeurIPS instances to https://github.com/PyVRP/Instances?

leonlan (approved):

leonlan:
> We'd have to update them;

We should not do it for CVRPLIB instances, because they are keeping track of the BKS themselves. But I'm in favor of doing it for instances that we maintain ourselves, e.g., the PCVRP instances. We could set up a BKS submission workflow and let other users submit PRs with the new best solutions.

N-Wouda:
I think that's a reasonable default. We provide BKS for benchmark instances that do not exist elsewhere, and assume users can obtain BKSs for instances maintained by outside parties (e.g., CVRPLIB).

leonlan:
> @wouterkool @leonlan I think we can also add the 100 final Euro/NeurIPS instances to [PyVRP/Instances](https://github.com/PyVRP/Instances)?
> 
> (if we use those for benchmarking)

Forgot the reply on this one. I don't think we should use Euro/NeurIPS instances for benchmarking because they are too small (most are 200-500 customers) and they not established yet in the literature. 

I do think there is a need for new VRPTW benchmark instances, perhaps by making larger instances based on the Euro-NeurIPS data. I suggested this to @wouterkool a long time ago, are there any plans to work on that?

N-Wouda:
I'm going to merge this in for now

> Forgot the reply on this one. I don't think we should use Euro/NeurIPS instances for benchmarking because they are too small (most are 200-500 customers) and they not established yet in the literature.

Sure, although I don't think very small instances are useless (if anything to check we can solve those almost all to (near) optimality). But if we do want small instances, other established benchmarks might be a better go-to.

N-Wouda:
I think we can merge this in soon. The prize-collecting instances are also mostly done (in https://github.com/PyVRP/Instances/pull/1), but we do not need to wait for that to merge this.

PR #218, opened 2023-05-13
v0.2.0 benchmark

leonlan:
Gaps of CVRP went up a bit, VRPTW went down a bit.


N-Wouda:
Great, thanks! I'm merging this in and will release a new version!

N-Wouda:
This one's against `instances`, not `main`. I've opened #219 as the same PR, but then against main.

PR #219, opened 2023-05-13
v0.2.0 benchmark results

N-Wouda:
See #218.


N-Wouda:
Nvm I got this wrong. #218 was based on `instances`, so it makes sense that one would merge back into #216. I've removed that dependency in #220 and merged that into main, so we can wait a little while longer before merging the instances into main.

PR #221, opened 2023-05-14
Route object

N-Wouda:
This PR adds a `Route` object that stores the route plan, and some statistics. This intends to replace the separate diagnostics module.


N-Wouda:
I think this is ready for review now.

pyvrp/cpp/Individual.h
```diff
@@ -4,15 +4,61 @@
...
+public:
+    /**
+     * A simple Route class that contains the route plan (as a vector), and
+     * some route statistics.
+     */
+    class Route
+    {
+        using Plan = std::vector<Client>;
+
+        Plan plan_ = {};         // Route plan (list of clients)
```
N-Wouda, line 25: I'm unsure about the name. We can probably do better. Maybe `tour`?

pyvrp/cpp/Individual.cpp
```diff
@@ -170,18 +120,110 @@ Individual::Individual(ProblemData const &data, Routes routes)
...
+        }
+
+        prevClient = plan_[idx];
     }
 
+    // Last client has depot as successor.
+    distance_ += data.dist(plan_.back(), 0);
+    duration_ += data.duration(plan_.back(), 0);
+    time += data.client(plan_.back()).serviceDuration
+            + data.duration(plan_.back(), 0);
```
leonlan, line 161: Refactor: maybe assign variable `last = plan.back()` here?

pyvrp/cpp/Individual.h
leonlan: - I find `plan` a bit vague, mostly because it's terminology I haven't heard before in the context of VRPs. First question that came to mind when reading this name was "a plan of what?" to which my answer was a plan of client visits.
	- How about simply renaming to `clients`? Does this add too much confusion?
	- Otherwise, what about `visits`? This also generalizes easily to more general vehicle routing problems (for example, pickups and deliveries).
- `tour` is OK with me as well for now but it implies that it always start/end at the same depot. If we ever allow for open VRPs/multiple depots then this might need to be changed again.

leonlan (approved):

N-Wouda:
@leonlan thanks, I'll have a look at this either tonight or tomorrow.

pyvrp/cpp/Individual.cpp
N-Wouda: Yes, good call!

pyvrp/cpp/Individual.h
N-Wouda: Of these options, I think I like `visits` the most. I'll implement that now.

examples/vrptw.ipynb
```diff
@@ -266,149 +270,15 @@
    ]
   },
   {
+   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Inspect routes\n",
```
wouterkool, line 272: I'm fine with replacing the diagnostics module, but can we somehow still print statistics per route? I think that is quite useful?

pyvrp/cpp/Individual.cpp
```diff
@@ -170,18 +120,112 @@ Individual::Individual(ProblemData const &data, Routes routes)
...
 
+    Client const last = visits_.back();  // lLast client has depot as successor
+    distance_ += data.dist(last, 0);
+    duration_ += data.duration(last, 0);
+    time += data.client(last).serviceDuration + data.duration(last, 0);
+    timeWarp_ += std::max(time - data.depot().twLate, 0);  // depot closing tw
+    excessLoad_ = data.vehicleCapacity() < demand_
+                      ? demand_ - data.vehicleCapacity()
+                      : 0;
+}
```
wouterkool, line 165: Doesn't have to be part of this PR but given that you're touching it now I would like to make the suggestion of using the TimeWindowSegment class to do the computations. That would eliminate the redundant logic of wait time/time warp computation and additionally we would also get the earliest/latest start times of the routes which is quite useful. What do you think of that (I can make it a separate issue)?

pyvrp/cpp/Individual_bindings.cpp
```diff
@@ -10,8 +10,49 @@ namespace py = pybind11;
 
 PYBIND11_MODULE(_Individual, m)
 {
+    py::class_<Individual::Route>(m, "Route")
+        .def(py::init<ProblemData const &, std::vector<int>>(),
+             py::arg("data"),
+             py::arg("plan"))
```
wouterkool, line 16:
```suggestion
             py::arg("visits"))
```

wouterkool (approved):
Few small comments but looks very nice overall to me! :)

pyvrp/cpp/Individual.cpp
N-Wouda: Yeah that could work! It's probably best to handle this separately through a new issue. When we start working on that issue it could also be a good time to have another look at how TWS is currently used (and see if we can maybe get rid of its default constructor, since that thing makes no sense).

N-Wouda: Can you open an issue for this?

examples/vrptw.ipynb
N-Wouda: I forgot to replace this bit with something new again. I'm going to do that now.

N-Wouda: I've added such a table back into the notebook in my most recent commit.

N-Wouda:
Alright, I'll merge this in once the checks pass. @leonlan @wouterkool thanks for the review!

PR #233, opened 2023-06-03
Re-enable unused parameter build warnings

N-Wouda:
I recently learned there's a `[[maybe_unused]]` attribute in C++ that can be used to indicate parameters might be unused. Some parameters aren't used everywhere (e.g., CVRP does not use a lot of the TW related parameters). To deal with that we had disabled the unused parameter warnings.

This PR instead uses that attribute where needed, and re-enables the compiler warnings we previously disabled.

<details>

**Notes**:

</details>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -24,22 +24,10 @@ if get_option('buildtype') == 'debug'
     add_project_link_arguments('--coverage', language: 'cpp')
 endif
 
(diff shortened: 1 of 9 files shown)


N-Wouda:
I'm merging this in as-is because it's purely build related, and does not change anything functionality-wise. Re-enabling some warnings is not controversial, I think.

PR #234, opened 2023-06-03
Add support for double precision

### LINKED ISSUE #47: Compute using double precision

N-Wouda:
This PR:

- [x] Closes #47.

### Benchmarks

All on VRPTW.

Integer precision (bd1b87e97c29cad55fd3f07b9da127d0266b1003; job ID 2370987; 10x seeds/1h with `configs/vrptw.toml`):
```
 Avg. objective: 333303.7 (min: 333107, max: 333384)
Avg. iterations: 102962.3 (min: 101098, max: 104870)
```
So this is basically unchanged.

Double precision (a2de0879a383903b884faa34fa36fcadc44369f4; job ID 2372134; 10x seeds/1h with `configs/vrptw.toml`):
```
 Avg. objective:  33418.1 (min: 33405, max: 33433)
Avg. iterations:  80966.0 (min: 77283, max: 87658)
```

<details>

**Notes**:

</details>

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,23 +15,20 @@ jobs:
     name: >
       Build and test pyvrp using ${{ matrix.compiler }} 
       ${{ matrix.compiler-version }} and Python ${{ matrix.python-version }}
+      in ${{ matrix.precision }} precision mode
     runs-on: ubuntu-latest
     strategy:
       fail-fast: true
       matrix:
-        include:
-          - compiler: gcc
+        compiler: [gcc, clang]
+        compiler-version: ['10', '12', 'latest']
+        python-version: ['3.8', '3.11']
+        precision: [integer, double]
+        exclude:
+          - compiler: clang
             compiler-version: '10'  # minimum gcc version
-            python-version: '3.8'   # minimum Python version
           - compiler: gcc
-            compiler-version: latest
-            python-version: '3.11'
-          - compiler: clang
             compiler-version: '12'  # minimum clang version
-            python-version: '3.8'   # minimum Python version
-          - compiler: clang
-            compiler-version: latest
-            python-version: '3.11'
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python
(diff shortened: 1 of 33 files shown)


wouterkool (approved):

N-Wouda:
No new SINTEF BKS. We do match a few known BKS, but without the vehicle restriction we're not going to push the limits there.

N-Wouda:
This is more or less the simplest thing I could think of. A small (absolute) tolerance of the same value as used in HGS-CVRP (1e-5). The hashing isn't perfect but it seems to work OK.

We pay a small price for all the extra comparison work (# iters has gone down quite a bit when comparing double with integer precision), but the results are fairly competitive. I'm OK with this as is.

N-Wouda:
@wouterkool @leonlan we never really discussed this, but should the default remain integer precision? I think yes, because:
1. We already used to do it that way;
2. It's more performant;
3. There's no meaningful loss in precision in the real world when working with for example meters and seconds.

.github/workflows/CI.yml
```diff
@@ -15,23 +15,34 @@ jobs:
...
+        precision: [integer, double]
+        exclude:
+          - compiler: clang
             compiler-version: '10'  # minimum gcc version
-            python-version: '3.8'   # minimum Python version
           - compiler: gcc
-            compiler-version: latest
-            python-version: '3.11'
-          - compiler: clang
             compiler-version: '12'  # minimum clang version
```
leonlan, line 31: The `minimum <compiler> version` comments are a bit outdated since we exclude versions now. Maybe update to something like ` test only 12 and latest ` for clang and ` test only 10 and latest ` for GCC?

leonlan (approved):

.github/workflows/CI.yml
N-Wouda: This hasn't changed compared to earlier CI runs - we already only ran 10/latest for GCC, and 12/latest for clang.

PR #237, opened 2023-06-10
Base route angle on centroid of client locations

N-Wouda:
This PR modifies the angle computation to be relative to the center point of all client locations, rather than the depot location. That should result in a more effective route sorting and crossover when the depot is not centrally located.

### Benchmarks

All on the X instances, where the depot's location varies quite a bit between instances.

Current main (3e90a1008fc6380f66af546cc9ed0b1d510ef66d; job ID 2505221; 10x seeds, $2.4n$ runtime with `configs/cvrp.toml`):
```
 Avg. objective: 63317.1 (min: 63301.6, max: 63342.8)
Avg. iterations: 100779.6 (min: 83520, max: 105537)
```

This PR (99c977915118e77d1a1a3408d5c95f3c3f661c08; job ID 2503917 and 2504107; 10x seeds, $2.4n$ runtime with `configs/cvrp.toml`):
```
 Avg. objective: 63302.5 (min: 63293.1, max: 63311)
Avg. iterations: 89561.6 (min: 82269, max: 104784)
```

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/_ProblemData.pyi
+++ b/pyvrp/_ProblemData.pyi
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Tuple
 
 class Client:
     """
@@ -101,6 +101,15 @@ class ProblemData:
         -------
         Client
             A simple data object containing the depot's information.
+        """
+    def centroid(self) -> Tuple[float, float]:
+        """
+        Center point of all client locations (excluding the depot).
+
+        Returns
+        -------
+        tuple
+            Centroid of all client locations.
         """
     def dist(self, first: int, second: int) -> int:
         """
--- a/pyvrp/cpp/ProblemData.cpp
+++ b/pyvrp/cpp/ProblemData.cpp
@@ -38,6 +38,11 @@ ProblemData::Client::Client(Coordinate x,
 
 ProblemData::Client const &ProblemData::depot() const { return client(0); }
 
+std::pair<double, double> const &ProblemData::centroid() const
+{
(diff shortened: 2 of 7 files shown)


N-Wouda:
In the GH instances the depot is very central, so this change doesn't do anything. However, the X instances do have non-central depots. I'll run a benchmark on the CVRP instances to check.

leonlan (approved):

wouterkool:
I missed the chance to review but I still wanted to share my thoughts.

I think it is quite typical to see 'pie slices' in the solution, even if the depot is in the corner. Taking the global centroid as origin for the route angle means it can change 180 degrees if the route centroid slightly changes towards/away from the depot. 

Ultimately the centroid/circle sector is taken as a proxy for route similarity/overlap, which I think would be better if it is robust to small changes. So could it be that this is not a better proxy per se? Given the reduced number of iterations, my hypothesis is we get 'worse' solutions out of SREX taking 10% more local search. Now the overall result is still (slightly but significant) better which could actually be due to added diversity (we have seen this before) which is fine but a rather indirect/unintended effect maybe?

Also, I am a bit concerned about drawing conclusions based on a single number, how are the convergence curves and does the result depend on size of instances? I don't question the numbers but I'm wondering whether a change that reduces iterations by 10% (despite the improvement in results) should be worth some more investigation since it is unclear if the net effect may actually be negative in some cases. What do you think?

As an extreme idea what would happen if we randomly order routes?

PR #238, opened 2023-06-11
Remove LS dependency on rng, use separate shuffle method

N-Wouda:
This PR is part of #157.

<details>

**Notes**:

</details>


N-Wouda:
My cluster is out for the day, but some simple timeit's on `ls.search` and `ls.intensify` suggest this does not make any difference in speed.

wouterkool (approved):
I think this is a nice improvement overall which can be merged in. :) Some thoughts, maybe for later:

How about (additionally) binding functions that would take an explicit order for nodes/routes and maybe operators? That would align nicely with how we can set the neighbourhood from Python and give more flexibility for the user:
- one could try smarter strategies than random shuffling without touching c++. 
- Additionally, if we would allow `orderNodes` to be a subset this could also be used to focus search on a part of the solution.
- If all dependencies on rng are replaced by explicit values we could replace the XorShift implementation by the standard rng from numpy and just pass a numpy vector to c++

leonlan:
> How about (additionally) binding functions that would take an explicit order for nodes/routes and maybe operators? That would align nicely with how we can set the neighbourhood from Python and give more flexibility for the user.

I think this is nice to have.

> Additionally, if we would allow orderNodes to be a subset this could also be used to focus search on a part of the solution.

This is already covered by the granular neighborhood if we set an empty neighborhood for some client.

pyvrp/cpp/educate/LocalSearch.h
```diff
@@ -95,15 +94,19 @@ class LocalSearch
...
+     * individual, and returns a new, hopefully improved individual.
      */
     Individual intensify(Individual &individual,
                          CostEvaluator const &costEvaluator,
                          int overlapToleranceDegrees = 0);
 
-    LocalSearch(ProblemData &data, XorShift128 &rng, Neighbours neighbours);
+    /**
+     * Shuffles the order in which the node and route pairs are evaluated.
+     */
```
leonlan, line 106: Suggestion: also add something about the operators like "and the order in which the node and route operators are applied."

PR #241, opened 2023-06-12
Use TWS for route timing statistics computation

wouterkool:
This PR:
* Use `TimeWindowSegment` to compute timings in `Individual::Route`.
* `Individual::Route` now computes `waitDuration` assuming the optimal starttime.
* `Individual::Route` now has `travelDuration`, `serviceDuration`, `waitDuration` and `totalDuration`
* `Individual::Route` now has `earliestStart` and `latestStart`
* Rename `TWS::totalTimeWarp` to `TWS::timeWarp`
* Add `_` to `TWS` properties
* Make indices in `TWS` `size_t`
* Add convenient constructor to `TWS` for single client

<details>

**Notes**:

</details>


N-Wouda:
I don't have much time today, so I cannot review this in full now. But for the route statistics: I think it's very helpful if we do that as `start_time()` (or similar; is the earliest start time) and `slack()`. If a user cares about the latest start time they can compute that as start + slack. For users that do not care a fixed start allows us to compute derived values like the `end_time()`, and e.g. arrival/departure times at clients, which are generally useful to know. Those can then all be shifted to the future by at most `slack()`, if a user wants.

pyvrp/cpp/Individual.cpp
```diff
@@ -238,13 +227,22 @@ Load Individual::Route::demand() const { return demand_; }
...
+{
+    return tws_.duration() - travel_ - service_;
+}
+
+Duration Individual::Route::totalDuration() const { return tws_.duration(); }
+
+Duration Individual::Route::earliestStart() const { return tws_.twEarly(); }
 
-Duration Individual::Route::waitDuration() const { return wait_; }
+Duration Individual::Route::latestStart() const { return tws_.twLate(); }
```
N-Wouda, line 245: Can we support something like this using a `slack()` attribute that's the difference between `twEarly` and `twLate`? Then `start()` can unambiguously be the earliest start time.

pyvrp/cpp/Individual.h
```diff
@@ -3,6 +3,7 @@
 
 #include "Measure.h"
 #include "ProblemData.h"
+#include "TimeWindowSegment.h"
```
N-Wouda, line 6: It's one thing that we use the concatenation scheme in the implementation, but I'd really like to avoid it in the declarations. In the internal state of `Individual::Route`, can we store only what we need, rather than the whole `TWS`? Then this include here can be avoided.

N-Wouda:
I'm really not sure about the increase in coupling. Now TWS needs to know about ProblemData(/Client), and Individual::Route about TWS. Would an implementation of start/end and slack computation on top of Individual::Route have been (much) harder?

pyvrp/cpp/TimeWindowSegment.h
```diff
@@ -26,20 +27,21 @@ class TimeWindowSegment
...
+    [[nodiscard]] inline Duration timeWarp() const;
+    [[nodiscard]] inline Duration twEarly() const;
+    [[nodiscard]] inline Duration twLate() const;
 
     TimeWindowSegment() = default;  // TODO at least require client index
 
-    inline TimeWindowSegment(int idxFirst,
-                             int idxLast,
+    inline TimeWindowSegment(size_t idxFirst_,
+                             size_t idxLast_,
```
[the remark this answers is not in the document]
wouterkool: Oops!

pyvrp/cpp/Individual.cpp
```diff
@@ -1,5 +1,6 @@
 #include "Individual.h"
 #include "ProblemData.h"
+#include "TimeWindowSegment.h"
```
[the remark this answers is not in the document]
wouterkool: I think this is not a bad thing. In my opinion, we want to avoid coupling primarily to avoid global state/access to objects which can lead to fragile code and unexpected behavior. In this case, I consider `TWS` a useful helper class/data type (it is in the root not in `search` for a good reason) grouping certain related attributes together and implementing time window computation logic ; I don't see the problem of using it? On the other hand I think it's not good to 'reimplement' the same logic in Individual (even if it's small ; we already see we messed up with the wait time computation, or at least not computing minimum waiting time). Can we keep it like this?

pyvrp/_Individual.pyi
```diff
@@ -47,6 +47,19 @@ class Route:
         """
         Total waiting duration on this route.
         """
+    def total_duration(self) -> int:
+        """
+        Total duration on this route (travel + service + wait - timewarp).
+        It holds that duration = end time - start time.
```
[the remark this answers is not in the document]
wouterkool: No it is not since it depends on the start time, so there is no unambiguous end time. BTW, I noted the comment is wrong (duration does not exclude timewarp).

pyvrp/_Individual.pyi
```diff
@@ -47,6 +47,19 @@ class Route:
         """
         Total waiting duration on this route.
         """
+    def total_duration(self) -> int:
```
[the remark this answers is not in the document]
wouterkool: I thought to make it more explicit as we have different type of durations but I'm fine with reverting it to `duration`.

pyvrp/cpp/Individual.cpp
wouterkool: I think it's useful to add an explicit `slack()` attribute, however I'm not sure if we should define a default start time. Any start time between `earliestStart()` and `latestStart()` is a valid start time that minimizes the route duration, and having an attribute `start()` equal to the earliest possible start time is a (somewhat) arbitrary default and makes it easy for a user to overlook the slack. Also, client time windows are modeled as `earliest` and `latest` starts, not as `start` and `slack`, so I I prefer to model start times similarly and keep it like this, forcing the user to pick an explicit start time from the interval. What do you think?

pyvrp/_Individual.pyi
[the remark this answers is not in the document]
wouterkool: We could make a function `end_time(Duration start_time)`?

pyvrp/cpp/Individual.h
wouterkool: With the same reasoning as before I don't really see the problem, I think it's nice we can group the time related attributes where we need all of them (duration, time warp, earliest and latest) except the indices in TWS (which I'd rather get rid of as well). This object just holds timing info for the complete route, which is what the TWS is intended for (+ some logic related to it). If you feel so strongly I can change it but I don't see the point of reimplementing mostly the same attributes rather than holding the struct? Or is there some technical aspect I'm missing?

wouterkool:
> I'm really not sure about the increase in coupling. Now TWS needs to know about ProblemData(/Client), and Individual::Route about TWS. Would an implementation of start/end and slack computation on top of Individual::Route have been (much) harder?

I can relate to your mixed feelings about `TWS` depending on Client, although I think it makes the code nicer (constructor just initializes the TWS for a client). As an alternative, since TWS never changes for a `Client`, how about I add it as a property to `Client` and we can just use that in LocalSearch etc. rather than the constructor?

For the slack computation, it would not have been *much* harder, but there are some formulas and I see no point of implementing the slack computation/propagation (which is not difficult but also not trivial) logic twice if we have a nice class for that?

pyvrp/cpp/Individual.cpp
N-Wouda: @leonlan what do you think?

leonlan: I think it's easier for users if we have a `start()` time and refer to `slack()` in the docstring, mentioning that the routes can start later.

pyvrp/cpp/educate/LocalSearch.cpp
```diff
@@ -272,13 +272,7 @@ void LocalSearch::loadIndividual(Individual const &individual)
...
     for (size_t client = 0; client <= data.numClients(); client++)
     {
-        clients[client].tw = {static_cast<int>(client),  // TODO cast
-                              static_cast<int>(client),  // TODO cast
-                              data.client(client).serviceDuration,
-                              0,
-                              data.client(client).twEarly,
-                              data.client(client).twLate};
-
+        clients[client].tw = {client, data.client(client)};
```
leonlan: Can we also rename `Node::tw` to `Node::tws` (and also for `Node::twBefore` and `Node::twAfter`?

pyvrp/cpp/Individual.cpp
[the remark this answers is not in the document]
leonlan: I don't have any particular strong opinion on this; what's the downside of this coupling?  TWS will always be used to compute route(segment) data, so why don't we want to use this?

Except when introducing time-dependent travel times, I am not aware of any commonly used route definitions that cannot be computed by the concatenation scheme in TWS.

pyvrp/cpp/educate/LocalSearch.cpp
N-Wouda: > Can we also rename `Node::tw` to `Node::tws` (and also for `Node::twBefore` and `Node::twAfter`?

Not opposed, but I suggest we leave this as-is. I'm planning to refactor `Node` and `Route` quite a bit in a few weeks, and this will probably change as part of that.

pyvrp/cpp/Individual.cpp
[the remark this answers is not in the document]
wouterkool: Actually, one of the reasons I do really want `Solution` to use TWS is time-dependent travel times, since I can then swap out TWS and have consistent time computations in both `Solution` and `LocalSearch` rather than implementing that (much more complicated) logic twice. So @N-Wouda can we keep this?

pyvrp/_Individual.pyi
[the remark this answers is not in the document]
wouterkool: @N-Wouda any thoughts?

pyvrp/cpp/Individual.cpp
wouterkool: Ok, just to confirm you want `start()` and `slack()` but not `earliest`/`latest`, but keep client time windows as `twEarly` + `twLate` (rather than e.g. `twStart` + `twSlack`)? That means we implicitly pick/suggest the earliest start time as a default if a user ignores slack?

pyvrp/cpp/Individual.h
wouterkool: @N-Wouda is this a problem or can we keep it?

pyvrp/cpp/Individual.cpp
N-Wouda: Yes.

N-Wouda (approved):
LGTM

N-Wouda:
I've added a note about how `slack()` works to the documentation:
image

N-Wouda:
@wouterkool I cannot ask you for review because it's your PR. Do you want to have a look at this again?

pyvrp/cpp/Solution.cpp
```diff
@@ -287,7 +279,18 @@ Duration Solution::Route::serviceDuration() const { return service_; }
...
+Duration Solution::Route::waitDuration() const
+{
+    return duration_ - travel_ - service_;
+}
+
+Duration Solution::Route::travelDuration() const { return travel_; }
+
+Duration Solution::Route::startTime() const { return startTime_; }
+
+Duration Solution::Route::endTime() const { return startTime_ + duration_; }
```
wouterkool, line 291: I think this is only correct if `timeWarp = 0`, it should be `endTime = startTime + duration - timeWarp` if we want this to be consistent with the definition of `endTime` in `TimeWindowSegment` (defined implicitly in `merge` as the start time for the travel).

See this code in `TimeWindowSegment::merge`:
```
auto const travelDuration = durationMatrix(idxLast, other.idxFirst);
auto const diff = duration - timeWarp + travelDuration;
auto const diffWait = std::max<Duration>(other.twEarly - diff - twLate, 0);
auto const diffTw = std::max<Duration>(twEarly + diff - other.twLate, 0);
```
The arrival time at the first client of the other segment is computed as `twEarly + diff`, which is `twEarly + duration - timeWarp + travelDuration` suggesting that `twEarly + duration - timeWarp` is the endTime corresponding to the starttime `twEarly`

We could consider a test for this (I can propose one but it may take a few days...).

pyvrp/cpp/Solution.h
```diff
@@ -95,27 +98,61 @@ class Solution
...
+         * the route can leave the depot and have a minimal duration. It may be
+         * possible to leave earlier, but that would incur additional waiting
+         * time.
+         *
+         * .. note::
+         *
+         *    It may also be possible to leave later, without (additional)
+         *    violations of time window constraints or increasing the route
+         *    duration. In that case there is slack in the schedule: the amount
+         *    of slack can be obtained through :meth:`~slack`.
```
wouterkool, line 136: I'm fine with this but I think it makes more sense if the description mentions the standard/preferred situation (leaving between earliest and latest start) and the note mentions the exceptions (leaving before earliest and after latest). Something along the lines of (did not check formatting).

```suggestion
         * Start time of this route. This is the earliest possible time at which
         * the route can leave the depot and have a minimal duration and time warp.
         * If :meth:`~slack` > 0, the start time can be delayed by at most the slack
         * without increasing the total (minimal) duration or time warp.
         *
         * .. note::
         *
         *    It may be possible to leave before the start time (if the depot time window
         *    allows for it) but this will introduce additional waiting time, increasing the
         *    total duration which will then no longer be minimal.
         *    Leaving after start time + slack will always increase the time warp, which, if
         *    it the time warp is currently 0, will cause the route to become infeasible w.r.t.
         *    time windows.
```

pyvrp/cpp/Solution.h
```diff
@@ -95,27 +98,61 @@ class Solution
...
+
+        /**
+         * End time of the route. This is equivalent to
+         * ``start_time + duration``.
+         */
+        [[nodiscard]] Duration endTime() const;
+
+        /**
+         * Time by which departure from the depot can be delayed without
+         * resulting in (additional) time window violations or increased route
```
wouterkool, line 148: We interchangably use 'time window violations' (which sounds as 1 or 2 or 3 violations) and 'time warp' which is a related but slightly different concept. I don't really mind but wanted to raise the question if this does need explanation/reformulation?

pyvrp/cpp/TimeWindowSegment.h
```diff
@@ -87,18 +115,25 @@ TimeWindowSegment TimeWindowSegment::merge(
...
-            other.idxLast,
-            duration + other.duration + arcDuration + diffWait,
-            timeWarp + other.timeWarp + diffTw,
-            std::max(other.twEarly - diff, twEarly) - diffWait,
-            std::min(other.twLate - diff, twLate) + diffTw,
-            std::max(releaseTime, other.releaseTime)};
+    using Dur = pyvrp::Duration;
+
+    // arcDuration is the travel duration from our last to other's first client,
+    // and atOther the time at which we arrive at the other's first client.
```
wouterkool, line 121: Note: `atOther` is relative to the start time of the route so the comment is not technically correct and therefore I'm also not sure if `atOther` is the best name... `diff` was more true to the relative aspect but also vague... maybe `firstToOtherFirstDuration`?
```suggestion
    // and atOther the total duration between starting at our first client to arriving at the other's first client.
```

pyvrp/tests/test_Solution.py
```diff
@@ -286,41 +286,126 @@ def test_route_access_methods():
...
+    sol = Solution(data, [[1, 3], [2, 4]])
+    routes = sol.get_routes()
+
+    # The first route has timewarp, so there is no slack in the schedule. We
+    # should thus depart as soon as possible to arrive at the first client the
+    # moment its time window opens.
+    assert_(routes[0].has_time_warp())
+    assert_allclose(routes[0].slack(), 0)
+    start_time = data.client(1).tw_early - data.duration(0, 1)
+    assert_allclose(routes[0].start_time(), start_time)
```
wouterkool, line 366: Add test for `end_time = start_time + duration - time_warp` here?

wouterkool (commented):
Thanks for picking this up, it looks neat overall. I have a few comments to be adressed, specifically `end_time` computation which I think should take into account `time_warp` (at least if we want to be consistent with `TimeWindowSegment`). Others are just questions/suggestions but feel free to ignore.

pyvrp/cpp/TimeWindowSegment.h
```diff
@@ -104,18 +107,22 @@ TimeWindowSegment TimeWindowSegment::merge(
...
     return {};
 #else
-    auto const travelDuration = durationMatrix(idxLast_, other.idxFirst_);
-    auto const diff = duration_ - timeWarp_ + travelDuration;
-    auto const diffWait
-        = std::max<Duration>(other.twEarly_ - diff - twLate_, 0);
-    auto const diffTw = std::max<Duration>(twEarly_ + diff - other.twLate_, 0);
+    // Arc travel from our last to other's first, and the time at which we
+    // arrive at the other's first client.
+    auto const arcDuration = durationMatrix(idxLast_, other.idxFirst_);
```
wouterkool: We don't use the concept of `arc` anywhere in PyVRP but we use `travelDuration` to refer to the duration of travelling between clients... so I think this should be `travelDuration`, why did you change it back?

N-Wouda: We use arc here and there in the documentation, so it's not completely alien. We use edge more often, so I've renamed it to that. We use `travelDuration` only as an aggregate, not for single edges. In this context it could just as well mean "travel duration of all clients in the TWS". I find that a bit confusing.

pyvrp/cpp/Solution.h
N-Wouda: I've switched the two cases we have over to time warp.

pyvrp/cpp/TimeWindowSegment.h
N-Wouda: I've updated the comment to reflect this.

pyvrp/cpp/Solution.cpp
N-Wouda: I've added this in, along with a test!

N-Wouda:
Just checked the docs, and those look OK as well!

PR #243, opened 2023-06-16
Remove distances and prizes from Individual string representation

leonlan:
The motivation for this PR is to make the benchmarking process easier. 

The current prize and distance fields cannot be effectively used to benchmark PC-VRP instances, because we cannot use them to compute the "minimize distance + uncollected prizes" objective. So to get the cost for these solutions, we need to load these individuals and use a cost evaluator object to evaluate them. 

It would be much more convenient to have a single "Cost" field in the solution output. But this is not accessible from within Individual. What this PR proposes is to:
- Remove the distances and prizes fields from the individual string representation. 
- Add a cost field when writing the solution to a file in `cli.py` by using the cost evaluator.


pyvrp/cli.py
```diff
@@ -175,6 +175,7 @@ def solve(
 
         with open(where, "w") as fh:
             fh.write(str(result.best))
+            fh.write(f"Cost: {result.cost()}" + "\n")
```
N-Wouda, line 178:
```suggestion
            fh.write(f"Cost: {result.cost()}\n")
```

pyvrp/tests/test_Individual.py
```diff
@@ -322,10 +322,6 @@ def test_str_contains_essential_information():
             for client in route:
                 assert_(str(client) in str_route)
 
-        # Last lines should contain the travel distance and collected prizes.
```
N-Wouda, line 325: I'm pretty sure the asserts above in this test now fail (because + 2 is no longer needed). Can we update the comments/asserts there as well?

N-Wouda: Maybe also rename the test to `test_str_contains_routes` or similar?

leonlan: Sorry, that was sloppy 😅

N-Wouda (approved):

PR #248, opened 2023-06-17
Rename pyvrp.educate to pyvrp.search

### LINKED ISSUE #239: Rename some stuff to make it less dependent on HGS terminology
Particularly
- [x] `Individual` to `Solution`
- [x] `pyvrp.educate` to `pyvrp.improve`? Or `pyvrp.search`?

For the latter I think I prefer `pyvrp.search`: `LocalSearch` fits the bill already, and a future `LargeNeighbourhoodSearch` or `VariableNeighbourhoodSearch` will also work.

N-Wouda:
This PR:

- [x] Closes #239.

<details>

**Notes**:

</details>


N-Wouda:
@leonlan could you also review this final change for #239? Again fairly mechanical, `pyvrp.educate` -> `pyvrp.search`.

pyvrp/GeneticAlgorithm.py
leonlan, line 1: Should we also rename `GeneticAlgorithm._educate` to `GeneticAlgorithm._search`?

leonlan: With #46 we will probably pass a `SearchMethod` to the GeneticAlgorithm. Since "educate" is nothing more than applying a search method, we can get rid of educate entirely. And "Search" is in HGS, so I think that's also another reason to change it :-)

leonlan (approved):

pyvrp/GeneticAlgorithm.py
N-Wouda: Sure!

Issue #122, opened 2023-02-24
LocalSearch API


wouterkool:
We should think about how to use the `LocalSearch` object. There are several things for which we should consider if they should:
* be set via constructor (i.e. assuming it will be constant after initializing)
* set via property (i.e. can be changed from time to time)
* or taken as argument in `run`/`search`/`intensify` (assumed to change for every run)

We should consider:
* problem data (constructor / fixed)
* operators (constructor / fixed?)
* neighbourhoods (constructor and/or property, changing occasionally)
* penalty manager (penalties/objective object that changes frequently so argument to `run`)
* individual (argument to `run`, will change frequently)
* parameters (mix, some are fixed, some can change e.g. intensification params)

N-Wouda:
Oh, I think I can put the comment from #121 here:

>Instead of a degree in [0, 360] for `overlapToleranceDegrees` in `LocalSearch::intensify`, I think it makes much more sense to use a percentage [0, 1.0]. We can translate that back to a degree/CircleSector integer internally, but then users don't have to understand that there's a circle sector thing around at all. It's just "how much should routes overlap before we consider them in the intensification routine". How we determine that overlap exactly is more of a detail; to a user, we just present a percentage.

N-Wouda:
Is there anything here that we should still like to change? If not, I propose we close this issue.

Issue #126, opened 2023-02-25
Route neighbours


leonlan:
#121 reimplemented the circle sector overlap for identifying two neighbouring routes. Although quite effective in terms of performance, we should generalize this to a `RouteNeighbours` object that checks if two routes (U, V) are pairs.

N-Wouda:
A problem with the circle sector things is that they do far too much (i.e., too many overlapping routes) when there aren't a lot of clustered customers. This is partially fixed by the centroid angle thing we have now (since #251), but that isn't all that great either. We probably need something else.

N-Wouda:
With, of course, the massive caveat that I'm not at all convinced we even need intensification in its current form at all.

wouterkool:
Summary of considerations/ideas:
* Only based on centroid + tolerance (as implemented in #251) seems too limited to capture variety of routes.
* Doing a circle sector that captures all clients is probably too much
* Given that we have a centroid, we could compute a measure of (angular) spread of the route (e.g. angle variance relative to centroid) and set a threshold for overlap (e.g. 1 or 2 stddevs)
* We could sort routes by angle and do k-nearest, making the threshold 'dynamic'
* We could define route neighbours implicitly through the client granular neighbourhood. For a pair of routes U,V, count how many client pairs u,v are granular neighbours. If this exceeds a threshold number and/or percentage, consider the routes as neighbours. This would also work in non-euclidean case.

N-Wouda:
> * I think there's some ideas in https://optimization-online.org/2021/03/8316/ that are relevant.

I briefly skimmed this paper, and yeah, that does seem relevant!

N-Wouda:
I toyed with this a bit today, but I don't have a clear answer yet. I have learned some relevant things:

>We could define route neighbours implicitly through the client granular neighbourhood. For a pair of routes U,V, count how many client pairs u,v are granular neighbours. If this exceeds a threshold number and/or percentage, consider the routes as neighbours. This would also work in non-euclidean case.

I tried something similar to this (essentially compute the client pair count and then take the top k routes as neighbours, rather than a treshold). This is hard to make fast and does not seem to be better than what we have now.

Similar to this idea, I have tried to simply track the routes that were involved in an improving move, and add them to each other's neighbourhoods (with a max size to avoid neighbourhoods growing really large; removing the oldest insert to make space for the new one). That worked surprisingly well, but gets stuck in a local optimum quickly. It certainly isn't better than what we currently do based on the centroid.

[...]

PR #251, opened 2023-06-21
Remove circle sector

N-Wouda:
This PR removes the CircleSector stuff in favour of something a little simpler. It's a bit of an experiment, so I'm not sure if this will ever leave the draft phase.

- [x] Relates somewhat to #122 with the overlap tolerance parameter.
- [x] Adds and passes relevant tests. (some, not a lot)
- [x] Profiled.

### Benchmark

On VRPTW, avg. over ten seeds with 1h runtime:

| category     | Main| This PR / tol `0.05` | This PR / tol `0.1` |
|-----|-------|---------------|----------------|
| C1  | 413653.2 | 413715.9 | **413624.8** |
| C2  | 161914.0 | 161911.7 | **161904.6** |
| R1  | 473236.7 | **472958.3** | 473476.1 |
| R2  | 277276.7 | **277151.2** | 277180.0 |
| RC1 | **442251.7** | 442278.5 | 442496.8 |
| RC2 | 232344.3 | **232172.1** | 232267.2 |

This PR with tol `0.1` (dda1c3754553ad6d08c2f541f2734d217f9d7034; job ID 3191043; 10x seeds/1h):
```
 Avg. objective: 333491.7 (min: 333335, max: 333615)
Avg. iterations: 46777.7 (min: 37213, max: 54719)
```

This PR with tol `0.05` (9a367b12c83863c24a0e623aafe1b2bb8ec3290d; job ID 3197392; 10x seeds/1h):
```
 Avg. objective: 333364.7 (min: 333266, max: 333541)
Avg. iterations: 65629.2 (min: 54852, max: 73735)
```

Current main (44814444bdc641a6bd2b97db4f776de9ef94b934; job ID 3194371; 10x seeds/1h):
```
 Avg. objective: 333446.2 (min: 333378, max: 333578)
Avg. iterations: 53109.8 (min: 42391, max: 60488)
```

<details>

Current main:
```
                mean       min       max
      inst                              
obj   C1    413653.2  413399.5  413894.1
      C2    161914.0  161875.7  161940.3
      R1    473236.7  472823.4  473928.7
      R2    277276.7  277137.7  277423.1
      RC1   442251.7  441941.2  442696.9
      RC2   232344.3  232142.0  232515.8
iters C1     86897.7   67948.0   96072.3
      C2     89695.5   70547.7  104632.9
      R1     32816.6   24127.4   39345.9
      R2     31566.0   25629.5   37211.7
      RC1    41553.3   33352.8   49832.1
      RC2    36129.7   30638.5   40896.0
```

Tol 5%:
```
                mean       min       max
      inst                              
obj   C1    413715.9  413506.7  413963.9
      C2    161911.7  161881.9  161940.2
      R1    472958.3  472276.9  473682.5
      R2    277151.2  276883.0  277421.0
      RC1   442278.5  441943.8  442802.4
      RC2   232172.1  232039.6  232297.8
iters C1     68754.8   58413.7   75266.2
      C2    125908.9  104863.5  137205.7
      R1     37943.0   29827.4   46800.1
      R2     55186.0   44400.6   62332.7
      RC1    43416.2   37265.2   48655.1
      RC2    62565.9   52367.9   73215.2
```

Tol 10%:
```
                mean       min       max
      inst                              
obj   C1    413624.8  413419.1  413787.6
      C2    161904.6  161876.3  161952.9
      R1    473476.1  472934.1  474472.5
      R2    277180.0  276856.6  277492.0
      RC1   442496.8  442036.7  442847.5
      RC2   232267.2  232094.3  232446.5
iters C1     49744.7   38400.1   58399.0
      C2     82816.3   65034.2  101933.3
      R1     27435.0   21179.1   33515.0
      R2     40568.2   33302.2   45933.8
      RC1    32510.7   25672.3   40133.3
      RC2    47591.7   35629.1   56144.2
```

</details>


N-Wouda:
Profiling this doesn't look too bad. `std::atan` isn't fast, but it doesn't take more than 0.5% of the time, so I don't think it's worth swapping that out for something faster right now.

SWAP* is much more concerning. Something to look into in the future.

N-Wouda:
Took a little calibrating, but an `overlap_tolerance` of 5% seems to do the trick.

N-Wouda:
@leonlan @wouterkool could either/both of you have a look? This PR does the overlap thing for routes in a simpler way than what we currently have with the `CircleSector`, and that seems to be OK in terms of performance (benchmarks in first comment above).

leonlan:
It's been a while since I touched the CircleSector code, but IIRC, the circle sectors are "pie-slices" that cover the entire route right? And the current PR proposes to simplify this by only looking at the route centroids? Are there any cases where this may go wrong really badly?

leonlan (approved):
This PR removes the CircleSector, which was a simple idea to select two routes for intensification. But the implementation was somewhat complicated. 

The new route centroid idea to select two routes is somewhat less clear as an idea (it's harder to visualize for me). But we already work a lot with route centroids (greedy repair, sorting routes), so the route centroids are not that hard to understand. We do need to introduce a single parameter (overlap tolerance) required to make it work. I think this new method is less robust, but I prefer the simpler implementation.

Overall I agree to remove the circle sector stuff for this simpler implementation.

N-Wouda:
>We do need to introduce a single parameter (overlap tolerance) required to make it work. I think this new method is less robust, but I prefer the simpler implementation.

(we already have this parameter currently as well, it's not new)

I can see us remove this particular thing pretty quickly anyway, particularly if we do anything about #126. It's more of a temporary solution that at least removes some stuff that's pretty complicated and apparently not really needed.

N-Wouda:
> It's been a while since I touched the CircleSector code, but IIRC, the circle sectors are "pie slices" that cover the entire route right? And the current PR proposes to simplify this by only looking at the route centroids? Are there any cases where this may go wrong really badly?

It could result in very few routes overlapping, which might be bad. But I can't think of a nontrivial example where that'd happen, and in any case, we also still have search and the granular neighbourhood that covers some of that. In other cases, ~1/10th of the circle (+5% or -5% overlap) is enough for the centroids to be "close enough". I think that works particularly well on the random instances (R1/R2) where previously the sectors would overlap almost all the time, but the centroids could be pretty far apart.

leonlan:
> (we already have this parameter currently as well, it's not new)

That's true, but it was unused. I don't know when we added this parameter, but HGS-CVRP did not have this parameter.

> I can see us remove this particular thing pretty quickly anyway, particularly if we do anything about https://github.com/PyVRP/PyVRP/issues/126. It's more of a temporary solution that at least removes some stuff that's pretty complicated and apparently not really needed.

If we follow up with #126 then I think we should definitely do this. (#126 is very low priority though because I don't have any "route neighbors" ideas right now.)

N-Wouda:
>I don't know when we added this parameter, but HGS-CVRP did not have this parameter.

From what I can gather it's a leftover from the [EURO/NeurIPS baseline](https://github.com/ortec/euro-neurips-vrp-2022-quickstart/blob/77a1a24e6b30baf3aab8931c020302c3975ca77e/baselines/hgs_vrptw/Params.h#L111). We removed it in our implementation because in the end we didn't do this intensification regularly, and then reintroduced it during our performance panic in February :-).

PR #254, opened 2023-06-27
Release times

### LINKED ISSUE #244: Bring back support for release times
We removed release times in https://github.com/PyVRP/PyVRP/commit/3065a577076d92326f3f03aa701976bd8a2bd7e0. Since release times are quite common in VRPs, it's probably good to add this back again. I also would like to use the PyVRP solver for my research on dynamic vehicle routing.

----
- [ ] Include release times crossover::greedyRepair? https://github.com/PyVRP/PyVRP/blob/3065a577076d92326f3f03aa701976bd8a2bd7e0/pyvrp/cpp/crossover/crossover.cpp

leonlan:
This PR:

- [x] Closes #244.

<details>

**Notes**:

</details>


pyvrp/read.py
```diff
@@ -146,12 +151,15 @@ def read(
     if demands[0] != 0:
         raise ValueError("Demand of depot must be 0")
 
-    if time_windows[0, 0] != 0:
```
leonlan, line 149: I'm not sure why we require the depot time window to start at 0 in `read.py`. Idem for the other data. In Solution, we track the time starting at depot tw_early:

```
Duration time = data.depot().twEarly;
```

pyvrp/cpp/TimeWindowSegment.h
```diff
@@ -8,12 +8,13 @@ class TimeWindowSegment
...
-    Duration duration = 0;  // Total duration, incl. waiting and servicing
-    Duration timeWarp = 0;  // Cumulative time warp
-    Duration twEarly = 0;   // Earliest visit moment of first client
-    Duration twLate = 0;    // Latest visit moment of last client
+    int idxFirst = 0;          // Index of the first client in the segment
+    int idxLast = 0;           // Index of the last client in the segment
+    Duration duration = 0;     // Total duration, incl. waiting and servicing
+    Duration timeWarp = 0;     // Cumulative time warp
+    Duration twEarly = 0;      // Earliest visit moment of first client
+    Duration twLate = 0;       // Latest visit moment of first client
```
leonlan, line 16: Changed "last" to "first" here (per [Cattaruzza et al. (2016)](https://pubsonline.informs.org/doi/abs/10.1287/trsc.2015.0608)).

leonlan:
I'm going to generate some benchmark instances for VRPTW-RT using the EURO-NeurIPS instances.

pyvrp/Model.py
```diff
@@ -128,7 +130,12 @@ def add_client(
         return client
 
     def add_depot(
-        self, x: int, y: int, tw_early: int = 0, tw_late: int = 0
+        self,
+        x: int,
+        y: int,
+        tw_early: int = 0,
+        tw_late: int = 0,
+        release_time: int = 0,
```
leonlan, line 138: `tw_early` and `release_time` essentially do the same thing. Not sure if we should include release time here.

pyvrp/_Solution.pyi
```diff
@@ -49,6 +49,10 @@ class Route:
         """
         Total waiting duration on this route.
         """
+    def release_time(self) -> int:
+        """
+        Release time (that is, earliest time to leave depot) of this route.
```
leonlan, line 54:
```suggestion
        Release time (that is, the earliest time allowed to leave the depot) of this route.
```

leonlan:
I'm going a benchmark now on 100 VRPTW-RT instances from the Euro-NeurIPS competition.

pyvrp/Model.py
N-Wouda: I don't know what a release time for the depot would mean. Let's just use tw early here and assume the depot release time is zero.

pyvrp/cpp/Solution.cpp
```diff
@@ -203,7 +203,11 @@ Solution::Route::Route(ProblemData const &data,
     if (visits_.empty())
         return;
 
-    Duration time = data.depot().twEarly;
+    release_ = std::max(release_, data.depot().releaseTime);
```
N-Wouda, line 206:
```suggestion
```

pyvrp/cpp/TimeWindowSegment.h
```diff
@@ -82,20 +85,26 @@ TimeWindowSegment TimeWindowSegment::merge(
 #endif
 }
 
-Duration TimeWindowSegment::totalTimeWarp() const { return timeWarp; }
+Duration TimeWindowSegment::totalTimeWarp() const
+{
+    auto const startDiff = releaseTime - twLate;
+    return timeWarp + std::max(startDiff, static_cast<Duration>(0));
```
N-Wouda, line 91:
```suggestion
    return timeWarp + std::max<Duration>(releaseTime - twLate, 0);
```

pyvrp/cpp/TimeWindowSegment_bindings.cpp
```diff
@@ -22,13 +22,14 @@ template <typename... Args> TWS merge(Matrix<int> const &mat, Args... args)
 PYBIND11_MODULE(_TimeWindowSegment, m)
 {
     py::class_<TWS>(m, "TimeWindowSegment")
-        .def(py::init<int, int, int, int, int, int>(),
+        .def(py::init<int, int, int, int, int, int, int>(),
```
N-Wouda, line 25: Except for the first two (indices), these should probably all be `Value`.

pyvrp/_TimeWindowSegment.pyi
```diff
@@ -29,6 +30,8 @@ class TimeWindowSegment:
             Earliest visit moment of the first client.
         tw_late
             Latest visit moment of the last client.
```
N-Wouda, line 32:
```suggestion
        tw_late
            Latest visit moment of the first client.
```

pyvrp/tests/test_Solution.py
```diff
@@ -176,6 +176,18 @@ def test_feasibility():
     assert_(not sol.has_time_warp())
 
 
+def test_feasibility_release_times():
+    data = read("data/OkSmallReleaseTimes.txt")
+
+    # Client 1 is released at 20000, but client 2 time window ends at 19500.
+    sol = Solution(data, [[1, 2], [3], [4]])
+    assert_(not sol.is_feasible())
```
N-Wouda, line 184: Can we test that that incurs exactly as much time warp as we would expect from leaving depot at 20K, driving to 1, doing service, and driving to 2 compared to client 2's `tw_late`?

pyvrp/tests/test_Solution.py
```diff
@@ -290,6 +302,15 @@ def test_route_time_warp_and_wait_duration():
     assert_equal(routes[1].wait_duration(), 12_000 - 1_944)
 
 
+def test_route_release_time():
+    data = read("data/OkSmallReleaseTimes.txt")
+    sol = Solution(data, [[1, 3], [2, 4]])
+    routes = sol.get_routes()
+
+    assert_allclose(routes[0].release_time(), 20000)
+    assert_allclose(routes[1].release_time(), 5000)
```
N-Wouda, line 311: Can we make this into more of a story? Which clients have which release times?

N-Wouda (commented):
Looks good to me. A few small comments, but nothing substantial.

leonlan:
@N-Wouda I've implemented your comments. Performance is also OK (see https://github.com/PyVRP/PyVRP/pull/254#issuecomment-1617650508).

N-Wouda (approved):
LGTM!

pyvrp/read.py
N-Wouda: I'm a bit late to this party, but yes, I think so. We should move what we really need to check into ProblemData's constructor. We might also want to lift some restrictions at that point.

leonlan: I have some initial work for this on the branch https://github.com/PyVRP/PyVRP/tree/non-zero-depot-early-tw. I'll finish it this weekend.

PR #255, opened 2023-07-02
Use Meson subprojects for pybind dependency

N-Wouda:
This PR switches over to Meson's subprojects feature for getting the pybind11 dependency, rather than bringing it in via Poetry/PyPI. This is the standard Meson way for managing C++ dependencies, and doing it that way will be useful for when we want to bring in other C++ dependencies as well.


N-Wouda:
I'm merging this in because it's build related, and I don't think standardising on Meson dependencies is particularly controversial.

meson.build
```diff
@@ -63,24 +63,9 @@ libcommon = static_library(
...
-    # Fallback: this looks for the pybind11 header files using pybind11-config.
-    # Getting pybind11 using Meson's dependency mechanism is nicer, but that
-    # sometimes fails even when pybind11 is actually available on the system.
-    message('Could not find pybind11 as a dependency: trying headers.')
-
-    pybind11 = find_program('pybind11-config')
-    pybind11 = run_command(pybind11, ['--includes'], check: true)
-    pybind11 = pybind11.stdout().split('-I')[-1].strip()
-    pybind11 = declare_dependency(include_directories: [pybind11])
-endif
```
eli-schwartz, line 81: The really nice thing about this is https://mesonbuild.com/Release-notes-for-1-1-0.html#new-pybind11-custom-dependency

Current versions of meson will detect pybind11 using the pybind11-config script installed as part of `pip install pybind11`. For older versions of python, it's fine to just fail and use the subproject fallback.

N-Wouda: Very neat! I'm a big fan of Meson so far; it's much more intuitive (to me, at least) than CMake ever was.

eli-schwartz: I'm very happy to hear that you're enjoying it. :)

PR #260, opened 2023-07-05
Use Ruff instead of flake8 and isort

N-Wouda:
Replaces flake8 and isort with Ruff in the pre-commit configuration. Ruff is almost a drop-in replacement, but there are a few additional checks that we have not previously enforced (and thus result in various changes in this PR).


N-Wouda:
Blocked until https://github.com/astral-sh/ruff/issues/5514 is in Ruff.

pyproject.toml
```diff
@@ -93,10 +93,14 @@ pyvrp = "pyvrp.cli:main"
...
 
 
-[tool.isort]
-case_sensitive = true
-line_length = 79
-profile = "black"
+[tool.ruff]
+ignore-init-module-imports = true
+line-length = 79
+select = ["E", "F", "I", "NPY", "RET", "RSE", "RUF", "SLF", "SIM", "TCH"]
```
N-Wouda, line 99: We might expand this list later. For now I've taken the defaults (E and F) and added a bunch of additional ones that seemed useful.

pyproject.toml
```diff
@@ -93,10 +93,15 @@ pyvrp = "pyvrp.cli:main"
...
+
+
+[tool.ruff.isort]
+# case-sensitive = true
```
N-Wouda, line 103: https://github.com/astral-sh/ruff/issues/5514 is fixed; I'm waiting until it's released.

PR #261, opened 2023-07-07
Fix rare segfault in greedyRepair

N-Wouda:
We take the first route as an initial default, and then check if there are better routes in the set. If no such routes exist, the first route remains the "best route". If that route is empty, however, we do index it when determining the insertion offset. That leads to a fairly rare segfault.


N-Wouda:
I'll release `v0.4.2` as a hotfix directly after merging this, and then yank `0.4.1` from PyPI.

PR #265, opened 2023-07-08
Remove using namespace directives

N-Wouda:
Follow-up to #264: this PR removes all `using namespace` directives.


N-Wouda:
@leonlan this was way less work than I thought it would be. Let's see if clang also accepts this :)

Issue #259, opened 2023-07-04
Example notebooks


leonlan:
With all the features that we implemented over the past month, I think it's a good moment to re-think about what kind of example notebooks we want.

Ideas:
- [x] **Model interface**: One notebook demonstrating the use of the `Model` interface, and how it can be used to solve different variants of VRPs (CVRP, VRPTW, PC-VRP, HF-VRP, VRP-RT, etc.). The target audience is users interested in solving VRPs but with little experience in building algorithms.
- [x] **Custom genetic algorithms**: showcasing how to customize a genetic algorithm to solve a single VRP variant (e.g., VRPTW). The target audience for this notebook is users interested in customizing the genetic algorithm for research or performance purposes. Examples are using different crossovers, and other (local) search strategies.
- [ ] **Other (meta-)heuristics**: showing users how to use PyVRP components to build their own VRP algorithm, like Iterated Local Search.

Questions:
- We currently have two example notebooks: one for [VRPTW](https://pyvrp.readthedocs.io/en/latest/examples/vrptw.html) and one for [CVRP](https://pyvrp.readthedocs.io/en/latest/examples/cvrp.html). What do we want to do with these notebooks? I suggest re-using some parts of these notebooks for the new suggestions above. I find these notebooks on themselves are not very useful for users.

Other ideas/suggestions?

N-Wouda:
First one yes, definitely, as a really simple introduction. We can have a look at how e.g. OR-Tools does this. Their intro is pretty good. This needs to be really accessible, much more so than the current notebooks.

Second: yes, as a more in-depth notebook to complement the high-level one using the modelling interface. Can re-use quite a bit from the existing notebooks, but should have much more descriptive text explaining how things work.

Third: sure? I don't mind having such a notebook, but I am not going to write it myself.

N-Wouda:
I'll have a go at the first two soon.

N-Wouda:
I'm unassigning myself because the two items I wanted to do for this issue have been completed in #266.

leonlan:
I don't think it's necessary to have the third item, because it's already clear from https://pyvrp.org/examples/using_pyvrp_components.html how to use PyVRP components to build a custom algorithm. So I consider this issue to be done.

PR #266, opened 2023-07-09
Update notebooks

N-Wouda:
This PR:

- [x] Is a part of #259.

<details>

**Notes**:

</details>


N-Wouda:
TODO:
- [x] Add short tutorial notebooks
- [ ] Simplify and rewrite CVRP/VRPTW notebooks
- [ ] Add notebook for prize collecting/multiple vehicle capacities?

N-Wouda:
@leonlan I think this is done. Can you have a look? I've added something akin to a quickstart, a notebook where we solve benchmark instances for CVRP and VRPTW, and a notebook where we implement the `solve` routine from scratch and explain along the way how all the objects needed for it work.

.github/workflows/CI.yml
```diff
@@ -114,10 +114,3 @@ jobs:
         run: |
           VALGRIND_CMD="pyvrp pyvrp/tests/data/RC208.txt --instance_format solomon --round_func dimacs --max_runtime 0.5 --seed 1"
           poetry run env PYTHONMALLOC=malloc valgrind --tool=memcheck --quiet $VALGRIND_CMD
-      - if: matrix.compiler == 'clang' && matrix.python-version == '3.11'
-        # Run this only for one job to reduce workflow runtimes.
-        name: Run example notebooks
-        run: |
-          for f in examples/*.ipynb;
-            do poetry run jupyter nbconvert --to notebook --execute $f;
-          done
```
N-Wouda, line 123: RTD now executes each notebook before rendering the docs, so this is not needed anymore.

examples/using_pyvrp_components.ipynb
leonlan, line 1: Then in "A tour of PyVRP":

We should reference the [Intro to HGS](https://pyvrp--266.org.readthedocs.build/en/266/setup/introduction_to_hgs.html) page and assume that the readers have read that. We can then get rid of the following paragraph:

> The GeneticAlgorithm object manages a population of solutions. In each iteration, two solutions are selected from this population for crossover using a crossover operator from pyvrp.crossover, which generates a new offspring solution. That offspring solution is the improved (educated) using a method from pyvrp.search. The improved offspring solution is then added to the population. This process continues until a stopping condition is reached (see pyvrp.stop for different conditions). Let’s have a look at the different parts of pyvrp that implement this algorithm.

Then instead of describing what the genetic algorithm does, we should describe what (code) components are needed to build a genetic algorithm. We can start by presenting the signature of GeneticAlgorithm here.

The outline of this section is currently:
- Random number generator
- Local search, cost evaluator and penalty manager
- Population, diversity and initial solutions
- Crossover
- Stopping criterion

I think this outline should be visible somewhere, either at the beginning of this section, or as subsection headers.

leonlan, line 1: Then in "the `solve` function:

- The GeneticAlgorithm params/statistics section is bit verbose for what it’s worth. Maybe introduce GeneticAlgorithmParams as final PyVRP component.

leonlan (commented):
In summary:
- `A quick tutorial`: some minor comments, but for the rest good to go!
- `Basic VRPs`: Goal of the notebook does not match the description yet. What direction to go into is up for discussion: I think it's useful to include another PC-VRPTW instance.
- `Using PyVRP component's`: Content is there, but structure is missing.

The review comments have become a mess because it's chronologically ordered, sorry for that!

examples/basic_vrps.ipynb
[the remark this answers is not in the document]
N-Wouda: My execution environment is really slow. RTD is faster. I'll move this over to iteration-based stuff to make sure, and mention somewhere that we also have the runtime one but that the results then can be quite variable.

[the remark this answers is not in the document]
N-Wouda: I removed this for now. We might add prize collecting or other variants at some point, but I don't feel like doing that now.

docs/source/index.rst
```diff
@@ -30,6 +33,7 @@ Contents
 
    setup/introduction_to_vrp
    setup/introduction_to_hgs
+   examples/getting_started
```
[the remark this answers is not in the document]
N-Wouda: The notebooks are automatically copied into the documentation from `examples/`. I would rather not make an exception (so it has to stay `examples/`).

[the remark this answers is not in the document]
N-Wouda: I'm OK with keeping `setup/`, I think. We also call it this way in ALNS where it covers both setting up the library, and getting started. Getting started to me is part of setting stuff up. The user doesn't really see this in any case.

N-Wouda:
@leonlan want to have another look?

leonlan:
> @leonlan want to have another look?

I'll have a look now

examples/quick_tutorial.ipynb
leonlan, line 1: -  In the `TIME_WINDOWS` variable:  `(0, 100),  # location 0 - the depot (not restricted)`, what does the "(not restricted)" refer to? Can we remove this comment?

examples/basic_vrps.ipynb
leonlan, line 1: - The list in VRPTW - Reading the instance section is not rendered correctly. Needs a new line.

examples/quick_tutorial.ipynb
N-Wouda: The OR Tools example restricts the depot departure time (I think? It's not obvious) to be between `(0, 5)`. See the code and data [here](https://developers.google.com/optimization/routing/vrptw). I've increased this value because I don't want that fairly arbitrary constraint in the tutorial. I'd rather keep this in here because else I'm going to forget this is modified.

leonlan (approved):
Looks good to me now! 

I think we can improve the design of the notebooks. The code blocks don't have a margin at the bottom. I'll make a separate issue for this.

PR #267, opened 2023-07-09
Benchmarks v0.4.2

leonlan:
This PR updates the benchmark for v0.4.2.

### DIFF
--- a/docs/source/dev/benchmarking.rst
+++ b/docs/source/dev/benchmarking.rst
@@ -26,7 +26,7 @@ The time limit should be scaled by the PassMark score of your CPU.
 Each instance is run ten times with different seeds.
 Each run is performed on a single core.
 For each instance, we take the average objectives of these ten runs with different seeds.
-These are compared against the most recent best-known solutions in the ``PyVRP/Instances`` repository.
+These are compared against the latest best-known solutions in the ``PyVRP/Instances`` repository at the time of the benchmark.
 
 To get the benchmark instance sets quickly, the ``PyVRP/Instances`` repository may be initialised as a submodule:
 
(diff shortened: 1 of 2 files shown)


leonlan:
Pretty significant drop in performance on CVRP instances (0.25% -> 0.37%). This also holds for V0.4.0.

N-Wouda:
This is w.r.t. to the BKS on 28/2/23 right? We should probably add a note of some sorts to the page to describe why we do that.

Should we mention that we changed computer clusters between 0.3.0 and 0.4.2?

leonlan:
> This is w.r.t. to the BKS on 28/2/23 right? We should probably add a note of some sorts to the page to describe why we do that.

It's w.r.t. the most recent BKS but they're the same as 28/2/23.

> Should we mention that we changed computer clusters between 0.3.0 and 0.4.2?

I actually realize that made a mistake due to migration. I rewrote some scripts and I forgot to use the "cvrp.toml" config for the CVRP benchmark. I'm going to run this benchmark again.

N-Wouda:
> I actually realize that made a mistake due to the migration. I rewrote some scripts and I forgot to use the "cvrp.toml" config for the CVRP benchmark. I'm going to run this benchmark again.

Oh, that could well be it!

N-Wouda:
> It's w.r.t. the most recent BKS but they're the same as 28/2/23.

I feel like we need to clarify this somehow/somewhere. If our performance remains the same but the BKS do change, it looks like we're getting worse w.r.t. previous versions of PyVRP - even though that need not be the case at all. When the BKS do change we might want to add a column that compares to the BKS as they were when we started benchmarking (28/2/23), and the BKS as they are at the time of release. Something to circle back to later when it's actually relevant.

leonlan:
It indeed turned out to be the configuration file: the gap with the `cvrp.toml` config file is 0.18%.

N-Wouda (approved):

N-Wouda:
LGTM. Let's wait for the docs build to see if the table now renders correctly, and then we can merge it in. 

Thanks!

PR #269, opened 2023-07-10
Use Solution string representation in Result string representation

leonlan:
This PR uses `Solution`s string representation to print the routes in the summary/string representation of `Result`.

### DIFF
--- a/pyvrp/Result.py
+++ b/pyvrp/Result.py
@@ -94,7 +94,7 @@ class Result:
             "",
             "Routes",
             "------",
+            str(self.best),
         ]
-        summary.extend(str(self.best).splitlines())
 
         return "\n".join(summary)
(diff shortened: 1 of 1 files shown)


leonlan:
I just noticed this while reviewing the new notebooks. Will become handy when we change the Solution string representation, for example when we include vehicle type information.

pyvrp/Result.py
```diff
@@ -95,9 +95,6 @@ def __str__(self) -> str:
             "Routes",
             "------",
         ]
-
-        for idx, route in enumerate(self.best.get_routes(), 1):
-            if route:
-                summary.append(f"Route {idx:>2}: {route}")
+        summary.extend(str(self.best).splitlines())
```
N-Wouda, line 98: Or,
```suggestion
            "Routes",
            "------",
            str(self.best),
        ]
```
?

leonlan: Yea this works!

N-Wouda (approved):
LGTM!

PR #271, opened 2023-07-11
Switch to sphinx-immaterial theme

### LINKED ISSUE #270: New documentation theme
The RTD theme has some issues, see details below about margins in example notebooks:
<details>
<summary><b>Margins in example notebooks</b></summary>
Unlike regular text blocks, code blocks have zero margins at the bottom. For a sequence of code blocks, this looks fine.

But when for notebooks where code and text are continuously alternating (for example PyVRP components), this results in the following:
image

This looks a bit too compressed to me. I'd want to see some margins between the code block and the subsection header. And also some margin between the code block and the next paragraph. For example:
image

Increasing the bottom margin of code blocks "fixes" this problem, but it makes a sequence of code blocks look ugly. 
I think the solution is to add some top margin to paragraphs and sections headers (for h3 and smaller, h1-h2 already have some marigns at the top).
</details>

This issue serves to discuss using a new documentation theme.

N-Wouda:
This PR:

- [x] Closes #270.

TODO:
- [x] Getting started section
- [x] Examples
- [x] API docs
- [x] Developing PyVRP

Known issues:
- The HGS pseudocode in intro to HGS does not render correctly. Not sure what we can do about that.
- [also in current theme] API type hints sometimes render as `~typing.List`, rather than `List`.

<details>

**Notes**:

</details>


N-Wouda:
WIP because we need to make a few adjustments here and there to really make this work. But it already looks really good!

N-Wouda:
API docs look OK. They're not perfect, but it'll do for now.

N-Wouda:
@leonlan could you have a look? I think this is OK now for a first version. There are some warts (see known issues above) that I'll copy over to a new issue after.

leonlan (approved):
Looks good!

PR #274, opened 2023-07-11
Add VehicleType depot field and use that in places

N-Wouda:
This PR:

- [x] Is part of #5, working towards supporting multiple depots.

We first need to move all uses of `0` for the start/end depot to using the vehicle's depot. Those will remain hardcoded at `0` in this PR, because I'm pretty sure just replacing all the zeroes is pretty big already. A follow-up PR can then add support for multiple depots and relax that hard-coded `0`.

### Benchmark

On VRPTW. This PR (698a9d70eea87ad35c1a9e0df533896169c2aa52; job ID 3066145; 10x seeds/1h with `configs/vrptw.toml`):
```
 Avg. objective: 333449.2 (min: 3333344, max: 333586)
Avg. iterations: 87442.2 (min: 69533, max: 96136)
```

On VRPTW. This PR (698a9d70eea87ad35c1a9e0df533896169c2aa52; job ID 3067936; 10x seeds/1h with `configs/vrptw.toml`):
```
 Avg. objective: 333402.8 (min: 333220, max: 333486)
Avg. iterations: 86955.2 (min: 67928, max: 97156)
```

Current main (cad80f7895b0e91ee69745f4f40b80b873aefc7b; job ID 3067260; 10x seeds/1h with `configs/vrptw.toml`):
```
 Avg. objective: 333409.6 (min: 333273, max: 333527)
Avg. iterations: 93797.9 (min: 74649, max: 107517)
```

<details>

**Notes**:

</details>


N-Wouda:
We first need to move all uses of `0` for the start/end depot to using the vehicle's start and end depot. Those will remain hardcoded at `0` in this PR, because I'm pretty sure just replacing all the zeroes is pretty big already. A follow-up PR can then add support for multiple depots and relax that hard-coded `0`.

pyvrp/tests/test_Solution.py
```diff
@@ -47,6 +47,7 @@ def test_route_constructor_heterogeneous():
 
 def test_route_eq():
     data = read("data/OkSmall.txt")
+    data = make_heterogeneous(data, [VehicleType(10, 1), VehicleType(20, 2)])
```
N-Wouda, line 50: Strange that this had not segfaulted earlier.

pyvrp/cpp/crossover/crossover.cpp
```diff
@@ -117,13 +115,13 @@ void pyvrp::crossover::greedyRepair(Routes &routes,
...
             {
-                prev = 0;
+                prev = 0;  // TODO hardcoded depot
                 next = route[0];
             }
             else if (idx == route.size())  // try before depot
             {
                 prev = route.back();
-                next = 0;
+                next = 0;  // TODO hardcoded depot
```
N-Wouda, line 124: I don't have a good solution to this yet so I suggest we table this until the PR that actually addresses multiple depots.

pyvrp/cpp/search/Route.h
```diff
@@ -18,8 +18,10 @@ class Route
     ProblemData const &data;
     size_t const vehicleType_;
 
-    std::vector<Node *> nodes;  // List of nodes (in order) in this solution.
-    CircleSector sector;        // Circle sector of the route's clients
+    // List of nodes (in order) in this solution. Start depot is not included;
+    // last element is the endDepot. See LS for details.
+    std::vector<Node *> nodes;
```
leonlan, line 23: Question: What's the reason here again behind the start depot not being included but the end depot is included? Why can we not remove the last depot?

leonlan (approved):

pyvrp/cpp/search/Route.h
N-Wouda: I don't know? I wanted to get rid of that anyway, so let me try to remove that now.

N-Wouda: Tried doing this. It's surprisingly complicated, because many things assume we can go to the end depot (e.g., `nodes.back()` assumes nodes is non-empty, the `Route::update()` mechanism needs to handle the end depot as well). I think it's best to leave this for now.

pyvrp/cpp/Solution.cpp
```diff
@@ -247,14 +250,14 @@ Solution::Route::Route(ProblemData const &data,
     }
 
     Client const last = visits_.back();  // last client has depot as successor
-    distance_ += data.dist(last, 0);
-    duration_ += data.duration(last, 0);
+    distance_ += data.dist(last, vehType.depot);
+    duration_ += data.duration(last, vehType.depot);
```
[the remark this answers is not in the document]
N-Wouda: Tried that, it turns into a bit of a mess and I don't think we need it for multiple depots. 

More problematically: what about empty routes where start depot != end depot? There's travel there that we would not currently register. Do we want to support such cases? (not sure; I'm leaning no for now)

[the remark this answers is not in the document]
leonlan: I just checked the [definition](https://www.sciencedirect.com/science/article/pii/S0377221705006983?casa_token=lPWh9AVQc2cAAAAA:lTbsTdxb0OsZ0Ql5bAGsdCQtHx_buWmAg9fQsemAa3RM54NaNTVmPcaX_KK03i7ZKiQIC3Ye2eTE) of the classic multi-depot VRPs, which assumes that start = end depot, so yeah we probably don't need it for now.

Issue #273, opened 2023-07-11
Docs: next steps


N-Wouda:
The documentation is currently rendered using the sphinx-immaterial theme. See #270 and #271 for details. There are still some known issues, however:
- [x] The HGS pseudocode in intro to HGS does not render correctly. I opened an issue for this: https://github.com/jbms/sphinx-immaterial/issues/272.
- [x] API type hints sometimes render as `~typing.List`, rather than `List`.

Some additional things:
- [x] As a bit of a long-term goal, I'd much rather we host our stuff using GitHub, which we already use for everything else. It's not hard to build the documentation in a separate workflow, and we can just put the rendered docs in a `gh-pages` branch for GitHub to pick up.
- [x] A custom domain name, like https://pyvrp.org.
- [ ] Versioned documentation. We should start adding pinned docs versions for releases from now on, including a `latest` (last released version) and `develop` that defaults to current `main`. Or something like that. Tracked in #478.
- [ ] Previews for PRs is something GitHub is working on; that's tracked [here](https://github.com/orgs/community/discussions/7730). Also relevant: https://github.com/actions/deploy-pages/issues/180. We'll keep using RTD for this in the meantime. Tracked in #622.

N-Wouda:
I've added a workflow that renders the documentation and deploys to GitHub Pages. Together with the domain name I bought yesterday we now deploy to https://pyvrp.org/. 🎉

N-Wouda:
HGS pseudocode has been fixed upstream in the `sphinx-immaterial` theme. All we have to do is wait for that to hit PyPI.

N-Wouda:
The versioning is something I'll look into sometime soon, possibly upcoming weekend.

N-Wouda:
The notes are aligned kind of ugly here:
<img width="649" alt="image" src="https://github.com/PyVRP/PyVRP/assets/16272507/99847d0c-2270-432b-9607-e097152d33fc">
I opened an issue about this in the theme a while ago, and that has now been fixed upstream. All we have to do is wait for that fix to hit PyPI and then upgrade the theme.

N-Wouda:
Later, at some point: check if this is fixed https://github.com/jbms/sphinx-immaterial/issues/345 and update to latest Sphinx.

N-Wouda:
I'm closing this issue since all the open points are now tracked in other places.

PR #276, opened 2023-07-12
Create DOC.yml

N-Wouda:
This PR deploys our documentation to GitHub pages, rather than RTD. See #273 for details. I'll not remove RTD because it's still in use for PR builds.


N-Wouda:
Also not sure if this workflow works, so there's also that.

Issue #96, opened 2023-02-19
Refactor Node and Route structures


N-Wouda:
These structures are tied together quite closely, but that's not evident from how they are currently used. Further, these objects are clearly not in charge of their internal state: LS and the operators modify this repeatedly. 

Additionally, we do no really test these objects yet since they are mostly internal to the C++ implementation of LS. We should probably bind and test them in much the same way we did for `TimeWindowSegment`.

N-Wouda:
To further develop this a bit, I'd like `Route::Node` to be a really simple class or struct that stores only the bare minimum like
```cpp
struct Node
{
    size_t client;
    size_t position;
    Route *route;
}
```
The time window, load, and distance statistics can all be moved over to `Route`. Similarly, inserting/removing nodes is something the `Route` should be in charge of. I think we can also get rid of the linked list model with `p()` and `n()`, using positions and indexing/slicing instead, perhaps using [`std::span`](https://en.cppreference.com/w/cpp/container/span) for efficiency.

Getting to that point will take significant refactoring, but should leave us with objects that are (a) in control of their internal state, and (b) easier to test because they can only be modified in very clear ways.

wouterkool:
This may be related to #51 (proposal to use `end` index as exclusive).

N-Wouda:
We will no longer need the `position` attribute once `Route` owns its nodes. That can trivially be computed as
```cpp
int Node::position()
{
    auto const &nodes = route->nodes;
    return std::distance(nodes.begin(), this);
}
```

N-Wouda:
[from #320]

I think two/three more PRs after this:
- Make `Route` take ownership of the `Route::Node`s - no more pointers to memory owned by `LocalSearch`. The LS can maintain pointers to nodes instead, if needed. Remove the `idx` field; it can be calculated then (see #96 for details).
- Move the `Route::Node` TWS data to `Route`.
- Bind this to Python and add tests.

N-Wouda:
> - [ ] [maybe?] Make `Route` take ownership of the `Route::Node`s - no more pointers to memory owned by `LocalSearch`. The LS can maintain pointers to nodes instead, if needed. Remove the `idx` field.

I might try this sometime later, but for now I don't think this is super important.

PR #281, opened 2023-07-16
Route overhaul

N-Wouda:
This PR does some refactoring on `Node` and `Route`. Part of #96.

### Benchmarks

Current main (fe6c756f0ef981a9efe39f8cc17306b667744ebf; job ID 3069860; 10x seeds/1h with `configs/vrptw.toml`):
```
 Avg. objective: 333414.6 (min: 333218, max: 333644)
Avg. iterations: 83515.8 (min: 68249, max: 94540)
```
(with #280; don't have a commit hash for that)

This PR (4ee227150ea11f3a0bfe0f3cf977c847b07861db; job ID 3071212; 10x seeds/1h with `configs/vrptw.toml`):
```
 Avg. objective: 333512.2 (min: 333407, max: 333668)
Avg. iterations: 78339.7 (min: 62474, max: 86598)
```

This PR (505b9617d924e32488daf357fdfdab113d5a8fdd; job ID 3071365; 10x seeds/1h with `configs/vrptw.toml`):
```
 Avg. objective: 333419.6 (min: 333301, max: 333543)
Avg. iterations: 92350.6 (min: 72816, max: 101671)
```


N-Wouda:
Since this is getting pretty big already I'll do the rest in a follow-up PR.

pyvrp/cpp/search/Route.cpp
```diff
@@ -80,55 +121,52 @@ bool Route::overlapsWith(Route const &other, int const tolerance) const
...
             = TWS::merge(data.durationMatrix(), p(node)->twBefore, node->tw);
+
+        node = n(node);
     }
 
+    load += data.client(endDepot.client).demand;
+    distance += data.dist(p(&endDepot)->client, endDepot.client);
+
+    reverseDistance += data.dist(endDepot.client, p(&endDepot)->client);
+    reverseDistance -= data.dist(p(&endDepot)->client, endDepot.client);
```
leonlan, line 156: Could we rename `reverseDistance`? `reverseDistance` implies that it's the distance from `endDepot` to `startDepot`. 

I think `reversalDistance` is fine since it's consistent with `cumulatedReversalDistance`, but it'd be even better to add "delta" somewhere. But then the name will get pretty verbose so I'm not really sure if that works.

leonlan (approved):
Looks good! Nice that you managed to remove the end depot from `nodes` 👍

pyvrp/cpp/search/Route.cpp
N-Wouda: It's only used in `TwoOpt::evalWithinRoute`. I want to test whether that move actually adds any value, or can be removed. In the latter case this whole thing can go. The in-route 2OPT move evaluates reversing a route segment, which doesn't seem all that useful for VRPTW. But I'll let benchmarks for CVRP and VRPTW decide.

N-Wouda: I'd not mind replacing our `TwoOpt` with a generic, templated $k$-OPT, and then provide several such operators. That seems more useful than what we have currently with 2OPT.

N-Wouda: #285.

PR #291, opened 2023-07-20
Move docstrings into C++

N-Wouda:
This PR moves docstrings from the type stubs into the header files. It then uses [docblock](https://pypi.org/project/docblock/) to extract the documentation blocks from the headers, and places those directly into the bindings. This ensures we only have one docstring for everything, rather than separate C++ and Python docstrings.

Additionally, it stops  our abuse of the pyi type stubs.

<details>

**Notes**:

</details>


N-Wouda:
@leonlan would you still have time to look at this before TSL? It's not that big other than lots of docstrings moving around. The meat's in `extract_docstrings.py`.

PR #293, opened 2023-07-21
Update docblock to v0.0.5

N-Wouda:


N-Wouda:
Not sure if this works with the new operator matching.

PR #297, opened 2023-07-24
SearchMethod protocol

### LINKED ISSUE #46: SearchMethod protocol

N-Wouda:
This PR:

- [x] Closes #46, part of #122.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/GeneticAlgorithm.py
+++ b/pyvrp/GeneticAlgorithm.py
@@ -160,7 +160,7 @@ class GeneticAlgorithm:
             offspring = self._crossover(
                 parents, self._data, self._cost_evaluator, self._rng
             )
-            self._handle_offspring(offspring)
+            self._improve_offspring(offspring)
 
             new_best = self._cost_evaluator.cost(self._best)
 
@@ -174,7 +174,7 @@ class GeneticAlgorithm:
         end = time.perf_counter() - start
         return Result(self._best, stats, iters, end)
 
-    def _handle_offspring(self, sol: Solution):
+    def _improve_offspring(self, sol: Solution):
         def is_new_best(sol):
             cost = self._cost_evaluator.cost(sol)
             best_cost = self._cost_evaluator.cost(self._best)
--- a/pyvrp/GeneticAlgorithm.py
+++ b/pyvrp/GeneticAlgorithm.py
@@ -8,8 +8,8 @@ from .Result import Result
 from .Statistics import Statistics
 
 if TYPE_CHECKING:
-    from pyvrp.search.LocalSearch import LocalSearch
-    from pyvrp.stop import StoppingCriterion
+    from pyvrp.search.SearchMethod import SearchMethod
+    from pyvrp.stop.StoppingCriterion import StoppingCriterion
 
     from .PenaltyManager import PenaltyManager
     from .Population import Population
@@ -75,8 +75,8 @@ class GeneticAlgorithm:
         Random number generator.
     population
         Population to use.
-    local_search
-        Local search instance to use.
+    search
+        Search method to use.
     crossover_op
         Crossover operator to use for generating offspring.
     initial_solutions
@@ -96,7 +96,7 @@ class GeneticAlgorithm:
         penalty_manager: PenaltyManager,
         rng: XorShift128,
         population: Population,
-        local_search: LocalSearch,
+        search: SearchMethod,
         crossover_op: CrossoverOperator,
         initial_solutions: Collection[Solution],
         params: GeneticAlgorithmParams = GeneticAlgorithmParams(),
@@ -108,8 +108,8 @@ class GeneticAlgorithm:
         self._pm = penalty_manager
         self._rng = rng
         self._pop = population
-        self._ls = local_search
-        self._op = crossover_op
+        self._search = search
+        self._crossover = crossover_op
         self._initial_solutions = initial_solutions
         self._params = params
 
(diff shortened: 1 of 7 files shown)


N-Wouda:
@leonlan could you have a look? This is basically what we discussed in #46 the other day.

pyvrp/search/LocalSearch.py
```diff
@@ -76,7 +76,7 @@ def get_neighbours(self) -> List[List[int]]:
         """
         return self._ls.get_neighbours()
 
-    def run(
+    def __call__(
         self,
         solution: Solution,
         cost_evaluator: CostEvaluator,
```
N-Wouda, line 82: I want to see if we can move this down into C++, because that'd avoid loading/unloading solutions more than needed, and allow this to be done more efficiently (no need to get the cost evaluator involved, for example).

Only difficulty is shuffling, but I'll tinker with that a bit to figure something out.

N-Wouda: Shuffling once before starting an (otherwise deterministic) loop of search/intensify seems to be enough randomisation (for myself: job ID 3230968):
```
 Avg. objective: 333438.6 (min: 333248, max: 333589)
Avg. iterations: 55452.5 (min: 45535, max: 62129)
```
This is extremely similar to current main, so this direction could work.

N-Wouda: I've implemented this in a branch on my own fork. I'm running a job overnight (ID: 3232120) to see if that's similar to what I found above. If so I'll open a PR tomorrow.

N-Wouda: Ooh, something seems to be broken if we don't import/export every time. Cool! I'll see if I can debug that later this week.

pyvrp/search/LocalSearch.py
```diff
@@ -76,7 +76,7 @@ def get_neighbours(self) -> List[List[int]]:
         """
         return self._ls.get_neighbours()
 
-    def run(
```
wouterkool, line 79: Rather than changing `run` to `__call__`, can't we just pass `LocalSearch.run` to genetic algorithm which (I think) implements the `SearchMethod` protocol. The user would then (in theory) also have the option to pass `LocalSearch.search` or `LocalSearch.intensify`, as well as any other function (see also https://github.com/PyVRP/PyVRP/issues/46#issuecomment-1649599134).

N-Wouda: The user already has that option, because `search` and `intensify` also implement `SearchMethod`. So any of

- `GA(..., search_method=ls, ...)`
- `GA(..., search_method=ls.search, ...)`
- `GA(..., search_method=ls.intensiyf, ...)`

Should work. For `LocalSearch` to be a search method itself we do need `__call__` here, not `run`.

pyvrp/search/LocalSearch.py
wouterkool: Why do we need to move this to c++? I agree loading/unloading is unnecessary but 1) is that really a bottleneck and 2) we can also fix that by binding the loading/unloading functions seperately, so you can just do that once.

Binding load/export seperately would make the c++ api less robust (i.e. you need to load the solution before searching just like you need to set the neighbourhood), but I consider that internal as we expose the Python API and this way we keep all the flexibility at the Python level making it much easier for users to adapt (e.g. doing a max number of iterations, only intensify sometimes, etc.).

N-Wouda: We already have all that in #298 without having to break encapsulation.

wouterkool: I don't understand? In #298 the search loop is in C++, so if a user wants to customize it from Python he/she first needs to implement it in Python again and then will incur the overhead of loading/exporting solutions repeatedly which you aim to resolve. What do you mean with 'we already have all that' then?

I think moving things back from Python to C++ for no clear reason is not the direction we should go. Your primary reason seems efficiency which is unclear if it's a bottleneck?

pyvrp/GeneticAlgorithm.py
```diff
@@ -157,10 +157,10 @@ def run(self, stop: StoppingCriterion):
             curr_best = self._cost_evaluator.cost(self._best)
 
             parents = self._pop.select(self._rng, self._cost_evaluator)
-            offspring = self._op(
+            offspring = self._crossover(
                 parents, self._data, self._cost_evaluator, self._rng
             )
-            self._search(offspring)
+            self._handle_offspring(offspring)
```
leonlan, line 163: Nit: `handle_offspring` is a bit vague. We already assume that the search method improves the solution, so I'd prefer to call this `improve_offspring` (this also better reflects the description of [HGS](https://pyvrp.readthedocs.io/en/latest/setup/introduction_to_hgs.html)). 

Related: we no longer use the concept of 'educating offspring.' There's only one exception: it's still mentioned in the notebook [here](https://github.com/search?q=repo%3APyVRP%2FPyVRP%20educate&type=code). Let's get rid of that there too.

leonlan (approved):
LGTM

pyvrp/GeneticAlgorithm.py
N-Wouda: Sure thing

pyvrp/search/LocalSearch.py
N-Wouda: I'm going to copy some of this to #298 because it's off topic here.

PR #298, opened 2023-07-24
Implements LocalSearch.run in C++

N-Wouda:
See #297 for a discussion.

<details>

**Notes**:

</details>


N-Wouda:
This is broken but I don't know how yet. So a draft for now.

pyvrp/cpp/search/SwapStar.cpp
```diff
@@ -205,8 +207,11 @@ Cost SwapStar::evaluate(Route *routeU,
 
     Cost deltaCost = static_cast<Cost>(deltaDist);
 
-    // It is not possible to have UAfter == V or VAfter == U, so the positions
-    // are always strictly different
+    // We cannot have that UAfter == V or VAfter == U, as the positions must
+    // always be strictly different.
+    assert(best.VAfter->position != best.U->position);
+    assert(best.UAfter->position != best.V->position);
```
N-Wouda, line 213: These fail at some point. Still not sure why.

N-Wouda: I think I figured it out. We have a caching mechanism for route ops that we need to update.

N-Wouda: It's probably a good idea to explicitly test the claim we make here though, so I'm leaving the asserts in.

wouterkool:
See discussion here https://github.com/PyVRP/PyVRP/pull/297#discussion_r1273459586, I don't really understand why this is an improvement?

N-Wouda:
@wouterkool wrote in #297:

>Why do we need to move this to c++? I agree loading/unloading is unnecessary but 1) is that really a bottleneck and 2) we can also fix that by binding the loading/unloading functions seperately, so you can just do that once.
>
>Binding load/export seperately would make the c++ api less robust (i.e. you need to load the solution before searching just like you need to set the neighbourhood), but I consider that internal as we expose the Python API and this way we keep all the flexibility at the Python level making it much easier for users to adapt (e.g. doing a max number of iterations, only intensify sometimes, etc.).

N-Wouda:
@N-Wouda wrote in #297:

>We already have all that in https://github.com/PyVRP/PyVRP/pull/298 without having to break encapsulation.

N-Wouda:
@wouterkool wrote in #297:

>I don't understand? In https://github.com/PyVRP/PyVRP/pull/298 the search loop is in C++, so if a user wants to customize it from Python he/she first needs to implement it in Python again and then will incur the overhead of loading/exporting solutions repeatedly which you aim to resolve. What do you mean with 'we already have all that' then?
>
>I think moving things back from Python to C++ for no clear reason is not the direction we should go. Your primary reason seems efficiency which is unclear if it's a bottleneck?

leonlan (approved):

PR #301, opened 2023-07-27
Allow incomplete solutions

N-Wouda:
LNS-type methods typically iteratively apply a destroy and repair step. After the destroy step the solution is in an incomplete state, which the repair step then fixes up.

To support LNS methods, PyVRP should allow incomplete solutions. This PR does that.

<details>

**Notes**:

</details>


pyvrp/cpp/Solution.cpp
```diff
@@ -180,23 +174,6 @@ Solution::Solution(ProblemData const &data, std::vector<Route> const &routes)
...
-            msg << "Client " << client << " is required but not present.";
-            throw std::runtime_error(msg.str());
-        }
-
-        if (visits[client] > 1)
-        {
-            std::ostringstream msg;
-            msg << "Client " << client << " is visited more than once.";
-            throw std::runtime_error(msg.str());
-        }
```
N-Wouda, line 197: We also had this check because it was really easy to do given that we wanted to check required clients. I don't think it's really worth having on its own, so I have removed `visits` tracking altogether.

N-Wouda:
I'm really excited about this because once this hits PyPI, we can use PyVRP's local search as a repair operator to [ALNS](https://github.com/N-Wouda/ALNS/).

leonlan (commented):
Currently, incomplete solutions can be easily mistaken as feasible solutions. We should add an `isComplete` method here that evaluates true when all required customers are visited and check for this in the feasibility method.

wouterkool:
What @leonlan said! However, technically that adds 'completeness' as a 'feasibility' dimension next to load and time window feasibility. I wonder what this would mean for the local search? Some options:
- local search doesn't consider 'completeness' and requires solutions to be complete when loaded (e.g. a solution needs to be repaired externally before it can be used in local search)
- local search repairs incomplete solutions immediately (e.g. upon loading or by always applying a move that inserts a required client)
- local search allows incomplete solutions -> do we need additional penalty terms? I think this option would be too complex

For this PR, I suggest the first option which means we should check completeness upon loading in local search.

If we have the concept of 'incompleteness' could (later) we also use this more consistently in SREX? E.g. after the route exchange the solution is incomplete but repaired inside SREX. If we 'allow' incomplete solutions, maybe we can get this repair step out of SREX / make it a separate 'building block'?

wouterkool (changes requested):
What @leonlan said! However, technically that adds 'completeness' as a 'feasibility' dimension next to load and time window feasibility. I wonder what this would mean for the local search? Some options:
- local search doesn't consider 'completeness' and requires solutions to be complete when loaded (e.g. a solution needs to be repaired externally before it can be used in local search)
- local search repairs incomplete solutions immediately (e.g. upon loading or by always applying a move that inserts a required client)
- local search allows incomplete solutions -> do we need additional penalty terms? I think this option would be too complex

For this PR, I suggest the first option which means we should check completeness upon loading in local search.

If we have the concept of 'incompleteness' could (later) we also use this more consistently in SREX? E.g. after the route exchange the solution is incomplete but repaired inside SREX. If we 'allow' incomplete solutions, maybe we can get this repair step out of SREX / make it a separate 'building block'?

leonlan:
I'm not keen on enforcing completeness for local search. A search method does not need to produce feasible solutions, as is currently also the case with load/time-warp-infeasible solutions.

Completeness to me is just another dimension of feasibility in which we allow relaxations (as opposed to, say, routes without no depots).

N-Wouda:
> Currently, incomplete solutions can be easily mistaken as feasible solutions. We should add an `isComplete` method here that evaluates true when all required customers are visited and check for this in the feasibility method.

We'll probably also want a quantity for this, like the number of missing (but required) clients. 

>I'm not keen on enforcing completeness for local search. A search method does not need to produce feasible solutions, as is currently also the case with load/time-warp-infeasible solutions.

Especially if we want to do this we'll need penalties (@wouterkool's option 3). If we add more such soft penalty-based constraint we'll probably want to refactor how we handle all that soon. This is also important for pickup and delivery type infeasibilities which I want to work on soon-ish.

Not all of that has to be in scope here. What about the following for this PR:
- We add `bool isComplete()`/`size_t numMissingClients()` to `Solution`, and update `isFeasible`.
- We enforce `isComplete()` with a (debug-time) `assert` in `LocalSearch::loadSolution` for now, with a TODO to allow infeasibility in this regard as well. That'll then have to wait until we have refactored penalty management a bit.

leonlan:
> We'll probably also want a quantity for this, like the number of missing (but required) clients.

What's the added benefit of such a quantity, i.e., "to what extent is the solution incomplete"? I think that having a boolean or having all missing clients as a collection are interesting options here. I can't really see the usefulness of just the number.

> We add bool isComplete()/size_t numMissingClients() to Solution, and update isFeasible. 
> We enforce isComplete() with a check in LocalSearch::loadSolution for now, with a TODO to allow infeasibility in this regard as well. That'll then have to wait until we have refactored penalty management a bit.

👍

wouterkool:
> Not all of that has to be in scope here. What about the following for this PR:
> 
> * We add `bool isComplete()`/`size_t numMissingClients()` to `Solution`, and update `isFeasible`.
> * We enforce `isComplete()` with a check in `LocalSearch::loadSolution` for now, with a TODO to allow infeasibility in this regard as well. That'll then have to wait until we have refactored penalty management a bit.

Yes, sounds good!

I have no objections against supporting it in the LS ; I think it may even be interesting as it would enable LS to temporarily 'park' 'inconvenient' customers. It's just more work and complex (especially with refactoring the penalty manager) while it is not strictly necessary.

Side node: I'm curious if the 'interaction' between penalties could be relevant. With the possibility to unassign customers, strictly there is no need to allow capacity violations anymore, so in that sense 'incompleteness' and 'load excess' can be seen as redundant in relaxing the feasibility constraints.

N-Wouda:
> What's the added benefit of such a quantity, i.e., "to what extent is the solution incomplete"?

For now probably not a whole lot. But we'll need it if we add a penalty around this. I'll leave it out for now, because that's still a bit far off on the horizon. We can add it in when we actually have a use case.

N-Wouda:
> Side node: I'm curious if the 'interaction' between penalties could be relevant. With the possibility to unassign customers, strictly there is no need to allow capacity violations anymore, so in that sense 'incompleteness' and 'load excess' can be seen as redundant in relaxing the feasibility constraints.

Yeah that's a good one. Parking customers would also resolve time warp for example. But it wouldn't help much with pickup-delivery (precedence) violations, where removing things could make the penalties increase. Something to think about when we redesign the penalty system: what do we need to keep, and what is already (partially) covered by other penalties as well.

leonlan (approved):

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -282,6 +282,9 @@ void LocalSearch::update(Route *U, Route *V)
 
 void LocalSearch::loadSolution(Solution const &solution)
 {
+    if (!solution.isComplete())  // TODO allow incomplete at some point
```
wouterkool, line 285: I don't mind but I thought you preferred issues over TODO's in the code?

pyvrp/cpp/Solution.h
```diff
@@ -149,12 +150,13 @@ class Solution
...
-    Cost prizes_ = 0;             // Total collected prize value
-    Cost uncollectedPrizes_ = 0;  // Total uncollected prize value
-    Duration timeWarp_ = 0;       // Total time warp over all routes
+    size_t numClients_ = 0;         // Number of clients in the solution
+    Distance distance_ = 0;         // Total distance
+    Load excessLoad_ = 0;           // Total excess load over all routes
+    Cost prizes_ = 0;               // Total collected prize value
+    Cost uncollectedPrizes_ = 0;    // Total uncollected prize value
+    Duration timeWarp_ = 0;         // Total time warp over all routes
+    size_t numMissingClients_ = 0;  // Number of required but missing clients
```
wouterkool, line 159: Have you considered 'unplanned' rather than 'missing'? (with a slight preference for unplanned I'm fine with either)

wouterkool (approved):
LGTM

pyvrp/cpp/search/LocalSearch.cpp
N-Wouda: #307!

pyvrp/cpp/Solution.h
N-Wouda: We do use unplanned in another context (greedy repair for crossover), but I feel like that's not the same thing as missing. The unplanned customers are just all clients that are not in the current solution, but they need not be missing if they're not required. I think those that are missing are unplanned _and_ required.

PR #302, opened 2023-07-28
Simplify SWAP*

wouterkool:
This PR:

Simplifies SWAP* implementation by extracting redundant route evaluation logic into a function.

### DIFF
--- a/pyvrp/cpp/search/SwapStar.cpp
+++ b/pyvrp/cpp/search/SwapStar.cpp
@@ -93,6 +93,78 @@ std::pair<Cost, Route::Node *> SwapStar::getBestInsertPoint(
     return std::make_pair(deltaCost, p(V));
 }
 
+Cost SwapStar::evaluateRoute(Route::Node *V,
+                             Route::Node *U,
+                             Route::Node *UAfter,
+                             CostEvaluator const &costEvaluator)
+{
+    auto const route = V->route;
+    Cost deltaCost = 0;
+
+    auto const uDemand = data.client(best.U->client).demand;
+    auto const vDemand = data.client(best.V->client).demand;
+
+    deltaCost += costEvaluator.loadPenalty(route->load() + uDemand - vDemand,
+                                           route->capacity());
+    deltaCost -= costEvaluator.loadPenalty(route->load(), route->capacity());
+
+    deltaCost -= costEvaluator.twPenalty(route->timeWarp());
+
+    if (UAfter == p(V))
+    {
+        // Special case: insert U in place of V
+        // Remove p(V) -> V -> n(V)
+        // Add p(V) -> U -> n(V)
+        auto const deltaDist = data.dist(p(V)->client, U->client)
+                               + data.dist(U->client, n(V)->client)
+                               - data.dist(p(V)->client, V->client)
+                               - data.dist(V->client, n(V)->client);
+        deltaCost += static_cast<Cost>(deltaDist);
+
+        auto const tws = TWS::merge(
+            data.durationMatrix(), UAfter->twBefore, U->tw, n(V)->twAfter);
+        deltaCost += costEvaluator.twPenalty(tws.totalTimeWarp());
+    }
+    else
+    {
+        // Removal of V and insertion of U is in different parts of the route
+        // For V, remove p(V) -> V -> n(V) and add p(V) -> n(V)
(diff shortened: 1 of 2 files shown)


pyvrp/cpp/search/SwapStar.cpp
```diff
@@ -93,6 +93,78 @@ std::pair<Cost, Route::Node *> SwapStar::getBestInsertPoint(
     return std::make_pair(deltaCost, p(V));
 }
 
+Cost SwapStar::evaluateRoute(Route::Node *V,
+                             Route::Node *U,
+                             Route::Node *UAfter,
+                             CostEvaluator const &costEvaluator)
```
N-Wouda, line 99: We call this a `(U, V)` pair of clients in many other operators. I think it makes sense to preserve that order here as well. Can we swap the arguments to be something like
```suggestion
Cost SwapStar::evaluateMove(Route::Node *U,
                            Route::Node *V,
                            Route::Node *remove,
                            CostEvaluator const &costEvaluator)
```
with the interpretation that we're inserting `U` after `V` (like (1, 0)-exchange), while simulateously removing `remove`. Or something similar.

pyvrp/cpp/search/SwapStar.h
```diff
@@ -71,10 +71,10 @@ class SwapStar : public LocalSearchOperator<Route>
         Cost cost = 0;
 
         Route::Node *U = nullptr;
-        Route::Node *UAfter = nullptr;
+        Route::Node *UAfter = nullptr;  // Insert U after this node in route V
```
N-Wouda, line 74:
```suggestion
        Route::Node *UAfter = nullptr;  // Insert U after this node in V's route
```

pyvrp/cpp/search/SwapStar.h
```diff
@@ -71,10 +71,10 @@ class SwapStar : public LocalSearchOperator<Route>
...
 
         Route::Node *V = nullptr;
-        Route::Node *VAfter = nullptr;
+        Route::Node *VAfter = nullptr;  // Insert V after this node in route U
```
N-Wouda, line 77:
```suggestion
        Route::Node *VAfter = nullptr;  // Insert V after this node in U's route
```

pyvrp/cpp/search/SwapStar.h
```diff
@@ -97,6 +97,13 @@ class SwapStar : public LocalSearchOperator<Route>
 
     BestMove best;
 
+    // Evaluates the delta cost for route V, when inserting U after UAfter
+    // while simultaneously removing V
+    Cost evaluateRoute(Route::Node *V,
+                       Route::Node *U,
+                       Route::Node *UAfter,
+                       CostEvaluator const &costEvaluator);
```
N-Wouda, line 105: It's a bit confusing that we're using U, V for both the route and the nodes here. I don't have a good solution. Maybe the rename and interpretation I suggested above helps a bit.

pyvrp/cpp/search/SwapStar.cpp
```diff
@@ -177,117 +249,11 @@ Cost SwapStar::evaluate(Route *routeU,
...
-    deltaCost += costEvaluator.loadPenalty(routeV->load() + uDemand - vDemand,
-                                           routeV->capacity());
-    deltaCost -= costEvaluator.loadPenalty(routeV->load(), routeV->capacity());
-
-    return deltaCost;
+    auto const deltaCostU
+        = evaluateRoute(best.U, best.V, best.VAfter, costEvaluator);
+    auto const deltaCostV
+        = evaluateRoute(best.V, best.U, best.UAfter, costEvaluator);
+    return deltaCostU + deltaCostV;
```
N-Wouda, line 256: Bit of a nit, but I think we can simplify this further:
```suggestion
    return evaluateMove(best.U, best.V, best.VAfter, costEvaluator)
           + evaluateMove(best.V, best.U, best.UAfter, costEvaluator);
```

N-Wouda (commented):
Much better! I wonder what we'll need to change further to be able to express SWAP* in terms of a sequence of $(1, 0)$ and $(1, 1)$ exchange moves.

N-Wouda:
I'll try to finish this one as well to clean up the PR queue a bit.

N-Wouda (approved):
LGTM. Thanks @wouterkool!

N-Wouda:
I'll wait for the CI to pass and then merge it in.

PR #304, opened 2023-07-28
Add Evaluatable concept

N-Wouda:
This PR contributes a `concept`/interface/requirement for what exactly can be passed into `CostEvaluator::cost` and the like. This might be helpful because we have many things that can be evaluated, like whole solutions, routes, etc.

<details>

**Notes**:

</details>


N-Wouda:
This is pretty recent C++ so I'm not sure if all compilers in the CI will take this.

N-Wouda:
Cool that this just compiles!

pyvrp/cpp/CostEvaluator.h
```diff
@@ -4,8 +4,24 @@
...
 #include "Solution.h"
 
+#include <concepts>
+#include <limits>
+
 namespace pyvrp
 {
+// The following methods must be implemented for a type to be evaluatable by
+// the CostEvaluator.
+template <typename T> concept Evaluatable = requires(T arg)
```
wouterkool, line 14: Not very strong on this but just wondering if we should we make the name a bit more specific? Since e.g. a move is also 'evaluatable' (it even has a function 'evaluate') but that's something different? Not sure though what the name should be. Maybe `CostEvaluatable` aligns better as subject for `CostEvaluator` but doesn't completely resolve ambiguity with moves...

wouterkool (approved):
I think this is very cool if it works on the platforms we support! So (without knowing much of the 'concept' concept) LGTM.

pyvrp/cpp/CostEvaluator.h
N-Wouda: There's no ambiguity if we move this into `CostEvaluator` since that clearly specifies the scope, which I'd prefer over a longer name.

N-Wouda: Ah that's not allowed as concepts are not allowed in class scope. Oh well. `CostEvaluatable` it is.

Issue #305, opened 2023-07-28
Multiple depots


N-Wouda:
#### Is your feature request related to a problem? Please describe

N/A

#### Describe the solution you'd like

Support for multiple depots. This is discussed already a little bit in #5, but there is no issue dedicated to just this extension.

#### Describe alternatives you have considered

N/A

#### Additional context

N/A

N-Wouda:
"The" multi-depot VRP has vehicles that belong to different depots. Each vehicle starts and ends at a fixed depot, but that depot need not be the same for different vehicles. I propose we start with this classic variant.

Other variants also allow travel between depots, but we need not support that out of the box because then it gets tricky with e.g. empty routes.

N-Wouda:
An earlier PR #274 removed several `0` in favour of a depot field on the vehicle types. That's a start, but the work is still unfinished, mostly because `crossover::greedyRepair` is not currently able to handle different vehicle types/depots. I suggest we fix that by splitting off `greedyRepair` in the manner of #182. We might want to add that to a `pyvrp.repair` module.

This issue becomes a lot easier if we can figure out #182 first.

N-Wouda:
Not sure what to do about BPD.

N-Wouda:
#360 gets us like 90% of the way there. All we still have to figure out is:
- What to do about BPD. It has no way of knowing how many depots there are, because we removed its access to the data instance a while back. There are IMO two solutions:
  - Count the depots. Since these are always unassigned, that means we will never have 100% broken pairs again, but I don't think that's a problem at all.
  - Have BPD take the data instance as an argument. That feels more heavy-handed to me.
- Find benchmark instances and figure out how to read them with VRPLIB. Should not be too hard; I know there are some around. But they'll probably not be in VRPLIB format.

leonlan:
Here are benchmark instances for:
- [MDVRPTW](http://www.vrp-rep.org/variants/item/mdvrptw.html): Vidal uses 28 instances from 360-960 customers. Cordeau proposes 20 instances with 48-288 customers (probably not useful). See [Vidal et al. (2012)](https://pubsonline.informs.org/doi/abs/10.1287/opre.1120.1048?casa_token=Y3sTFfPqZF0AAAAA:Q639iEOqjCcAT2V1j6XYrGjtt8PcGGZcnSHZjHAgRA6M7FWOODeXNtn3dUqjnYHHyw0BG9ZjO-1PDw) for more details about MDVRP.

[...]

PR #306, opened 2023-07-28
Validate arguments in ProblemData

leonlan:
This PR moves data checks from `read` to ProblemData and also adds more checks to the ProblemData constructor.

<details>

**Notes**:

</details>


pyvrp/cpp/ProblemData.cpp
```diff
@@ -72,6 +72,21 @@ ProblemData::ProblemData(std::vector<Client> const &clients,
...
+    auto const &depot = clients[0];
+
+    if (depot.demand != 0)
+        throw std::invalid_argument("Depot demand must be 0.");
+
+    if (depot.serviceDuration != 0)
+        throw std::invalid_argument("Depot service duration must be 0.");
+
+    if (depot.releaseTime != 0)
+        throw std::invalid_argument("Depot release time must be 0.");
```
leonlan, line 88: I find it odd that we can only check depot data here. Suggestions:
- Rename `Client` to `Node` and add an `is_depot` flag to the `Client` constructor. We can then put these checks in the `Node` constructor.
- Introduce a new `Depot` class that inherits from `Client`.

This also depends on #305 and @N-Wouda plans for pickup-and-delivery.

N-Wouda: I'm leaning the second, and storing depots in a separate list. They probably won't be in a class hierarchy; I'll spec out some concepts for this. But I want to try some options first before settling on any one approach.

pyvrp/cpp/bindings.cpp
```diff
@@ -157,11 +157,25 @@ PYBIND11_MODULE(_pyvrp, m)
...
                     std::vector<std::vector<pyvrp::Value>> const &dist,
                     std::vector<std::vector<pyvrp::Value>> const &dur) {
-                     Matrix<pyvrp::Distance> distMat(clients.size());
-                     Matrix<pyvrp::Duration> durMat(clients.size());
+                     auto const numNodes = clients.size();
 
-                     for (size_t row = 0; row != clients.size(); ++row)
-                         for (size_t col = 0; col != clients.size(); ++col)
+                     for (auto &row : dist)
+                         if (dist.size() < numNodes || row.size() < numNodes)
```
N-Wouda, line 163: Should these not be exactly `numNodes`?

pyvrp/cpp/bindings.cpp
```diff
@@ -157,11 +157,25 @@ PYBIND11_MODULE(_pyvrp, m)
...
+                             throw std::invalid_argument(
+                                 "Distance matrix dimensions are smaller than "
+                                 "the number of clients.");
+
+                     for (auto &row : dur)
+                         if (dur.size() < numNodes || row.size() < numNodes)
```
N-Wouda, line 169: Same here.

pyvrp/cpp/bindings.cpp
```diff
@@ -157,11 +157,25 @@ PYBIND11_MODULE(_pyvrp, m)
                     std::vector<ProblemData::VehicleType> const &vehicleTypes,
                     std::vector<std::vector<pyvrp::Value>> const &dist,
                     std::vector<std::vector<pyvrp::Value>> const &dur) {
-                     Matrix<pyvrp::Distance> distMat(clients.size());
-                     Matrix<pyvrp::Duration> durMat(clients.size());
+                     auto const numNodes = clients.size();
```
N-Wouda, line 160: I want to avoid cluttering the bindings too much, but I don't think there's a better way to check this right now. Maybe once we have support for `py::array` in #61.

pyvrp/read.py
```diff
@@ -156,21 +156,6 @@ def read(
...
-        raise ValueError("Depot start of time window must be 0")
-
-    if service_times[0] != 0:
-        raise ValueError("Depot service duration must be 0")
-
-    if release_times[0] != 0:
-        raise ValueError("Depot release time must be 0")
-
-    if (time_windows[:, 0] > time_windows[:, 1]).any():
-        raise ValueError("Time window cannot start after end")
```
N-Wouda, line 172: Can we add a check like this to `ProblemData::Client`'s constructor?

pyvrp/tests/test_ProblemData.py
```diff
@@ -83,6 +83,96 @@ def test_raises_for_invalid_client_data(
...
+        duration_matrix=[[0]],
+    )
+
+
+def test_problem_data_matrix_dimensions():
+    """
+    Tests that the ``ProblemData`` constructor raises a ``ValueError`` when
+    the distance or duration matrix has less rows or columns than the number
+    of clients. The number of rows and columns are allowed to be larger than
+    the number of clients.
```
N-Wouda, line 139: I'm not sure if we should allow this. Why would we not want to enforce that `mat.shape = (dimension, dimension)`?

pyvrp/read.py
N-Wouda: Oh nvm, we already have that!

N-Wouda (approved):
LGTM, with some comments.

N-Wouda:
(is this still a draft?)

pyvrp/tests/test_ProblemData.py
leonlan: Yeah, I changed this. Not sure why I think I thought it was a good idea to pass "larger" matrices.

Issue #163, opened 2023-03-03
How to handle randomness?


wouterkool:
Note: this issue is just to have a discussion, I'm fine if we agree the current approach is fine.

Relating to #157, I think it would be helpful if we could explicitly pass around a `rng` object so it is clear which components are deterministic or not, and we can avoid components having random state. This could be a general type where `XorShift12`8 is just one implementation.

We could even consider a 'functional' approach where the `rng` object is actually immutable but can be forked when you need to seed a function call (this is e.g. how [JAX handles randomness](https://jax.readthedocs.io/en/latest/jax-101/05-random-numbers.html)). This improves reproducability and makes it easier to parallelize code without changing the result (as you can split the rng object). I like this idea as it makes randomness easier to reason about (i.e. passing the same immutable `rng` object to a function would give the same result twice) and I think PyVRP can benefit from parallelization (i.e. multiple local searches in parallel), but it may be a bit overkill as well. Just wanted to raise it :)

N-Wouda:
Also relates to #27.

N-Wouda:
One thing that might be possible is to hook into numpy via the [API](https://numpy.org/devdocs/reference/random/c-api.html). Then we wouldn't have to roll anything ourselves, but could just ask the user to do all the usual seed stuff via numpy themselves.

Not sure (a) how to do this, (b) whether to do this, or (c) whether it'd be fast enough for us.

N-Wouda:
I did some more reading and speed would probably be OK. The how bit is basically: get the relevant numpy headers and libraries from where the package's installed, and link against those when compiling. So that should also be possible. I think it's worth exploring, because that'd mean we can get rid of our custom RNG in favour of something much more standard.

Since we're already deeply dependent on numpy anyway, I think it'd be OK to do this.

wouterkool:
We could also try to get all randomness at the Python level and make the C++ api deterministic, which would also make it easier to reason about and test, and easier to customize by the user.
* I think LocalSearch only uses `rng` to shuffle the order of nodes/routes and nodeops/routeops, which we could also set from Python. This would also make the code more flexible and easier to customize (i.e. a user could try to optimize the node or nodeop order).
* For initializing the individual, we could just create the random routes in a Python `make_individual` function (I think I already saw that coming by somewhere?).
* For Individual::SelectiveRouteExchange we could just pass `startA` and `nMovedRoutes` from Python by wrapping the function in Python.
[...]

PR #308, opened 2023-07-29
Rename XorShift128 to RandomNumberGenerator

N-Wouda:
This PR renames `XorShift128` to `RandomNumberGenerator`. That's helpful because it (a) describes what it's supposed to do rather than how it does it, and (b) frees us from the XOR-shift implementation. A next step after this is to move things over to `numpy.random`.

This PR:

- [x] Is somewhat related to #163.

<details>

**Notes**:

</details>


N-Wouda:
At some point we might be rid of `RandomNumberGenerator` altogether, but as a first step I want to try if I can change its implementation to work via `numpy.random`'s C API.

N-Wouda:
@leonlan @wouterkool could either/both of you have a look? We should have done this some time ago but it wasn't all that high on my list.

leonlan (approved):
LGTM.

PR #309, opened 2023-07-30
Allow empty local search neighbourhoods

### LINKED ISSUE #272: Solving error due to empty neighbourhood in single client instance
#### Describe the bug

I'm solving an instance with one depot and one client through the `Model` interface. When trying to solve this instance, the solver raises a RuntimeError because the default neighbourhood of such an instance is empty.

#### Code sample

```python
import matplotlib.pyplot as plt

from pyvrp import Model
from pyvrp.plotting import plot_coordinates, plot_solution

COORDS = [
    (456, 320),  # location 0 - the depot
    (228, 0),    # location 1
]
DEMANDS = [0, 1]

m = Model()
m.add_vehicle_type(capacity=15, num_available=1)
depot = m.add_depot(x=COORDS[0][0], y=COORDS[0][1])
clients = [
    m.add_client(x=COORDS[idx][0], y=COORDS[idx][1], demand=DEMANDS[idx])
    for idx in range(1, len(COORDS))
]

locations = [depot, *clients]
for frm in locations:
    for to in locations:
        distance = abs(frm.x - to.x) + abs(frm.y - to.y)  # Manhattan distance
        m.add_edge(frm, to, distance=distance)

res = m.solve(stop=MaxRuntime(1))  # one second
```

Running this in a clean Colab session produces the following error:
```python
[/usr/local/lib/python3.10/dist-packages/pyvrp/search/LocalSearch.py](https://localhost:8080/#) in __init__(self, data, rng, neighbours)
     26         self, data: ProblemData, rng: XorShift128, neighbours: List[List[int]]
     27     ):
---> 28         self._ls = _LocalSearch(data, neighbours)
     29         self._rng = rng
     30 

RuntimeError: Neighbourhood is empty.
```

I'm using version 0.4.4:
```python
INSTALLED VERSIONS
------------------
     pyvrp: 0.4.4
     numpy: 1.22.4
matplotlib: 3.7.1
    vrplib: 1.1.0
      tqdm: 4.65.0
     tomli: 2.0.1
    Python: 3.10.12
```

#### Expected Output
I expect that:
- instances with one depot and at least one client can be solved, or
- we raise an error when `Model.solve` is called when the number of clients is less than two.

leonlan:
This PR:

- [x] Closes #272. By allowing empty local search neighbourhoods, the runtime error is no longer emitted and instances with zero or one client can be solved.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/Population.py
+++ b/pyvrp/Population.py
@@ -120,7 +120,7 @@ class Population:
         # may trigger a purge which needs to compute the biased fitness which
         # requires computing the cost.
         if solution.is_feasible():
-            # Note: the feasible subpopulation actually doet not depend
+            # Note: the feasible subpopulation actually does not depend
             # on the penalty values but we use the same implementation.
             self._feas.add(solution, cost_evaluator)
         else:
--- a/pyvrp/tests/test_Model.py
+++ b/pyvrp/tests/test_Model.py
@@ -270,7 +270,8 @@ def test_model_solves_empty_instance():
 
     # Solve an instance with no clients.
     res = m.solve(stop=MaxIterations(1))
-    assert_equal(res.best.get_routes(), [])
+    solution = [r.visits() for r in res.best.get_routes()]
+    assert_equal(solution, [])
 
     # Solve an instance with one client.
     clients = [m.add_client(x=0, y=0, demand=0)]
@@ -278,4 +279,5 @@ def test_model_solves_empty_instance():
     m.add_edge(clients[0], depot, distance=0)
 
     res = m.solve(stop=MaxIterations(1))
-    assert_equal(res.best.get_routes(), [[1]])
+    solution = [r.visits() for r in res.best.get_routes()]
+    assert_equal(solution, [[1]])
--- a/pyvrp/tests/test_Model.py
+++ b/pyvrp/tests/test_Model.py
@@ -259,6 +259,7 @@ def test_data_warns_about_scaling_issues(recwarn):
         model.data()
 
 
+@mark.filterwarnings("ignore::pyvrp.exceptions.EmptySolutionWarning")
 def test_model_solves_instance_with_zero_or_one_clients():
     """
     This test exercises the bug identified in issue #272, where the model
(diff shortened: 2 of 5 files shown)


pyvrp/tests/test_Model.py
```diff
@@ -257,3 +257,28 @@ def test_data_warns_about_scaling_issues(recwarn):
     model.add_edge(depot, client, distance=MAX_USER_VALUE + 1)
     with assert_warns(ScalingWarning):
         model.data()
+
+
+@mark.filterwarnings("ignore::pyvrp.exceptions.EmptySolutionWarning")
+def test_model_solves_instance_with_zero_or_one_clients():
```
N-Wouda, line 263: Can we instead check that this warning is raised for empty solutions? The filter here seems unnecessary for what is IMHO appropriate behaviour.

pyvrp/tests/test_Model.py
```diff
@@ -257,3 +257,28 @@ def test_data_warns_about_scaling_issues(recwarn):
...
+    could not solve an instance with zero clients or just one client.
+    """
+    m = Model()
+    m.add_vehicle_type(capacity=15, num_available=1)
+    depot = m.add_depot(x=0, y=0)
+
+    # Solve an instance with no clients.
+    res = m.solve(stop=MaxIterations(1))
+    solution = [r.visits() for r in res.best.get_routes()]
+    assert_equal(solution, [])
```
N-Wouda, line 275: Something like
```python
with assert_warns(EmptyWarning):
    res = m.solve(...)

sol = [r.visits() ...]
...
```

pyvrp/Population.py
```diff
@@ -120,7 +120,7 @@ def add(self, solution: Solution, cost_evaluator: CostEvaluator):
         # may trigger a purge which needs to compute the biased fitness which
         # requires computing the cost.
         if solution.is_feasible():
-            # Note: the feasible subpopulation actually doet not depend
+            # Note: the feasible subpopulation actually does not depend
```
N-Wouda, line 123: 👍

N-Wouda (approved):
LGTM, but can we change the test to be more explicit than just filtering the warning?

N-Wouda:
@wouterkool in #176 you weren't sure about this. In light of #272 I think we should allow this. Could you also have a look?

pyvrp/tests/test_Model.py
leonlan: Done!

wouterkool (approved):
LGTM

Issue #307, opened 2023-07-29
LNS support


N-Wouda:
This does not really have a dedicated issue yet, although #182 is a part of LNS support. Similarly, the open PR #301 also touches on this.

I think we should work towards supporting operators for large neighbourhood search, perhaps in a `pyvrp.repair` module. Two obvious candidates are:
- [x] Greedy repair, as tracked in #182.
- [x] A method based on our local search. Such a method should be possible if we can somehow get our local search to insert required customers into the solution. This came up in #301.

We could at some point also add a `pyvrp.destroy` for destroy operators. That has less priority to me because the destroy step in my experience is a lot simpler than the repair one, and less performance critical (so there's no real need to integrate it into PyVRP like there is for the more expensive repair operators).

leonlan:
With the new `pyrvrp.repair` module, we can also move `crossover::greedyRepair` to there and call it something like `nearestRouteGreedyRepair`/`nearestGreedyRepair`/`nearestRepair`.

N-Wouda:
I want to try removing `crossover::greedyRepair` in favour of one of the repair methods, yes!

N-Wouda:
@leonlan what other LNS repair methods are commonly used for VRP? I think you know that better than I do. We might be able to implement one or two if they're not too complex.

leonlan:
We could implement a random repair procedure: insert each unplanned client into a random position. That's fairly easy and potentially useful in an ILS.

I think regret-$k$ by [Ropke and Pisinger (2006)](https://pubsonline.informs.org/doi/10.1287/trsc.1050.0135) is also worthwhile to implement. Rather than inserting each client one by one, the procedure aims to find in each iteration the client with the highest regret (difference between insert cost for the best route and insert cost for k-best route). This client is then inserted greedily.

It's a bit more complicated to implement though, because it is rather expensive but it can be made efficient by caching the costs.

Another interesting operator could be a granular greedy repair operator. Haven't seen that one before!

N-Wouda:
Revisiting this, I don't really need a method based on LS per se, since it's pretty easy to build such a thing on top of it by users. We already provide all the building blocks for them to do so. We have a few operators once #386 is in, and have made the LS very capable for use in LNS. So I think we're pretty much done here (for now).

All that might need to be done still is $k$-regret, if we want to.

N-Wouda:
[...]

PR #312, opened 2023-07-31
Greedy repair operator

### LINKED ISSUE #182: Greedy repair with penalized costs
We currently have the `crossover::greedyRepair` function to repair incomplete offspring solutions after replacing routes in SREX. This greedy repair operator only naively takes into account (in)feasibility due the time windows.

It would be nice to have a greedy repair operator that uses the `Node` and `Route` structures to compute penalized costs, similar to how the local search operators work. This can then be used for developing ruin-and-recreate based algorithms such as (A)LNS.

N-Wouda:
This PR adds a greedy repair operator that can be used in LNS. It also removes the greedy repair step currently in SREX, as that is no longer needed.

As such, this PR:

- [x] Closes #182, as part of #307. 

### Benchmark on VRPTW

This PR (4ba1b250388dd01204b5459d3f35c86d4abe9175; job ID 3983916):
```
                MEAN       MIN       MAX  COV
      inst
obj   C1     41379.5   41345.7   41394.7  0.0
      C2     16198.2   16194.5   16205.0  0.0
      R1     47246.5   47200.2   47270.6  0.0
      R2     27795.8   27769.2   27846.0  0.1
      RC1    44206.7   44171.7   44233.2  0.0
      RC2    23303.7   23275.5   23326.0  0.1
iters C1     68684.2   64262.8   73367.7  4.2
      C2    115839.3  107820.1  124519.2  5.1
      R1     43058.0   38854.6   46890.2  6.0
      R2     59441.9   54664.9   65557.3  6.0
      RC1    47954.3   43963.4   51621.4  6.1
      RC2    60933.0   56294.0   65026.1  5.4
```

This PR (676fa72545a995b57fd9d26d6ab0a6c2face993b; job id 3952632):
```
                MEAN      MIN       MAX  COV
      inst
obj   C1    41378.9  41352.0   41397.8   0.0
      C2    16187.5  16183.8   16189.8   0.0
      R1    47479.3  47410.5   47540.6   0.1
      R2    27852.1  27825.4   27892.7   0.1
      RC1   44282.5  44251.5   44323.3   0.1
      RC2   23316.5  23308.0   23327.1   0.0
iters C1    64720.8  54798.7   69349.0   8.1
      C2    93780.1  83973.0  100242.1   6.1
      R1    33335.3  26211.1   37358.3  11.5
      R2    40101.6  35025.4   42837.2   7.0
      RC1   41528.0  33951.0   46951.1  10.4
      RC2   39986.2  32944.2   43053.3   9.4
```

Current main (from #325):
```
                MEAN      MIN       MAX   COV
      inst
obj   C1     41357.5  41328.8   41382.7   0.0
      C2     16191.4  16186.3   16194.4   0.0
      R1     47287.8  47229.8   47325.3   0.1
      R2     27706.3  27692.4   27732.9   0.0
      RC1    44217.5  44201.7   44235.4   0.0
      RC2    23216.1  23204.4   23238.8   0.0
iters C1     73683.2  55804.6   80062.6   8.9
      C2    123546.9  95605.4  131344.4   8.4
      R1     40122.2  29845.2   44674.9  10.5
      R2     57048.2  45092.9   62249.0   8.2
      RC1    45799.1  33013.6   54637.5  11.8
      RC2    65087.7  48964.3   71194.2   9.8
```

<details>

**Notes**:

</details>


N-Wouda:
I want to use some of `pyvrp::search::Route`'s stuff here, because it's easy to use for full evaluation. But that needs a bit of refactoring before it's ready for use.

N-Wouda:
Refactoring is largely done now, so I think I can continue here using the newly improved `pyvrp::search::Route`!

pyvrp/cpp/repair/greedy_repair.cpp
```diff
@@ -0,0 +1,152 @@
...
+            visits.push_back(node->client());
+
+        solRoutes.emplace_back(data, visits, route.vehicleType());
+    }
+
+    return {data, solRoutes};
+}
+
+// Evaluates the move of inserting U after V.
+pyvrp::Cost evaluate(Route::Node *U,
```
N-Wouda, line 70: This has overlap with LS::maybeInsert, and obviously also with (1, 0)-exchange. I'll make an issue to track that we could perhaps deduplicate this at some point.

N-Wouda: #341

N-Wouda:
Alright, this seems to work. Some tests and a benchmark, and we should be good to go.

N-Wouda:
I hope this is not _much_ slower. It'll definitely be a little slower because greedy repair is expensive, but if it's a lot slower I'll investigate before we merge this.

N-Wouda:
There's a bug in here somewhere that causes the benchmarks to hang. I'll try to isolate that this afternoon. Shouldn't get in the way of a review, but obviously needs to be fixed before merging.

pyvrp/cpp/repair/greedy_repair.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    deltaCost += costEvaluator.twPenalty(vTWS.totalTimeWarp());
+    deltaCost -= costEvaluator.twPenalty(route->timeWarp());
+
+    return deltaCost;
+}
+}  // namespace
+
+Solution pyvrp::repair::greedyRepair(Solution const &solution,
+                                     DynamicBitset const &unplanned,
```
leonlan, line 108: I think we should use `std::vector` here instead of a DynamicBitset.
- Vector preserves the ordering of clients: this may impact the effectiveness of repair procedures. The current implementation always inserts clients from 1 to n.
- We're not really using the set operations anyway in greedy repair, and I also don't think there will be a use case for it in the future.

pyvrp/repair/tests/test_greedy_repair.py
```diff
@@ -0,0 +1,113 @@
...
+
+def test_raises_given_no_routes_and_unplanned_clients():
+    data = read("data/OkSmall.txt")
+    cost_eval = CostEvaluator(1, 1)
+
+    empty = Solution(data, [])
+    unplanned = DynamicBitset(data.num_clients + 1)
+
+    # This call should not raise since unplanned is empty: there are no routes
+    # to insert into, which is OK since have nothing to insert.
```
leonlan, line 17:
```suggestion
    # to insert into, which is OK since we have nothing to insert.
```

leonlan:
> There's a bug in here somewhere that causes the benchmarks to hang. I'll try to isolate that this afternoon. Shouldn't get in the way of a review, but obviously needs to be fixed before merging.

Couldn't find it unfortunately!

pyvrp/repair/tests/test_greedy_repair.py
```diff
@@ -0,0 +1,113 @@
+import pytest
+from numpy.testing import assert_, assert_equal, assert_raises
+
+from pyvrp import CostEvaluator, DynamicBitset, RandomNumberGenerator, Solution
+from pyvrp.repair import greedy_repair
+from pyvrp.tests.helpers import read
+
+
```
leonlan, line 8: Could we also add a test here that inserts a client into an empty route?

pyvrp/cpp/repair/greedy_repair.cpp
```diff
@@ -0,0 +1,164 @@
...
+                 Routes &routes,
+                 SolRoutes const &solRoutes,
+                 ProblemData const &data)
+{
+    assert(clients.empty() && routes.empty());
+
+    // Doing this avoids re-allocations, which would break the pointer structure
+    // that Route and Route::Node use.
+    clients.reserve(data.numClients() + 1);
+    routes.reserve(solRoutes.size());
```
leonlan, line 34: Should this be `routes.reserve(data.numVehicles)` instead be consistent with LS?

leonlan: Not to be consistent with LS per se, but allowing insertions into empty routes doesn't sound like a bad idea if it results in the lowest possible delta cost.

N-Wouda: I think this is a pretty bad idea, since our first use case is in SREX where we probably don't want to increase the number of routes if we can avoid it. I've bound the alternative overload with routes as well, to add the empty route test. That makes it possible for a user to do what you propose here as well, by providing a list of routes that additionally contain the empty routes. But having the option to restrict the route set can be useful to avoid increasing the number of routes I think.

(if not it'd also defeat the purpose of how we handle empty route moves in the LS, because they'd  be done by SREX anyway)

pyvrp/repair/tests/test_greedy_repair.py
N-Wouda: It's an absolute pain to document these overloads. Let's not do this again.

pyvrp/cpp/repair/greedy_repair.cpp
leonlan: OK, let's keep it like this then!

leonlan:
> There's a bug in here somewhere that causes the benchmarks to hang. I'll try to isolate that this afternoon. Shouldn't get in the way of a review, but obviously needs to be fixed before merging.

I benchmarked main overnight, but this did not run successfully. Something between https://github.com/PyVRP/PyVRP/commit/be47cfd640f60058dedfa0747a41d6b47f6ea7bf and main is merged that causes the benchmark hang.

RC1_10_1 with seed 5 and seed 10 are hanging for me.

N-Wouda:
That's good, because I couldn't find it in this PR. I'll have a look.

N-Wouda:
For me it's RC1_10_8.txt with seed 1, but only after a very long time (30min ish). This'll be annoying to debug.

N-Wouda:
I'm _guessing_ it might be #302, because that's one of the few functional changes we've made. But we can bisect the thing and then we'll know for sure.

N-Wouda:
Well, let's hope that the current benchmark completes. Should be in in an hour!

leonlan (approved):

N-Wouda:
Based on what completed (see above) I don't think this is fast enough. I'll have a look at how we can speed this up a bit later. There should be literature on this.

leonlan:
> Based on what completed (see above) I don't think this is fast enough. I'll have a look at how we can speed this up a bit later. There should be literature on this.

The performance is also influenced by #344. I got a gap of 0.65% on https://github.com/PyVRP/PyVRP/commit/28f8aedd2d805d35c739b5f045445aa1d27df006, whereas it was 0.55% before that.

N-Wouda:
Alright, I'll run a new benchmark for this now that should complete fully. (🤞)

N-Wouda:
So the benchmark completed nicely (that's a relief), but I still think this is far too slow to use in SREX. I'll have a look at whether there's something we can do about that after dinner.

leonlan:
From the benchmark results, I mainly observe that the new greedy repair results in less overall iterations. This makes sense to me, since we no longer insert in the nearest route, and thus spend way more time in SREX greedy repair. Do you also see this with the profiler?

Although unlikely, the new greedy repair could also be generating worse solutions,  because it uses an exact evaluation instead of an approximate one. That may result in faster (and worse) convergence in local search. I'm just speculating here though.

N-Wouda:
This last commit is somewhat of an experiment that I'm going to benchmark first.

N-Wouda:
OK, so just _not_ doing greedy repair in SREX is fine. Considerably faster than using this greedy repair operator, and performance is on par with main from a few weeks ago. The question then is: are we OK with SREX returning incomplete solutions? I think that's fine because there never were any feasibility guarantees anyway.

N-Wouda:
> The question then is: are we OK with SREX returning incomplete solutions?

This is the big part. The new code isn't that interesting.

N-Wouda:
> Do you see a significant increase of time spent in SREX with the profiler?

Yes, it's mostly this. The issue in SREX is the greedy repair step, which is just really expensive. The benchmark above suggests we can do without, and I think that's a reasonable way to go about this.

leonlan:
> OK, so just _not_ doing greedy repair in SREX is fine. Considerably faster than using this greedy repair operator, and performance is on par with main from a few weeks ago. The question then is: are we OK with SREX returning incomplete solutions? I think that's fine because there never were any feasibility guarantees anyway.

I'm OK with SREX returning incomplete solutions, too. We can see this as another dimension of infeasibility (without penalty at the moment, but maybe with penalties in the future).

The conclusion that I draw here from repairing offspring is that it really doesn't matter too much; just don't spend too much time and let LS do its work. So I'm fine with the current SREX + LS combination. I think it's even better if we can implement the old "nearest route greedy repair," which I think will also be useful for an ILS-based approach later. But we can do that later, I'll add it to #307.

pyvrp/crossover/tests/test_selective_route_exchange.py
```diff
@@ -127,74 +127,30 @@ def test_srex_sorts_routes():
...
     # This results in two incomplete offspring [[2, 3], [1]] and [[3], [1, 2]],
-    # which are both repaired using greedy repair. After repair, we obtain the
-    # offspring [[2, 3, 4], [1]] with cost 8188, and [[3, 4], [1, 2]] with
-    # cost 9725. The first one is returned since it has the lowest cost.
+    # the first with penalised cost 7898, the second with cost 9495. Thus, the
+    # first one is returned.
     offspring = cpp_srex((sol1, sol2), data, cost_evaluator, (0, 0), 1)
-    expected = Solution(data, [[2, 3, 4], [1]])
+    expected = Solution(data, [[2, 3], [1]])
     assert_equal(offspring, expected)
```
leonlan, line 152: This test still has to be updated

leonlan: Since we no longer repair in SREX, we could try to remove OkSmallGreedyRepair.txt too.

leonlan (approved):

PR #317, opened 2023-08-01
BugFix: `Solution::operator==` for unassigned clients

### LINKED ISSUE #315: `Solution::operator==` doesn't check assigned customers
#### Describe the bug

The `Solution::operator==` doesn't check assignedness of customers. It is quite unlikely but in theory this can make different solutions evaluate as equal, specifically since the check relies on neighbours, which is [0, 0] for both unassigned customers and customers in a route with just a single customer. This only happens if all other solution properties are equal too (like distance), so is quite unlikely, yet incorrect.

#### Code sample

```python
def test_eq_unassigned():
    clients = [
        Client(x=0, y=0),
        Client(x=0, y=1, required=False),
        Client(x=1, y=0, required=False)
    ]
    vehicle_types = [VehicleType(1, 2)]
    dist = [[0, 1, 1], [1, 0, 1], [1, 1, 0]]

    data = ProblemData(clients, vehicle_types, dist, dist)
    
    sol1 = Solution(data, [[1]])
    sol2 = Solution(data, [[2]])
    
    assert_(sol1 != sol2)
```

#### Expected Output

The two solutions should not be considered equal and the test should pass.

wouterkool:
- Add test case for `Solution::operator==` for two different solutions with unassigned clients that fails in main since these solutions are considered equal in current main as it doesn't take into account whether clients are assigned or not
- Use `-1` in `Solution::neighbours` for unassigned clients, this distinguishes unassigned clients from clients adjacent to the depot (this will also affect the
- This also affects `pyvrp::diversity::brokenPairsDistance` which will now consider these as different

This PR:

- [x] Fixes #315.

### DIFF
--- a/pyvrp/cpp/Solution.cpp
+++ b/pyvrp/cpp/Solution.cpp
@@ -108,7 +108,7 @@ bool Solution::operator==(Solution const &other) const
 }
 
 Solution::Solution(ProblemData const &data, RandomNumberGenerator &rng)
-    : neighbours(data.numClients() + 1, {0, 0})
+    : neighbours(data.numClients() + 1, {-1, -1})
 {
     // Shuffle clients (to create random routes)
     auto clients = std::vector<int>(data.numClients());
@@ -153,7 +153,7 @@ Solution::Solution(ProblemData const &data,
 }
 
 Solution::Solution(ProblemData const &data, std::vector<Route> const &routes)
-    : routes_(routes), neighbours(data.numClients() + 1, {0, 0})
+    : routes_(routes), neighbours(data.numClients() + 1, {-1, -1})
 {
     if (routes.size() > data.numVehicles())
     {
--- a/pyvrp/cpp/Solution.h
+++ b/pyvrp/cpp/Solution.h
@@ -204,7 +204,7 @@ private:
     Routes routes_;
     std::vector<std::pair<Client, Client>> neighbours;  // pairs of [pred, succ]
 
-    // Determines the [pred, succ] pairs for each client.
+    // Determines the [pred, succ] pairs for each client (-1 if unassigned).
     void makeNeighbours();
 
     // Evaluates this solution's characteristics.
--- a/pyvrp/tests/test_Solution.py
+++ b/pyvrp/tests/test_Solution.py
@@ -157,15 +157,15 @@ def test_route_constructor_allows_incomplete_solutions():
 def test_get_neighbours():
     data = read("data/OkSmall.txt")
 
-    sol = Solution(data, [[3, 4], [1, 2]])
+    sol = Solution(data, [[3], [1, 2]])
     neighbours = sol.get_neighbours()
 
     expected = [
-        (0, 0),  # 0: is depot
+        (-1, -1),  # 0: is depot
         (0, 2),  # 1: between depot (0) to 2
         (1, 0),  # 2: between 1 and depot (0)
-        (0, 4),  # 3: between depot (0) and 4
-        (3, 0),  # 4: between 3 and depot (0)
+        (0, 0),  # 3: between depot (0) and depot (0)
+        (-1, -1),  # 4: unassigned
     ]
 
     assert_equal(data.num_clients, 4)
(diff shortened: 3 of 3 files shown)


N-Wouda:
I'm not a big fan of the -1 magic number, but I can live with it. I think this works, so we could merge this in as-is. 

However, I think we can do a lot better: "it's not there" can be made a lot more explicit with [`std::optional`](https://en.cppreference.com/w/cpp/utility/optional). It should not be hard to make this work in BPD either (the only place where `getNeighbours` is currently used), because we can provide a default value using `.value_or(UNASSIGNED)`, where we define `UNASSIGNED` in some anonymous namespace or something. Not sure.

On the Python side of things this is really easy as well: pybind translates a missing value to `None`, in line with `typing.Optional`.

@wouterkool do you want to try this out? I don't think it's too hard but I don't know how it works exactly. `std::optional` is still quite new (C++17), so I haven't seen it used a lot. But I think it should be fairly straightforward.

wouterkool:
I considered optional but wasn't sure if it was worth the complexity/overhead... do you like this version better?

pyvrp/cpp/Solution.cpp
```diff
@@ -108,7 +109,7 @@ bool Solution::operator==(Solution const &other) const
 }
 
 Solution::Solution(ProblemData const &data, RandomNumberGenerator &rng)
-    : neighbours(data.numClients() + 1, {0, 0})
+    : neighbours(data.numClients() + 1)
```
N-Wouda, line 112:
```suggestion
    : neighbours(data.numClients() + 1, std::nullopt)
```
is a bit more explicit

pyvrp/cpp/Solution.cpp
```diff
@@ -153,7 +154,7 @@ Solution::Solution(ProblemData const &data,
 }
 
 Solution::Solution(ProblemData const &data, std::vector<Route> const &routes)
-    : routes_(routes), neighbours(data.numClients() + 1, {0, 0})
+    : routes_(routes), neighbours(data.numClients() + 1)
```
N-Wouda, line 157:
```suggestion
    : routes_(routes), neighbours(data.numClients() + 1, std::nullopt)
```

pyvrp/cpp/Solution.h
```diff
@@ -202,9 +203,10 @@ class Solution
     size_t numMissingClients_ = 0;  // Number of required but missing clients
 
     Routes routes_;
-    std::vector<std::pair<Client, Client>> neighbours;  // pairs of [pred, succ]
+    // pairs of [pred, succ] per client, only if client is assigned
+    std::vector<std::optional<std::pair<Client, Client>>> neighbours;
```
N-Wouda, line 207:
```suggestion
    // Pairs of [pred, succ] per assigned client, null if not assigned.
    std::vector<std::optional<std::pair<Client, Client>>> neighbours;
```

pyvrp/cpp/diversity/broken_pairs_distance.cpp
```diff
@@ -13,13 +13,19 @@ double pyvrp::diversity::brokenPairsDistance(pyvrp::Solution const &first,
...
-        // broken if it is not in the second solution.
-        numBrokenPairs += fSucc != sSucc;
-        numBrokenPairs += fPred != sPred;
+            // An edge pair (fPred, j) or (j, fSucc) from the first solution is
+            // broken if it is not in the second solution.
+            numBrokenPairs += fSucc != sSucc;
+            numBrokenPairs += fPred != sPred;
+        }
+        else if (fNeighbours[j].has_value() || sNeighbours[j].has_value())
+            numBrokenPairs += 2;  // One assigned, other not so 2 broken edges
```
N-Wouda, line 28: This could be a much smaller change, right?

```cpp
std::pair<size_t, size_t> constexpr unassigned = {numClients + 1, numClients + 1};

...

auto const [fPred, fSucc] = fNeighbours[j].value_or(unassigned);
auto const [sPred, sSucc] = ...;
```

N-Wouda (commented):
I think this is a lot clearer than the `{-1, -1}` stuff before!

N-Wouda:
This is very nearly done and I'd like to have this on main, so I'll try to complete it today.

N-Wouda (approved):
LGTM. Thanks @wouterkool!

PR #319, opened 2023-08-02
Move route modification to Route

N-Wouda:
This PR moves `Route::Node::swapWith` and `Route::Node::insertAfter` to `Route`. This allows us to incrementally update the `nodes` list, ensuring it is always up-to-date. That lets us remove the linked list representation of `Node::prev` and `Node::next` altogether.

This PR:

- [x] Is part of #96.
- [ ] Adds and passes relevant tests.
- [x] Has well-formatted code and documentation.

### Benchmark on VRPTW

Benchmark because the linked list implementation was constant time and the vector-based stuff of this PR is technically linear (but cache friendly).

This PR (593d4181361130e95e457c3f067a0bc8ad56d596; job ID 3346130; 10x seeds):
```
               mean      min       max
      inst
obj   C1    41366.2  41336.6   41388.1
      C2    16192.6  16186.5   16195.6
      R1    47357.7  47310.9   47419.8
      R2    27721.3  27701.3   27752.6
      RC1   44251.9  44217.8   44294.9
      RC2   23227.2  23207.1   23252.8
iters C1    55046.2  39111.4   62840.9
      C2    90224.6  65856.4  103908.3
      R1    29471.6  21054.5   33433.9
      R2    40072.1  29772.8   47833.2
      RC1   32833.7  23804.7   40093.5
      RC2   45170.2  32032.1   52846.2
```

Fairly recent main (few days; see #298):
```
                mean      min       max
      inst
obj   C1     41359.8  41334.3   41385.0
      C2     16191.6  16186.3   16195.0
      R1     47309.9  47246.8   47351.0
      R2     27709.4  27695.4   27726.6
      RC1    44229.4  44200.8   44251.0
      RC2    23220.1  23205.1   23228.5
iters C1     67607.0  59160.1   72707.3
      C2    114033.6  97355.1  130571.2
      R1     35821.4  30369.2   38830.0
      R2     51190.2  44466.3   54158.0
      RC1    41248.0  35409.3   46337.5
      RC2    57659.0  51814.0   63145.3
```

<details>

**Notes**:

- It is essential that you add a test when making code changes.
  This keeps the code coverage level up, and helps ensure the changes work as intended.
- When fixing a bug, you must add a test that would produce the bug in the master branch, and then show that it is fixed with the new code. 
- New code additions must be well formatted. Changes should pass the pre-commit workflow, which you can set up locally using [pre-commit](https://pre-commit.com/#intro). 
- Docstring additions must render correctly, including escapes and LaTeX.

</details>

PR #320, opened 2023-08-02
Add depots to Route::nodes

N-Wouda:
This PR adds the depots to `Route::nodes`, at index 0 (`startDepot`) and index `Route::size() + 1` (`endDepot`). So this is now an inclusive/complete route.

Doing this lets us simplify a lot of edge cases, mostly by realigning `position` (1-based) and the index into `nodes` (0-based). Additionally, we can make the depots private because they can be accessed via `Route::operator[]`. Iteration over routes still iterates only over clients, and routes are empty (`Route::empty() == true`) when there are no clients in them.

This PR:

- [x] Is part of #96.

### Benchmarks on VRPTW

See #319 for an earlier batch of benchmarks.

This PR (79a1cc661163d38d997330668177829496f71e49; job ID 3350569):
```
                mean       min       max
      inst
obj   C1     41359.9   41333.3   41383.4
      C2     16191.5   16186.3   16194.4
      R1     47307.9   47264.4   47350.1
      R2     27710.4   27696.9   27728.9
      RC1    44229.0   44207.2   44246.1
      RC2    23217.8   23205.0   23226.7
iters C1     67896.1   64152.1   72730.3
      C2    114640.7  110379.3  122336.3
      R1     35754.7   34111.1   37694.9
      R2     50979.6   46661.8   52468.4
      RC1    40433.6   38074.8   44908.3
      RC2    59870.8   55444.4   62651.8
```

Fairly recent main (from a few days ago; see #298):
```
                mean      min       max
      inst
obj   C1     41359.8  41334.3   41385.0
      C2     16191.6  16186.3   16195.0
      R1     47309.9  47246.8   47351.0
      R2     27709.4  27695.4   27726.6
      RC1    44229.4  44200.8   44251.0
      RC2    23220.1  23205.1   23228.5
iters C1     67607.0  59160.1   72707.3
      C2    114033.6  97355.1  130571.2
      R1     35821.4  30369.2   38830.0
      R2     51190.2  44466.3   54158.0
      RC1    41248.0  35409.3   46337.5
      RC2    57659.0  51814.0   63145.3
```

<details>

This PR (abb6f5e48e5c4e7cc157253e4517c731500d3bfd; job ID 3348287):
```
               mean      min       max
      inst
obj   C1    41366.9  41335.8   41391.6
      C2    16192.4  16186.7   16194.7
      R1    47354.2  47294.4   47415.6
      R2    27720.2  27705.1   27747.8
      RC1   44249.6  44223.2   44276.4
      RC2   23227.1  23206.1   23250.4
iters C1    55317.1  43407.0   63864.4
      C2    91960.4  72240.9  107821.7
      R1    28772.0  22465.9   32226.3
      R2    41040.2  32322.1   48440.9
      RC1   33406.8  26224.5   37612.2
      RC2   46706.5  36193.7   51721.4
```

**Notes**:

</details>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -4,6 +4,7 @@ project(
     version: run_command('poetry', 'version', '--short', check: true).stdout(),
     default_options: [
         'cpp_std=c++20',
+        'b_ndebug=if-release',  # disables asserts in release builds
         'b_lto=true',  # sets -flto
         'werror=true',  # sets -Werror
         'warning_level=3',  # level 3 sets -Wextra and -Wpedantic
--- a/pyvrp/cpp/search/Route.cpp
+++ b/pyvrp/cpp/search/Route.cpp
@@ -113,8 +113,6 @@ void Route::swap(Node *first, Node *second)
 
 void Route::update()
 {
-    cumDist[0] = 0;
-    cumLoad[0] = 0;
     centroid = {0, 0};
 
     for (size_t idx = 1; idx != nodes.size(); ++idx)
--- a/pyvrp/cpp/search/Route.cpp
+++ b/pyvrp/cpp/search/Route.cpp
@@ -57,58 +57,58 @@ void Route::clear()
     cumLoad.push_back(0);
     cumLoad.push_back(0);
 
-    startDepot.position = 0;
+    startDepot.idx = 0;
     startDepot.twBefore = startDepot.tw;
 
-    endDepot.position = 1;
+    endDepot.idx = 1;
     endDepot.twAfter = endDepot.tw;
 }
 
-void Route::insert(size_t position, Node *node)
+void Route::insert(size_t idx, Node *node)
 {
-    assert(0 < position && position < nodes.size());
+    assert(0 < idx && idx < nodes.size());
     assert(!node->route);  // must previously have been unassigned
 
-    node->position = position;
+    node->idx = idx;
     node->route = this;
 
     cumDist.emplace_back();  // does not matter where we place these, as they
     cumLoad.emplace_back();  // will be updated by Route::update().
 
-    nodes.insert(nodes.begin() + position, node);
-    for (auto idx = position; idx != nodes.size(); ++idx)
-        nodes[idx]->position = idx;
+    nodes.insert(nodes.begin() + idx, node);
+    for (size_t after = idx; after != nodes.size(); ++after)
+        nodes[after]->idx = after;
 }
 
 void Route::push_back(Node *node) { insert(size() + 1, node); }
 
-void Route::remove(size_t position)
+void Route::remove(size_t idx)
 {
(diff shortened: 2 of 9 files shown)


N-Wouda:
Draft PR (doesn't work yet) that I'll continue with tomorrow.

N-Wouda:
I've renamed `Node::position` to `Node::idx`, because that's what it is now - the distinction between the two has become meaningless now that the routes include their depots.

N-Wouda:
This is much closer to what I had in mind in #96. I think two more PRs after this:
- Make `Route` take ownership of the `Route::Node`s - no more pointers to memory owned by `LocalSearch`. Remove the `idx` field; it can be calculated then (see #96 for details).
- Bind this to Python and add tests.

meson.build
```diff
@@ -4,6 +4,7 @@ project(
     version: run_command('poetry', 'version', '--short', check: true).stdout(),
     default_options: [
         'cpp_std=c++20',
+        'b_ndebug=if-release',  # disables asserts in release builds
```
N-Wouda, line 7: Coming from CMake (where a release build defines `NDEBUG` by default), I just *assumed* Meson would do the same. That assumption was wrong, the default is `false`. So we compiled with asserts always on. 🤡 

That immediately explains the performance issues of this PR and #319: they added a ton of assertions in really critical places. Disabling asserts in release mode should be substantially faster. I'll do a new benchmark run.

N-Wouda: Yeah, that fixes it.

pyvrp/cpp/search/Route.cpp
```diff
@@ -46,101 +46,106 @@ bool Route::overlapsWith(Route const &other, double tolerance) const
...
+    std::swap(first->idx, second->idx);
 }
 
 void Route::update()
 {
-    cumLoad.clear();
-    cumDist.clear();
-
+    cumDist[0] = 0;
+    cumLoad[0] = 0;
```
leonlan, line 117: Question: why is this needed?

leonlan (approved):

pyvrp/cpp/search/Route.h
```diff
@@ -9,6 +9,17 @@
 
 namespace pyvrp::search
 {
+/**
+ * This ``Route`` class supports fast delta cost computations and in-place
+ * modification. It can be used to implement move evaluations.
+ *
```
leonlan, line 15: We should probably add here that the in-place modifications do not automatically update the Route data, which should be done by explicitly calling `update()`.

N-Wouda:
I'll get to this in a minute. First releasing a new ALNS!

pyvrp/cpp/search/Route.cpp
N-Wouda: Yeah you're right! I'll update this.

PR #322, opened 2023-08-03
Make Client of type `size_t`

wouterkool:
This PR:
Replaces the use of `int` for `Client` by `size_t`.

### DIFF
--- a/pyvrp/cpp/Solution.cpp
+++ b/pyvrp/cpp/Solution.cpp
@@ -12,7 +12,7 @@ using pyvrp::Duration;
 using pyvrp::Load;
 using pyvrp::Solution;
 
-using Client = int;
+using Client = size_t;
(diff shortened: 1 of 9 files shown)


pyvrp/cpp/Solution.cpp
```diff
@@ -111,7 +111,7 @@ Solution::Solution(ProblemData const &data, RandomNumberGenerator &rng)
     : neighbours(data.numClients() + 1, {0, 0})
 {
     // Shuffle clients (to create random routes)
-    auto clients = std::vector<int>(data.numClients());
+    auto clients = std::vector<size_t>(data.numClients());
```
N-Wouda, line 114:
```suggestion
    auto clients = std::vector<Client>(data.numClients());
```

N-Wouda (approved):
LGTM

PR #324, opened 2023-08-04
Test Route and Route::Node

N-Wouda:
This PR adds a lot of tests to solidify the current behaviour of `Route` and `Route::Node`. It also corrects a few small issues the tests identified (none of which impacted to algorithm, but were problematic if the objects are used on their own, outside LS).

The PR:

- [x] Is part of #96.

<details>

**Notes**:

</details>


N-Wouda:
I'll first move the TWS stuff over to Route before writing tests for that, so the TWS stuff will be tested in a follow up PR.

pyvrp/search/tests/test_Route.py
leonlan, line 1: Suggestion: add a test where clients are inserted, and test that the route statistics are not yet updated. After calling `update()`, check that the attributes are updated.

leonlan (approved):

N-Wouda:
30 useful new tests. And, what's more: now that we have the `Node` and `Route` objects on the Python side, we can do really precise tests for the LS operators as well. That will be very helpful to lock down their behaviour better too.

Issue #252, opened 2023-06-22
Heterogeneous fleet: instances and solutions


N-Wouda:
From #245:

- [x] How to read heterogeneous instances and solutions
- [x] How to print heterogeneous solutions. Do we include vehicle type?
- [x] Whether we can remove `make_heterogeneous` helper [yes, done]
- [x] Do we need benchmark instances for heterogeneous capacity? [no, covered by other instances]

N-Wouda:
@leonlan especially the first item relates to VRPLIB. Do you know if there's any standard for, say, a `VEHICLE_SECTION`?

leonlan:
> @leonlan especially the first item relates to VRPLIB. Do you know if there's any standard for, say, a `VEHICLE_SECTION`?

I'll look into this in the coming week. As far as I know, TSPLIB/VRPLIB only considers homogeneous vehicle problem instances. I don't think there's a standard for a vehicle section, but I'll try to think of something.

leonlan:
I have not found any VRPLIB-related datasets that specify vehicle characteristics, so we need to come up with our own solution for this. I propose to introduce `VEHICLE_CAPACITY_SECTION`, which specifies the capacity for each vehicle. We can group by unique vehicle capacities to define vehicle types.

leonlan:
> Do we need benchmark instances for heterogeneous capacity?

There are not many benchmark instances for heterogeneous capacity. Based on the paper by Maximó, there are 18 instances with $n<120$ customers and 10 instances with $n \in [150-350]$. I don't really think it's worthwhile to benchmark this.

yang0824:
Is there a solution for the heterogeneous capacity problem in pyvrp now?

N-Wouda:
> I propose to introduce `VEHICLE_CAPACITY_SECTION`, which specifies the capacity for each vehicle. We can group by unique vehicle capacities to define vehicle types.

@leonlan I think this is the best long-term way to support multiple capacities. We can parse and group this ourselves when reading. Does this need changes to VRPLIB to work? I'm assuming no.

leonlan:
Nope, no changes are needed.

N-Wouda:
Alright then I think we should pick this up sometime soon to generate a few instances for the unit tests.

wouterkool:
[...]

PR #334, opened 2023-08-09
Heterogeneous vehicle test instance

leonlan:
This PR:

- [x] Part of #252.

<details>

**Notes**:

</details>


pyvrp/tests/helpers.py
```diff
@@ -1,30 +1,50 @@
...
+from typing import List, Optional
 
-from pyvrp import ProblemData, Solution, VehicleType
+from pyvrp import Client, ProblemData, Solution, VehicleType
 from pyvrp.read import read as _read
 from pyvrp.read import read_solution as _read_solution
 
 
-def make_heterogeneous(data: ProblemData, vehicle_types: List[VehicleType]):
+def customise(
```
leonlan, line 11: Removing `make_heterogeneous` was too much work. I instead replaced it with a `customise` function to make it more general.

N-Wouda: I'll do this in another PR if you think it's too much work.

N-Wouda (commented):
Can we rename the `VEHICLE_CAPACITY_SECTION` to just `CAPACITY_SECTION`, so it reuses the keyword already in use for vehicle capacities?

pyvrp/tests/helpers.py
leonlan: I think many tests can be rewritten without creating new data files. I'll try rewriting the tests.

Just to understand: why do you prefer to create new data files over a customize function?

N-Wouda: We've built a lot around the idea that the instance is immutable: once created, the `ProblemData` instance is _it_ - it never changes again. Using the `customise` function, that's still technically true, but only because we're actively working around this restriction by recreating the instance with different properties.

Using the `customise` functions means the test basically applies a diff to the original read instance. We have to keep track of "this part is the base data read from file" and "this is the stuff we apply on top of that". Not really an issue when we do it just once like for the vehicle types, but it's a slippery slope towards more and more parts that can be customised. Then, we're not far off from "why is my test failing and what's the actual instance we're solving here". I'd rather avoid that by not doing this at all. Then the only source of truth is the actual data file, which is easy to fully understand: just read the file.

N-Wouda:
From #359:

              I do think `make_heterogeneous` is a poor name for what it does. `replace_vehicle_types` seems better, but is quite verbose.

We might want to do what [`datetime`](https://docs.python.org/3/library/datetime.html) does: objects in that module expose a `replace()` method that can be called to replace aspects of the object. All `datetime` objects are immutable, so it returns a copy with the requested elements replaced by the arguments. Something like `ProblemData.replace` would fit in that overall setting.

_Originally posted by @N-Wouda in https://github.com/PyVRP/PyVRP/pull/359#discussion_r1309346711_

leonlan:
Closing this PR because too many conflicts, so it's easier to start over again.

PR #336, opened 2023-08-09
Import/export

### LINKED ISSUE #73: Overall Python structure
We should discuss the overall structure of the Python code, as well as how to deal with imports (absolute or relative) and what we make part of the public API (e.g. directly importable from `pyvrp`).

              @wouterkool agree. I don't really have a strong philosophy on these. In tests I typically prefer absolute imports, but in package code I use relative imports for siblings, and absolute imports (from `pyvrp`) for everything else. We should have a look at writing this down at some point. Would you be willing to open an issue for this?

_Originally posted by @N-Wouda in https://github.com/N-Wouda/pyvrp/issues/58#issuecomment-1420582556_

N-Wouda:
This PR:

- [x] Closes #73.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/GeneticAlgorithm.py
+++ b/pyvrp/GeneticAlgorithm.py
@@ -4,21 +4,20 @@ import time
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Callable, Collection, Tuple
 
-from .Result import Result
-from .Statistics import Statistics
+from pyvrp.Result import Result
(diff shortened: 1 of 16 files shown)


N-Wouda:
So... we also need to export the things we import in our package-level `__init__` files, or else linters complain in dependent code. Didn't know that. This PR fixes that using a redundant alias.

PR #337, opened 2023-08-10
Pickle support

### LINKED ISSUE #59: Persist Result or Statistics data to filesystem after a run

N-Wouda:
This PR:

- [x] Closes #59.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/tests/test_Result.py
+++ b/pyvrp/tests/test_Result.py
@@ -1,4 +1,5 @@
 import math
+import pickle
 
 from numpy.testing import assert_, assert_allclose, assert_equal, assert_raises
 from pytest import mark
@@ -82,3 +83,23 @@ def test_str_contains_essential_information(routes):
     # And make sure that all routes are printed as well.
     for route in sol.get_routes():
         assert_(str(route) in str_representation)
+
+
+def test_result_can_be_pickled():
+    data = read("data/OkSmall.txt")
+    cost_evaluator = CostEvaluator(20, 6)
+    rng = RandomNumberGenerator(seed=42)
+    pop = Population(broken_pairs_distance)
+    stats = Statistics()
+
+    for _ in range(10):
+        pop.add(Solution.make_random(data, rng), cost_evaluator)
+        stats.collect_from(pop, cost_evaluator)
+
+    best = Solution.make_random(data, rng)
+
+    before_pickle = Result(best, stats, 10, 1.2)
+    bytes = pickle.dumps(before_pickle)
+    after_pickle = pickle.loads(bytes)
+
+    assert_equal(after_pickle, before_pickle)
--- a/pyvrp/tests/test_Solution.py
+++ b/pyvrp/tests/test_Solution.py
@@ -1,3 +1,4 @@
+import pickle
 from copy import copy, deepcopy
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


pyvrp/cpp/Solution.h
```diff
@@ -188,21 +188,39 @@ class Solution
...
+              Duration travel,
+              Duration service,
+              Duration wait,
+              Duration release,
+              Duration startTime,
+              Duration slack,
+              Cost prizes,
+              std::pair<double, double> centroid,
+              VehicleType vehicleType);
     };
```
N-Wouda, line 209: Here, and for `Solution` below: I need a way to unserialise that does not require a `ProblemData` instance - because we do not have that available. So I'm restoring the complete object state via this constructor.

N-Wouda:
@leonlan this also opens up the option to store the whole `Result` object to disk in a benchmark run, rather than separately writing the solution and statistics. Shall we simplify the CLI to just have one disk output?

pyproject.toml
```diff
@@ -93,16 +93,13 @@ line-length = 79
...
 select = [
     "E", "F", "I", "NPY", "PYI", "Q", "RET", "RSE", "RUF", "SLF", "SIM", "TCH"
 ]
 
 
-[tool.ruff.per-file-ignores]
-"__init__.py" = ["F401"]  # unused imports
-
-
 [tool.ruff.isort]
```
N-Wouda, line 103: This was part of #336 but I forgot to commit it there 😞

pyvrp/tests/test_Statistics.py
```diff
@@ -31,3 +32,27 @@ def test_csv_serialises_correctly(tmp_path):
...
+        stats.collect_from(pop, cost_evaluator)
+
+    assert_equal(stats, stats)
+    assert_(stats != "test")
+
+    if num_iterations > 0:
+        assert_(stats != Statistics())
+
+        assert_equal(stats.feas_stats[0], stats.feas_stats[0])
+        assert_(stats.feas_stats[0] != "test")
```
leonlan, line 58: The `assert_equal`s here are trivial since they test against the same objects right?

Here's my suggestion:

```python
def test_eq():
    data = read("data/OkSmall.txt")
    cost_evaluator = CostEvaluator(20, 6)
    rng = RandomNumberGenerator(seed=42)
    pop = Population(broken_pairs_distance)

    stats = Statistics()
    other = Statistics()

    assert_equal(stats, other)
    assert_(stats != "test")

    # Populate the statistics object.
    for sol in make_random_solutions(10, data, rng):
        pop.add(sol, cost_evaluator)
        stats.collect_from(pop, cost_evaluator)
        other.collect_from(pop, cost_evaluator)

    assert_equal(stats, other)
    assert_(stats != Statistics())
    assert_(stats != "test")

    assert_equal(stats.feas_stats[0], stats.feas_stats[0])
    assert_(stats.feas_stats[0] != stats.feas_stats[1])
    assert_(stats.feas_stats[0] != "test")
```

leonlan (approved):

pyvrp/tests/test_Statistics.py
N-Wouda: > The `assert_equal`s here are trivial since they test against the same objects right?

No, it is not. This used to fail because of NaN's, which are never equal to themselves. In a sense it's trivial now, but it'd fail again the moment we remove the shortcut in `_Datum.__eq__`.

leonlan: Ah okay, makes much more sense then!

N-Wouda:
It'll probably be a few GB, yes (back of the envelope, for 60 instances x 10 seeds, all 100K iters: ~3GB compressed). Let's not do it then.

pyvrp/tests/test_Statistics.py
N-Wouda: I've updated the test a bit to reflect this!

PR #338, opened 2023-08-10
Cast Matrix/np.ndarray

### LINKED ISSUE #61: Improve interop between numpy and our own matrix class

N-Wouda:
This PR adds a type caster that converts between `np.ndarray` and our `Matrix` class. As such, this PR:

- [x] Closes #61.

<details>

**Notes**:

</details>


pyvrp/tests/test_ProblemData.py
```diff
@@ -185,11 +185,14 @@ def test_matrix_access():
...
         clients=clients,
         vehicle_types=[VehicleType(1, 2)],
-        distance_matrix=dist_mat,  # type: ignore
-        duration_matrix=dur_mat,  # type: ignore
+        distance_matrix=dist_mat,
+        duration_matrix=dur_mat,
     )
 
+    assert_allclose(data.distance_matrix(), dist_mat)
+    assert_allclose(data.duration_matrix(), dur_mat)
```
N-Wouda, line 193: I've added this (back in, it was there before we removed matrix access) to make sure getting the matrix stuff from C++ also works fine.

pyvrp/cpp/bindings.cpp
```diff
@@ -27,20 +28,55 @@ using pyvrp::Solution;
...
+{
+    static_assert(sizeof(T) == sizeof(pyvrp::Value));
+
+public:
+#ifdef PYVRP_DOUBLE_PRECISION
+    PYBIND11_TYPE_CASTER(Matrix<T>, _("numpy.ndarray[float]"));
+#else
+    PYBIND11_TYPE_CASTER(Matrix<T>, _("numpy.ndarray[int]"));
+#endif
+
```
leonlan, line 45: Question: this custom type casting is very new to me. What's the reasoning behind using this over how the bindings were defined previously?

leonlan (approved):

pyvrp/cpp/bindings.cpp
N-Wouda: Previously we had this `Matrix` thing in Python as well, which was a bit inconvenient to work with because nothing expects that argument. Numpy is much more convenient to work with.

Type casters convert a Python type to a C++ type, and vice versa. That's different from bindings, which merely exposes a C++ type to Python. Here we actually convert two types to/from each other. Now C++ can keep using our `Matrix` class, and pybind will convert numpy arrays to a `Matrix`when needed, or convert a `Matrix` back into a numpy array. To the outside (Python) world, it is as if `Matrix` does not exist: they're all numpy arrays.

N-Wouda: (I am like or two days of reading ahead of you on this, so I definitely don't know this perfectly either. But it seems to work and that's good enough for me for now!)

PR #340, opened 2023-08-12
Add a SWAP* test that checks in-place swaps

N-Wouda:
This PR tests an uncommon edge case we do not currently cover with tests.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/search/tests/test_SwapStar.py
+++ b/pyvrp/search/tests/test_SwapStar.py
@@ -1,7 +1,15 @@
+import numpy as np
 from numpy.testing import assert_, assert_equal
 from pytest import mark
 
-from pyvrp import CostEvaluator, RandomNumberGenerator, Solution
+from pyvrp import (
+    Client,
+    CostEvaluator,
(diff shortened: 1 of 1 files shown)


pyvrp/search/tests/test_SwapStar.py
```diff
@@ -71,3 +80,54 @@ def test_swap_star_on_RC208_instance(seed: int):
...
+    swap_star = SwapStar(data)
+
+    # Best is to exchange clients 1 and 3. The cost delta is all distance: it
+    # saves one expensive arc of cost 10, by replacing it with one of cost 1.
+    assert_equal(swap_star.evaluate(route1, route2, cost_eval), -9)
+
+    # Apply the move and test that it indeed swaps the nodes correctly.
+    swap_star.apply(route1, route2)
+    assert_(nodes[1].route is route2)
+    assert_(nodes[3].route is route1)
```
N-Wouda, line 133: I'm happy that we can do these kinds of "surgical" tests on the operators now.

Issue #341, opened 2023-08-12
Deduplicate removal/insert cost evaluations


N-Wouda:
We need the cost (delta) of just inserting or removing a node in quite a few places. Some examples:
- `LocalSearch::maybeInsert` and `LocalSearch::maybeRemove`, for prize collecting.
- In `greedyRepair` (`evaluate`), see #312.

It might be possible to implement this in a central place and re-use that implementation.

PR #354, opened 2023-08-26
Allow incomplete solutions in LS

N-Wouda:
This PR:

- [x] Is part of #307, #341.

### Benchmarks

VRPTW (1h; be86b1e3e1d6e0804657426f00eae24e35435716; job ID 3952687):
```
                MEAN       MIN       MAX  COV
      inst
obj   C1     41365.8   41352.4   41393.1  0.0
      C2     16201.1   16195.5   16205.1  0.0
      R1     47303.9   47275.2   47354.1  0.1
      R2     27774.1   27758.2   27786.7  0.0
      RC1    44221.4   44199.1   44237.9  0.0
      RC2    23307.2   23294.8   23329.8  0.1
iters C1     66204.6   63356.0   69032.7  2.9
      C2    114153.0  106311.5  123019.1  5.0
      R1     37433.0   35151.6   39375.5  5.2
      R2     50504.6   47588.9   54925.7  5.6
      RC1    44131.5   41229.7   47110.7  5.3
      RC2    58976.4   54085.0   61609.3  4.8
```
(basically same as main)

PCVRPTW (1h; be86b1e3e1d6e0804657426f00eae24e35435716; job ID 3952719):
```
Mean of gaps: 0.32%
```

<details>

**Notes**:

</details>


N-Wouda:
@leonlan could you have a look at this? It's another step towards general LNS support, and it might let us remove the greedy repair step from SREX as well in #312 (though I'm not entirely sure if we'd want to do that).

pyvrp/cpp/search/primitives.h
```diff
@@ -0,0 +1,61 @@
...
+
+// This file stores a few basic functions for (precisely) evaluating really
+// common moves. Those primitives may be useful implementing higher order
+// operators.
+namespace pyvrp::search
+{
+// Evaluates inserting U after V.
+/**
+ * Evaluates the delta cost of inserting U after V in V's route. The evaluation
+ * is exact.
```
leonlan, line 16: Question: regarding "the evaluation is exact", where in the codebase are the evaluations not exact?

pyvrp/cpp/search/primitives.h
```diff
@@ -0,0 +1,61 @@
...
+#include "CostEvaluator.h"
+#include "Measure.h"
+#include "Route.h"
+
+// This file stores a few basic functions for (precisely) evaluating really
+// common moves. Those primitives may be useful implementing higher order
+// operators.
+namespace pyvrp::search
+{
+// Evaluates inserting U after V.
```
leonlan, line 13:
```suggestion
```

leonlan (approved):
LGTM! I'm curious to see whether this will work for repairing offspring. The difference is that #312 will completely repair a solution before applying local search. Here, local search and repair are intertwined.

pyvrp/cpp/search/primitives.h
N-Wouda: A lot of places shortcut when they know the delta cost will not become negative if more is evaluated. Some places even do this inexactly, like SWAP*.

These two don't, which is needed for e.g. rewriting SWAP* to use these: that operator evaluates multiple removes and inserts before deciding on the overall delta cost. It cannot do that if these 'building blocks' already shortcut.

leonlan: Ah okay, that makes more sense now!

wouterkool:
Sorry (again) for my late involvement but if I understand correctly, this enables LS to complete incomplete solutions (which I think is great), but essentially should not change behavior if the solution is already complete, which I think is the case in our current benchmarks. Yet the code seems 10% slower (in iterations)... is that correct and can we explain that/is that worth it?

leonlan:
> Sorry (again) for my late involvement but if I understand correctly, this enables LS to complete incomplete solutions (which I think is great), but essentially should not change behavior if the solution is already complete, which I think is the case in our current benchmarks. Yet the code seems 10% slower (in iterations)... is that correct and can we explain that/is that worth it?

SREX no longer repairs the offspring (see #312). The old `crossover::greedyRepair` operator has been removed. The offspring are now repaired using LS' ability to repair solutions. That one is much slower than the old greedy repair, which was a "nearest route" repair operator.

Replacing this operator using the new `repair` protocol is proposed in #307. That one can be used to repair offspring again in SREX, which should at least fix the number of iterations.

joopvdh:
@wouterkool I wonder if the evaluation of the two offspring solutions (in `selective_route_exchange.cpp`) is still fair. In srex the two offspring solutions are evaluated using `pyvrp::costEvaluator::penalisedCost`. 
I believe that with the changes in this PR it is possible that the parent solutions do not include all required clients (`numMissingClients > 0`). Therefore, it is also possible that the number of missing clients differs among the offspring solutions. There is no penalty for missing required clients, so the it seems to be incorrect to use the `penalisedCost` method to compare the offspring solutions.

leonlan:
> @wouterkool I wonder if the evaluation of the two offspring solutions (in `selective_route_exchange.cpp`) is still fair. In srex the two offspring solutions are evaluated using `pyvrp::costEvaluator::penalisedCost`. I believe that with the changes in this PR it is possible that the parent solutions do not include all required clients (`numMissingClients > 0`). Therefore, it is also possible that the number of missing clients differs among the offspring solutions. There is no penalty for missing required clients, so the it seems to be incorrect to use the `penalisedCost` method to compare the offspring solutions.

Yeah you're right, SREX should probably now return any of the offsprings at random. I'll run a benchmark to check the performance difference.

leonlan:
> Does it not make more sense to return the offspring solution with the least number of missing clients instead of a random offspring solution?

The results for that are:
```
 mean: 333510 (min: 333368, max: 333594)
  gap: 0.57%
iters: 50886
```
So also not much of a difference either.

joopvdh:
Alright, thanks for sharing the results of the benchmarks!

wouterkool:
@leonlan thanks for running some benchmarks. Given that it doesn't make a difference, shouldn't we opt for the version that is simplest/most logical? Also I wonder what would be the results if we don't create two offspring in the first place but always use the first or second option (given that it seems not to matter much)?

leonlan:
> @leonlan thanks for running some benchmarks. Given that it doesn't make a difference, shouldn't we opt for the version that is simplest/most logical? Also I wonder what would be the results if we don't create two offspring in the first place but always use the first or second option (given that it seems not to matter much)?

I'm not opposed to having a simpler version of SREX. I won't have time to implement/benchmark it though the coming month.

PR #356, opened 2023-08-28
Cast measure

### LINKED ISSUE #355: Type casters for measure types
Our bindings have a lot of additional code since they need to convert measure types to `pyvrp::Value`. Similarly to how the `Matrix` class is handled, I think we can convert a measure type to `pyvrp::Value` using a type caster. That should simplify the bindings a bit.

N-Wouda:
This PR:

- [x] Closes #355.

<details>

**Notes**:

</details>


leonlan (approved):
LGTM!

PR #360, opened 2023-08-29
Decouple clients/depots lists

N-Wouda:
This PR:

- [x] Is part of #305.

<details>

**Notes**:

</details>


N-Wouda:
@leonlan @wouterkool I think this is ready for review. The next PR after this can probably introduce multiple depots (assuming #359 is merged, because multiple depots will need `SwapRoutes`).

leonlan (approved):

pyvrp/search/neighbourhood.py
```diff
@@ -81,23 +81,22 @@ def compute_neighbours(
...
-    proximity[:, 0] = np.inf  # clients do not neighbour depot
+    proximity[: data.num_depots, :] = np.inf  # depots have no neighbours
+    proximity[:, : data.num_depots] = np.inf  # clients do not neighbour depots
 
-    top_k = np.argsort(proximity, axis=1, kind="stable")[1:, :k]  # excl. depot
+    top_k = np.argsort(proximity, axis=1, kind="stable")[data.num_depots :, :k]
 
     if not params.symmetric_neighbours:
-        return [[], *top_k.tolist()]
+        return [[] for _ in range(data.num_depots)] + [*top_k.tolist()]
```
wouterkool, line 94:
```suggestion
        return [[] for _ in range(data.num_depots)] + top_k.tolist()
```

pyvrp/search/tests/test_MoveTwoClientsReversed.py
```diff
@@ -67,7 +67,7 @@ def test_RC208_instance(rc208, seed: int):
     ls = LocalSearch(rc208, rng, compute_neighbours(rc208, nb_params))
     ls.add_node_operator(MoveTwoClientsReversed(rc208))
 
-    single_route = list(range(1, rc208.num_clients + 1))
+    single_route = list(range(rc208.num_depots, rc208.dimension))
```
wouterkool, line 70: This comes up repeatedly and without the right knowledge it is not directly clear that `range(num_depots, dimension)` enumerates the client indices I think. Would it make sense to have a helper function? Or maybe (probably not let's at least consider the option) add `index` to `Client` such that you can iterate over `data.clients()`?

pyvrp/search/tests/test_Route.py
```diff
@@ -342,18 +344,15 @@ def test_route_tws_access(ok_small):
     the route.
     """
     route = Route(ok_small, idx=0, vehicle_type=0)
-    for client in range(1, ok_small.num_clients + 1):
+    for client in range(ok_small.num_depots, ok_small.dimension):
         route.append(Node(loc=client))
 
     route.update()
 
     for loc in [0, 1, 2, 3, 4, 5]:  # = [depot, *clients, depot]
```
wouterkool, line 352: This hardcodes the length of the route which the rest of the test doesn't so I'd suggest
```suggestion
    for loc in range(len(route) + 2):  # = [depot, *clients, depot]
```

pyvrp/tests/test_ProblemData.py
```diff
@@ -117,34 +117,47 @@ def test_raises_for_invalid_depot_data(
...
+def test_problem_data_raises_when_not_exactly_one_depot_is_provided():
     """
     Tests that the ``ProblemData`` constructor raises a ``ValueError`` when
-    no clients are provided.
+    not exactly one depot is provided.
     """
     with assert_raises(ValueError):
         ProblemData(
             clients=[],
+            depots=[],
```
wouterkool, line 136: So just to check not having a depot is not allowed but not having clients is ok?

pyvrp/tests/test_ProblemData.py
```diff
@@ -351,3 +368,36 @@ def test_vehicle_type_attribute_access():
     assert_allclose(vehicle_type.fixed_cost, 3)
     assert_allclose(vehicle_type.tw_early, 17)
     assert_allclose(vehicle_type.tw_late, 19)
+
+
+def test_depot_client_location_access():
```
wouterkool, line 373: It's great that you added this test but I'm not 100% sure about the interface. What is the benefit of having `.client()` that indexes from zero? I think it's currently used nowhere (or almost nowhere) in the code but it can cause a lot of confusion potentially.

I'd suggest either:
* not having `client(idx)` and then maybe also not have `depot(idx)` for consistency and just use `location(idx)` everywhere,
* or having `client(idx)` and `depot(idx)` using the 'absolute' index, i.e. just being convenience functions for `location(idx)`, but with a check on the `idx` (and, thus, a 'guarantee' on the returned type of the location).

I'm in favor of the second option:
* We could revert most `.location(client)` into `.client(client)` (if not all?) which is more consistent with using `.depot()` for indexing depots and additionally reduces the diff of this PR by quite a bit.
* Given that clients and depots are different things (we already have some 'conditions', i.e. prize/demand/service duration is 0), I can imagine a future where we'd strongly type them and I think this would be most compatible with that situation.

wouterkool (commented):
I think this change looks good overall and is a great step towards supporting multi-depot! Only I'm not 100% sure about having `data.client()` indexing from 0 (see comments), what do you think?

pyvrp/cpp/ProblemData.h
```diff
@@ -300,24 +374,36 @@ class ProblemData
...
+ProblemData::Client const &ProblemData::client(size_t idx) const
+{
+    return clients_[idx];
+}
+
+ProblemData::Client const &ProblemData::depot(size_t idx) const
 {
-    return clients_[client];
+    return depots_[idx];
 }
```
leonlan, line 407: On PyVRP 0.6.3, it is possible to use `ProblemData.client(idx)` for `idx > numClients`, which then returns completely random clients. Can we make this raise a runtime error? Same for depots and location.

N-Wouda: I'll add an assert to the function and move the error raising to the bindings, so we can avoid the check when this gets called from C++ (which happens a lot), but not when a user interacts with it from Python.

pyvrp/tests/test_ProblemData.py
N-Wouda: Yes

pyvrp/cpp/ProblemData.h
```diff
@@ -174,31 +175,83 @@ class ProblemData
...
+     * idx
+     *     Location index whose information to retrieve.
+     *
+     * Returns
+     * -------
+     * Client
+     *     A simple data object containing the requested location's
+     *     information.
+     */
+    [[nodiscard]] inline Client const &location(size_t idx) const;
```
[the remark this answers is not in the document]
N-Wouda: Maybe

pyvrp/tests/test_ProblemData.py
N-Wouda: I'm going for the first because it's easiest to implement, and keeps the indexing at least somewhat consistent. Indices that do not start from zero for separate things is a gotcha I'd like to avoid. Having everything in one function is somewhat consistent with e.g. `Model`.

pyvrp/search/tests/test_MoveTwoClientsReversed.py
N-Wouda: >Would it make sense to have a helper function?

I want to do less of these, not more.

> Or maybe (probably not let's at least consider the option) add `index` to `Client` such that you can iterate over `data.clients()`?

Possibly, I've thought about this. But I don't really know how to do that without breaking things I like, like `Client` being immutable after construction.

pyvrp/Model.py
```diff
@@ -82,21 +82,21 @@ def from_data(cls, data: ProblemData) -> "Model":
         Model
             A model instance representing the given data.
         """
-        clients = [data.client(idx) for idx in range(data.num_clients + 1)]
+        locs = [data.location(idx) for idx in range(data.dimension)]
```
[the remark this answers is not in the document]
N-Wouda: `dimension` is pretty standard for this, but sure.

pyvrp/search/tests/test_MoveTwoClientsReversed.py
N-Wouda: I think we can solve some of this with #384 as well.

N-Wouda:
@wouterkool when will you have time to review this again?

wouterkool (approved):
LGTM

N-Wouda:
I'm waiting for a benchmark, mostly to sanity check this. If that's all OK I'll merge this in. Barring bug fixes this is the last PR I'll work on for a few weeks. I have some other stuff to do.

PR #363, opened 2023-08-30
Remove params/attributes from API reference TOC

### LINKED ISSUE #361: Remove parameters from table of contents in API reference
It's somewhat hard to navigate the API reference, especially https://pyvrp.org/api/pyvrp.html. Part of this is due to the table of contents being cluttered with parameters.

I propose to remove parameters from table of contents. I think that already improves the navigation a lot.

leonlan:
This PR:

- [x] Closes #361.

<details>

**Notes**:

</details>

### DIFF
--- a/docs/source/api/search.rst
+++ b/docs/source/api/search.rst
@@ -2,8 +2,8 @@
    :synopsis: Search
 
 
-Search
-======
+Search methods
+==============
 
 The :mod:`pyvrp.search` module contains classes and search methods responsible for improving a newly created offspring solution.
 This happens just after :mod:`pyvrp.crossover` is performed by the :class:`~pyvrp.GeneticAlgorithm.GeneticAlgorithm`. 
--- a/docs/source/api/search.rst
+++ b/docs/source/api/search.rst
@@ -15,11 +15,13 @@ All search methods implement the :class:`SearchMethod` protocol.
 
    .. autoclass:: SearchMethod
       :members:
+      :special-members: __call___
 
 .. automodule:: pyvrp.search.LocalSearch
 
    .. autoclass:: LocalSearch
       :members:
+      :special-members: __call___
 
 .. automodule:: pyvrp.search.neighbourhood
    :members:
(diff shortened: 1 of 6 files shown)


docs/source/api/search.rst
```diff
@@ -15,11 +15,13 @@ All search methods implement the :class:`SearchMethod` protocol.
 
    .. autoclass:: SearchMethod
       :members:
+      :special-members: __call___
```
leonlan, line 18: Strange, this doesn't work for the `search` functions but it does work for `StoppingCriterion`.

N-Wouda:
> Parameters do not nicely cross-reference anymore: `numpydoc` isn't as good as `napoleon` for that. But the TOC can now be navigated much easier, which I think is more important.

Yeah, that's why I want to keep using `napoleon`. The problem you're trying to fix isn't solved by this, it's just that `numpydoc` doesn't seem to understand the code all that well. [This](https://jbms.github.io/sphinx-immaterial/apidoc/index.html#object-description-options) is a far more robust solution. I think the relevant key is `py.*Param`?

leonlan:
> > Parameters do not nicely cross-reference anymore: `numpydoc` isn't as good as `napoleon` for that. But the TOC can now be navigated much easier, which I think is more important.
> 
> Yeah, that's why I want to keep using `napoleon`. The problem you're trying to fix isn't solved by this, it's just that `numpydoc` doesn't seem to understand the code all that well. [This](https://jbms.github.io/sphinx-immaterial/apidoc/index.html#object-description-options) is a far more robust solution. I think the relevant key is `py.*Param`? Possibly also need to remove attributes, which I guess would be `py.*Attribute`.

I'll have a look!

N-Wouda (approved):

N-Wouda:
It might be possible to shorten some part using `napoleon_use_ivar = True`, but I don't know exactly what that does. I'll have a look locally.

N-Wouda:
OK nvm, this is good as is.

PR #364, opened 2023-08-30
Shift time windows

N-Wouda:
This PR:

- [x] Adds support for shift time windows on vehicle types.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/tests/test_Solution.py
+++ b/pyvrp/tests/test_Solution.py
@@ -869,6 +869,8 @@ def test_route_shift_duration(tw_early: int, tw_late: int, expected: int):
 
     # Overall route duration is, at the bare minimum, dist(0, 1) + dist(1, 2)
     # + dist(2, 0) + serv(1) + serv(2). That's 1'544 + 1'992 + 1'965 + 360
-    # + 360 = 6'221.
+    # + 360 = 6'221. We cannot service client 1 before 15'600, and it takes
+    # 1'544 to get there from the depot, so we leave at 14'056. Thus, the
+    # earliest complete time is 14'056 + 6'221 = 20'277.
     route = Route(data, [1, 2], vehicle_type=0)
     assert_allclose(route.time_warp(), expected)
--- a/pyvrp/tests/test_ProblemData.py
+++ b/pyvrp/tests/test_ProblemData.py
@@ -308,12 +308,21 @@ def test_vehicle_type_raises_invalid_data(
 def test_vehicle_type_does_not_raise_for_edge_cases():
     """
     The vehicle type constructor should allow the following edge case, of no
-    capacity, costs, and just a single vehicle.
+    capacity, costs, shift time windows, and just a single vehicle.
     """
-    vehicle_type = VehicleType(capacity=0, num_available=1, fixed_cost=0)
+    vehicle_type = VehicleType(
+        capacity=0,
+        num_available=1,
+        fixed_cost=0,
+        tw_early=0,
+        tw_late=0,
+    )
+
     assert_allclose(vehicle_type.capacity, 0)
     assert_equal(vehicle_type.num_available, 1)
     assert_equal(vehicle_type.fixed_cost, 0)
+    assert_equal(vehicle_type.tw_early, 0)
+    assert_equal(vehicle_type.tw_late, 0)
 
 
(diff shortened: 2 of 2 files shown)


pyvrp/Model.py
N-Wouda, line 1: The tricky bit here is that we need to do a bit of delayed instantiation, because the indices aren't yet available when adding data to the model (we'll only have those once we call `Model.data` to get the `ProblemData` instance). So we need to duplicate the types a bit to deal with accepting `Client` instances, rather than indices.

N-Wouda:
This doesn't necessarily need to get merged, but I need it anyway and I like to use the CI while working on it. It'll live in my own fork for now and then once I think it's done we can have a look at perhaps merging it in.

N-Wouda:
I'll rework this PR into shift time windows.

N-Wouda:
I'll finish this tomorrow. Should be really easy.

pyvrp/cpp/ProblemData.h
```diff
@@ -130,15 +149,25 @@ class ProblemData
...
+        std::optional<Duration> const twEarly;  // Start of shift
+        std::optional<Duration> const twLate;   // End of shift
 
-        VehicleType(Load capacity, size_t numAvailable, Cost fixedCost = 0);
+        VehicleType(Load capacity,
+                    size_t numAvailable,
+                    Cost fixedCost = 0,
+                    std::optional<Duration> twEarly = std::nullopt,
+                    std::optional<Duration> twLate = std::nullopt);
     };
```
N-Wouda, line 171: What I'm not sure about:

- Calling this a shift time window, and use `twEarly` and `twLate`. It makes sense to do for consistency, since we also call the depot open/close times `twEarly` and `twLate`. But `shiftStart` and `shiftEnd` could also work.
- Using `None`/`std::optional` for this. The issue is that we'd like to default to the depot's time window if this is not provided, but have no clean way to tell whether the data is provided or not if we use integers (e.g., `0` could be the default, or it could be given explicitly by the user). We could also say "anything but (0, 0) indicates user arguments, but (0, 0) means default to the depot", but I think that's not as explicit as this.

pyvrp/tests/test_Solution.py
```diff
@@ -845,3 +845,30 @@ def test_fixed_vehicle_cost(assignment: Tuple[int, int], expected: int):
...
+    data = read("data/OkSmall.txt")
+    data = make_heterogeneous(
+        data, [VehicleType(10, 2, tw_early=tw_early, tw_late=tw_late)]
+    )
+
+    # Overall route duration is, at the bare minimum, dist(0, 1) + dist(1, 2)
+    # + dist(2, 0) + serv(1) + serv(2). That's 1'544 + 1'992 + 1'965 + 360
+    # + 360 = 6'221.
+    route = Route(data, [1, 2], vehicle_type=0)
+    assert_allclose(route.time_warp(), expected)
```
leonlan, line 874: Could you make this test a bit more descriptive? For example, in the first test case it's not clear to me where the number 20'227 comes from.

pyvrp/cpp/ProblemData.h
leonlan: - I see a shift essentially as a vehicle time window, so keeping `tw_early` and `tw_late` is nice for consistency. But I can also see the value of calling it shiftStart and shiftEnd, since that better captures the meaning. I don't really have a preference for this.
- `(0, 0)` is a valid (although practically useless) shift time window so I don't think we should use that as default. I think the `None`s here are fine!

leonlan (approved):
Looks good to me 👍

wouterkool (commented):
Also too late, but great change! A few things I'm wondering:
- is the name `VehicleType` still appropriate when adding things like shifts that in the physical sense aren't properties of a vehicle type
- Do we still need the time window for the depot at all as it is redundant now? Since depot demand, service time etc. are not in use we could also choose to ignore depot time windows and only use shifts on vehicle types.
- Not a big thing, but with all extra properties on `VehicleType`, does it make more sense to have `num_available` as the first argument. Indeed I suggested it to be the second/last/optional argument (default) 1 previously, with the intention to keep it the last argument but I did not realize that would break the interface with every change, sorry for that.

PR #369, opened 2023-08-31
Fix bug

N-Wouda:
This PR fixes a bug in TwoOpt, which did not consider the fact two routes have different vehicle types, and thus different (shift) time window properties.

<details>

**Notes**:

</details>


N-Wouda:
I'm merging this once the CI completes, and will push v0.6.1 as a bug fix release to PyPI.

Issue #380, opened 2023-09-04
Heterogeneous fleet example hangs


leonlan:
#### Is your feature request related to a problem? Please describe
I have found a heterogeneous fleet instance that hangs. This example hangs when ran locally and also on Google colaboratory.
Note that it doesn't hang when the runtime is set to say 1 second.

```python
from pyvrp import Model

COORDS = [
    (456, 320),  # location 0 - the depot
    (228, 0),    # location 1
    (912, 0),    # location 2
    (0, 80),     # location 3
    (114, 80),   # location 4
    (570, 160),  # location 5
    (798, 160),  # location 6
    (342, 240),  # location 7
    (684, 240),  # location 8
    (570, 400),  # location 9
    (912, 400),  # location 10
    (114, 480),  # location 11
    (228, 480),  # location 12
    (342, 560),  # location 13
    (684, 560),  # location 14
    (0, 640),    # location 15
    (798, 640),  # location 16
]
DEMANDS = [0, 1, 1, 2, 4, 2, 4, 8, 8, 1, 2, 1, 2, 4, 4, 8, 8]

m = Model()

m.add_vehicle_type(capacity=15, num_available=4, fixed_cost=10)
m.add_vehicle_type(capacity=20, num_available=6, fixed_cost=10)

depot = m.add_depot(x=COORDS[0][0], y=COORDS[0][1])
clients = [
    m.add_client(x=COORDS[idx][0], y=COORDS[idx][1], demand=DEMANDS[idx])
    for idx in range(1, len(COORDS))
]

locations = [depot, *clients]
for frm in locations:
    for to in locations:
        distance = abs(frm.x - to.x) + abs(frm.y - to.y)  # Manhattan
        m.add_edge(frm, to, distance=distance)

from pyvrp.stop import MaxRuntime

res = m.solve(stop=MaxRuntime(10))  
print(res)
```

#### Describe the solution you'd like
This instance should be solved after ten seconds.

#### Describe alternatives you have considered
None.

#### Additional context
Some observations:
- It hangs in the application of node ops. (Could also hang in route ops, but it hangs when I add no route ops.)
- I suspect it has something to do with two opt. Exchange (1, 0) together with TwoOpt hangs. Only TwoOpt does not hang, probably due to smaller search space.
- Might be related to #369?

leonlan:
Might have to do something with TwoOpt & empty route moves.

N-Wouda:
[...]

PR #381, opened 2023-09-04
Evaluate fixed costs in TwoOpt

leonlan:
This PR fixes the bug identified in #380, which was caused by TwoOpt not considering fixed costs.

### DIFF
--- a/pyvrp/tests/test_Model.py
+++ b/pyvrp/tests/test_Model.py
@@ -339,6 +339,31 @@ def test_model_solves_instance_with_zero_or_one_clients():
     assert_equal(solution, [[1]])
 
 
+def test_model_solves_small_instance_with_fixed_costs():
+    """
+    High-level test that creates and solves a small instance with vehicle fixed
+    costs, to see if the model (and thus the underlying solution algorithm)
+    can handle that. This test exercises the bug identified in issue #380.
+    Before fixing this bug, the solver would hang on this instance.
+    """
+    m = Model()
+
+    for idx in range(2):
+        m.add_vehicle_type(capacity=0, num_available=2, fixed_cost=10)
+
+    m.add_depot(x=0, y=0, tw_early=0, tw_late=40)
+
+    for idx in range(5):
+        m.add_client(x=idx, y=idx, service_duration=1, tw_early=0, tw_late=20)
+
+    for frm in m.locations:
+        for to in m.locations:
+            m.add_edge(frm, to, distance=0, duration=5)
+
+    res = m.solve(stop=MaxIterations(2000))
+    assert_(res.is_feasible())
+
+
 def test_model_solves_small_instance_with_shift_durations():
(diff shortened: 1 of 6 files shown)


pyvrp/tests/test_Model.py
```diff
@@ -339,6 +339,31 @@ def test_model_solves_instance_with_zero_or_one_clients():
...
+    m.add_depot(x=0, y=0, tw_early=0, tw_late=40)
+
+    for idx in range(5):
+        m.add_client(x=idx, y=idx, service_duration=1, tw_early=0, tw_late=20)
+
+    for frm in m.locations:
+        for to in m.locations:
+            m.add_edge(frm, to, distance=0, duration=5)
+
+    res = m.solve(stop=MaxIterations(2000))
```
N-Wouda: Do we need this many iterations? What's the smallest amount (in, say, multiples of 100) that hangs?

N-Wouda (approved):
LGTM. Only comment I had was about the 2K iters, but you already handled that :)

N-Wouda:
Let's wait for the CI and then merge it in. I'll push a new version to PyPI once it's in main.

leonlan:
> Let's wait for the CI and then merge it in if we're hitting all cases. I'll push a new version to PyPI once it's in main.

I'll add an extra example to hit the other two cases, too.
```cpp
    if (uRoute->empty() && U->isDepot() && !n(V)->isDepot())
        deltaCost += uRoute->fixedCost();

	...

    if (!vRoute->empty() && V->isDepot() && n(U)->isDepot())
        deltaCost -= vRoute->fixedCost();
```

leonlan:
@N-Wouda do you know why the CI doesn't want to run?

```
Current runner version: '2.308.0'
Operating System
Runner Image
Runner Image Provisioner
GITHUB_TOKEN Permissions
Secret source: Actions
Prepare workflow directory
Prepare all required actions
Getting action download info
Download action repository 'actions/checkout@v3' (SHA:f43a0e5ff2bd294095638e18286ca9a3d1956744)
Error: Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_cbec4595-d85b-4884-85db-5c7f02255017/14cb0d45-847e-4586-be74-63508db3c71b.tar.gz. return code: 2.

```

N-Wouda:
We might need to update to checkout v4, which was release one hour ago (coincidence? I think not). In any case, I would not be surprised if there's an issue at GitHub, although the status page still claims Actions is healthy.

Pretty sure this is not us, however.

N-Wouda:
Relevant: https://github.com/actions/checkout/issues/949.

N-Wouda:
Maybe upgrade to checkout v4?

leonlan:
@N-Wouda all cases are (partially) hit. You can merge it in if this is OK.

Issue #345, opened 2023-08-14
Link to Github source code in documentation


leonlan:
#### Is your feature request related to a problem? Please describe
The API reference of packages like `scipy` includes links to the Github source code.

See [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.expm_multiply.html):
<img width="701" alt="image" src="https://github.com/PyVRP/PyVRP/assets/30997278/3bcc6e3d-d227-4dae-a00c-521e8a3a3ea2">

#### Describe the solution you'd like
Include a source link next to functions/classes in the documentation.

#### Additional context
This can be done using [`sphinx.ext.linkcode`](https://www.sphinx-doc.org/en/master/usage/extensions/linkcode.html). We have a mixed C++/Python library, so we need to resolve links depending on the language. I don't know how this works for C++, though.

The solution above only resolves links to the source code file but not to specific line numbers. [This](https://github.com/readthedocs/sphinx-autoapi/issues/202) issue includes some examples of how to do that.

leonlan:
I'm going to have a look at this now.

leonlan:
We are currently using `napoleon` and `numpydoc` at the same time. They kind of do the same thing, so we should just choose one ([like statsmodels does](https://github.com/statsmodels/statsmodels/blob/56e9c569fadb2805eeb3c64989ea122ea85a95af/docs/source/conf.py#L43)). I'm going for numpydocs.

N-Wouda:
Let's start with Python! We can keep this issue open for the C++ stuff.

leonlan:
With #382 the docs now link to Python source code, but it cannot yet handle native extensions. That will probably require customized parsing of the C++ code. I am not planning to do this soon so I'm removing myself from the issue for now.

N-Wouda:
I had a look at how to do this for the C++ implementations. It's difficult, so I don't think we'll ever really get around to adding source links for those, too. I'm closing this issue as a result.

PR #382, opened 2023-09-05
Link to Python source code

leonlan:
This PR partially addresses #345. It implements a custom `linkcode_resolve` function that generates a URL to the Github source code for documented objects. It currently only works for native Python objects.


docs/source/conf.py
```diff
@@ -26,6 +30,62 @@
...
+    rel_path = os.path.relpath(source_file, parent_dir_path)
+    start_line_no = inspect.getsourcelines(obj)[1]
+
+    # Get the current git revision. If it cannot be determined, return "main".
+    try:
+        cmd = "git rev-parse --short HEAD".split()
+        output = subprocess.check_output(cmd).strip()
+        revision = output.decode("utf-8")
+    except (subprocess.CalledProcessError, OSError):
+        revision = "main"
```
leonlan, line 83: Doesn't work yet for RTD.

leonlan: Maybe need to use [sphinx-git](https://sphinx-git.readthedocs.io/en/latest/) for this.

leonlan: https://readthedocs.org/api/v2/build/21846816.txt

CTRL-f "git revision" shows that the function does identify the correct git revision. But why then do the links not get resolved correctly?

leonlan: Found the problem: the URLs are wrong 😅 I ran `poetry run make docs` from inside `docs/`, so the paths are different than when it's called from the root directory.

```
https:///github.com/PyVRP/PyVRP/blob/4bf717d/../pyvrp/plotting/plot_coordinates.py#L9
https:///github.com/PyVRP/PyVRP/blob/4bf717d/../pyvrp/plotting/plot_demands.py#L9
```

leonlan:
This seems to work now. @N-Wouda could you review this?

N-Wouda:
I'll try to have a look at this tomorrow!

N-Wouda:
Is the `REVISION` value something the `linkcode` extension requires (like the function definition), or is it something we use to form the URL?

leonlan:
> Is the `REVISION` value something the `linkcode` extension requires (like the function definition), or is it something we use to form the URL?

The latter!

N-Wouda (approved):
LGTM! Let's wait for RTD to complete, and if that looks alright I think we can merge this in.

PR #376, opened 2023-09-02
Improve operator  evaluation

N-Wouda:
This PR is part of my profiling for #328. It improves a bit of the evaluation logic in Exchange operators. I think it might have some overlap with earlier PRs (e.g., #50), but I haven't checked. Seems to be a little bit better than current main, but I'll run a benchmark to be sure.

### Benchmarks on VRPTW

This PR (202fce998c1d6e443f0bb34a488f1dd550feab46; job ID 4070541):
```
                MEAN       MIN       MAX  COV
      inst
obj   C1     41380.7   41352.0   41401.0  0.0
      C2     16198.2   16194.4   16205.4  0.0
      R1     47254.0   47213.0   47271.1  0.0
      R2     27800.3   27773.5   27849.7  0.1
      RC1    44214.4   44186.6   44237.6  0.0
      RC2    23306.1   23282.3   23329.2  0.1
iters C1     66974.7   64859.8   69946.0  2.7
      C2    113820.2  110464.3  117223.8  2.0
      R1     40993.1   39720.0   42365.3  2.6
      R2     55481.5   51315.4   61336.5  5.5
      RC1    43842.9   41395.4   48010.0  4.8
      RC2    56753.7   49863.6   59919.6  5.7
```

Current main (c4873bdf18a9a0f8c780d1a641a0eeeacc8fb06d; job ID 4068464):
```
                MEAN       MIN       MAX  COV
      inst
obj   C1     41383.8   41347.6   41407.7  0.0
      C2     16198.5   16195.0   16205.0  0.0
      R1     47264.6   47212.2   47284.4  0.0
      R2     27805.4   27777.9   27851.3  0.1
      RC1    44216.6   44184.6   44253.2  0.0
      RC2    23309.6   23283.8   23334.2  0.1
iters C1     61640.8   57599.8   65465.8  4.1
      C2    103934.5  100356.9  107846.3  2.3
      R1     37931.5   34933.9   40534.5  5.4
      R2     51757.0   47558.1   56994.8  5.4
      RC1    41175.0   38850.1   43639.0  4.0
      RC2    51583.2   47265.4   56343.3  5.1
```

<details>

**Notes**:

- It is essential that you add a test when making code changes.
  This keeps the code coverage level up, and helps ensure the changes work as intended.
- When fixing a bug, you must add a test that would produce the bug in the master branch, and then show that it is fixed with the new code. 
[...]

PR #386, opened 2023-09-10
Nearest route insert

N-Wouda:
Adds _nearest route insert_, which we previously had as `crossover::greedyRepair`. It's intended to be a faster alternative to the quadratic time greedy repair operator we currently have. This_ PR:

- [x] Is part of #307.

### Benchmarks on VRPTW

This isn't really relevant anymore, because I'm not adding the operator to SREX. I'll make a new issue investigating what happened between #325 and #354.

This PR w/ nearest route insert in SREX (35822ce2c521af6e16c05361d7fb2b9593d3f080; job ID 4312717):
```
                MEAN      MIN       MAX   COV
      inst
obj   C1     41382.8  41359.0   41417.6   0.1
      C2     16205.4  16202.1   16208.6   0.0
      R1     47242.3  47201.0   47261.5   0.0
      R2     27834.7  27784.9   27867.9   0.1
      RC1    44209.9  44175.7   44226.0   0.0
      RC2    23352.6  23340.9   23361.3   0.0
iters C1     68355.8  60014.2   78929.3   8.6
      C2    115047.8  98367.5  128450.3   7.4
      R1     37874.7  34500.8   42683.1   6.5
      R2     55673.0  46289.7   69694.0  12.3
      RC1    45200.2  40464.6   58835.6  12.4
      RC2    57592.4  51078.2   65492.7   6.8
```

More or less main (the table below is from #376):
```
                MEAN       MIN       MAX  COV
      inst
obj   C1     41380.7   41352.0   41401.0  0.0
      C2     16198.2   16194.4   16205.4  0.0
      R1     47254.0   47213.0   47271.1  0.0
      R2     27800.3   27773.5   27849.7  0.1
      RC1    44214.4   44186.6   44237.6  0.0
      RC2    23306.1   23282.3   23329.2  0.1
iters C1     66974.7   64859.8   69946.0  2.7
      C2    113820.2  110464.3  117223.8  2.0
      R1     40993.1   39720.0   42365.3  2.6
      R2     55481.5   51315.4   61336.5  5.5
      RC1    43842.9   41395.4   48010.0  4.8
      RC2    56753.7   49863.6   59919.6  5.7
```

<details>

**Notes**:

</details>


N-Wouda:
@leonlan I think we should support repair operators that take lists of routes, not whole solutions. I've been thinking about this a bit, and that seems to me the most general approach: let users provide the routes. If they want to pass the routes beloning to a solution, then `sol.get_routes()` still lets them do that; and in other cases, routes give them more flexibility because then empty routes can explicitly be passed.

I'm not going to change greedy repair in this PR, but I think we should do so.

N-Wouda:
I'm not adding this back into SREX, because it doesn't seem to matter much and the current implementation is simpler.

leonlan (approved):

N-Wouda:
I would like `Solution -> Solution`, but we need the ability to at least increase the number of routes somehow. Else we're unnecessarily limiting any LNS to never increase the number of routes, which seems weird to me (if the vehicles are available, why not?).

But `Solution` currently isn't equipped to do that. So we'd have to change that first, and then settle on a single interface for the repair operators.

Issue #384, opened 2023-09-06
Development documentation


leonlan:
This issue serves to discuss and collect ideas for creating documentation for developers who want to contribute to PyVRP.

There are a lot of things that we do during our development process (adding to docs, generating stubs, benchmarking, squashing PRs, etc.), but which are not mentioned anywhere. There are also undocumented assumptions inside the codebase, such as ordering depots in the lower indices in the matrices.

We should slowly start documenting this information, not in the least because I tend to forget many things :-)

Some useful resources:
- https://numpy.org/doc/stable/dev/index.html
- https://www.statsmodels.org/dev/dev/index.html

## Small stuff 
- [x] Disable notebook execution  (maybe just add a comment here)
```python
# -- nbsphinx
nbsphinx_execute = "never"
```
- [x] Setup PyVRP development using Github codespaces.

N-Wouda:
One important addition to this, I think, is our support of the VRPLIB format. And then particular our extensions. That might just be more of a glossary of "supported keys/sections in VRPLIB-formatted files".

N-Wouda:
#399 seems somewhat related. Additionally, we need to explain the relation between the `prize` and `required` arguments to `Client` somewhere (that is, just setting a prize is insufficient to get an optional client). Might need to be updated in the API docs somewhere.

N-Wouda:
Things to discuss in the dev docs:
- The hard requirement that the depots are in the lower indices, that is, the integer range `[0, num_depots)`. And then clients in the range `[num_depots, num_depots + num_clients)`. This ties in more generally into "explain how the ProblemData object is laid out".
- The concatenation schemes used in the local search. Both the ones we have now, and the overall idea.

N-Wouda:
Skipping notebook execution is now explained in `docs/README.md` (which we link to on the contributing page to explain how to build the docs).

Basically do this:
```
SKIP_NOTEBOOKS=1 poetry run make html --directory=docs
```

N-Wouda:
#652 explains how to profile C++ stuff in the docs.

leonlan:
When I tried figuring out how to profile C++ on macOS (#652) I noticed that Instruments was not able to find the source code lines related to the samples. This might have to do with the fact that we move the shared libraries from `build/` into `pyvrp/`. This could also be related to my earlier issue with debugging mixed Python/C++ on macOS, where I couldn't step through the code in VS Code.

N-Wouda:
[...]

PR #391, opened 2023-09-15
Describe VRPLIB format

N-Wouda:
This PR:

- [x] Is part of #384.

<details>

**Notes**:

</details>

### DIFF
--- a/docs/source/dev/supported_vrplib_fields.rst
+++ b/docs/source/dev/supported_vrplib_fields.rst
@@ -1,10 +1,10 @@
 Supported VRPLIB fields
 =======================
 
-PyVRP supports reading VRPLIB-formatted files through the `VRPLIB <https://github.com/leonlan/VRPLIB>`_ package.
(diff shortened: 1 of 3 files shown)


N-Wouda:
@leonlan could you have a look at this? I'm not sure if we should say more, and what's really part of PyVRP vs VRPLIB. I'm mostly going for at least an overview of relevant fields for now.

leonlan (approved):
LGTM. 

In the coming months, I'll make an attempt to standardize the VRPLIB format when we have more support for heterogeneous fleet, multi-depot, etc. Maybe we can then refer to the VRPLIB docs. For now, this will work for PyVRP.

N-Wouda:
> I'll make an attempt to standardize the VRPLIB format when we have more support for heterogeneous fleet, multi-depot, etc.

That's the main reason I'm adding this page now. It's not too bad yet - we only have a few nonstandard fields now - but I suspect we'll need more soon and then it gets messy quick if we don't have a good description of the format somewhere.

PR #398, opened 2023-10-13
Citation metadata

N-Wouda:
Adds a badge and citation data to the README, and makes the repository citable.


N-Wouda:
(draft until I know for sure the DOI works; issue/volume/page numbers we can insert later)

PR #400, opened 2023-11-13
Don't re-open namespace std for specialisations

N-Wouda:
I read [this blog post](https://quuxplusone.github.io/blog/2021/10/27/dont-reopen-namespace-std/) on the train just yet, and they make some good points. I realised we do do this, but it's trivial to change this around so it also works without re-opening `namespace std`. This PR makes the required changes.


N-Wouda:
Not sure what's wrong with RTD, but that is clearly unrelated to this PR.

PR #407, opened 2023-11-21
Update docblock

N-Wouda:
I was toying with `alignas`, and realised `docblock` did not yet support that. This caused documentation extraction to fail. I just fixed the problem upstream, and this PR updates our dependency on `docblock` to the latest version.

<details>

**Notes**:

</details>


N-Wouda:
Ok v0.1.0 is a bit too new, let's wait 15min for everything to update.

PR #410, opened 2023-12-08
Remove superfluous compile flag in TWS

N-Wouda:
Bit of cleanup.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/cpp/TimeWindowSegment.h
+++ b/pyvrp/cpp/TimeWindowSegment.h
@@ -105,13 +105,10 @@ public:
                              Duration releaseTime);
 };
 
-TimeWindowSegment TimeWindowSegment::merge(
-    [[maybe_unused]] Matrix<Duration> const &durationMatrix,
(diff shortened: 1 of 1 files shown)


pyvrp/cpp/TimeWindowSegment.h
```diff
@@ -105,13 +105,10 @@ class TimeWindowSegment
                              Duration releaseTime);
 };
 
-TimeWindowSegment TimeWindowSegment::merge(
-    [[maybe_unused]] Matrix<Duration> const &durationMatrix,
-    [[maybe_unused]] TimeWindowSegment const &other) const
+TimeWindowSegment
+TimeWindowSegment::merge(Matrix<Duration> const &durationMatrix,
+                         TimeWindowSegment const &other) const
 {
```
N-Wouda, line 111: Method's private and only called from the public method with the same name. That method already has these flags, so there's no need to have them here, too.

PR #413, opened 2023-12-09
Add support for Python 3.12, drop Python 3.8

N-Wouda:
This PR:

- [x] Closes #408.
- [ ] Adds and passes relevant tests.
- [x] Has well-formatted code and documentation.

<details>

**Notes**:

- It is essential that you add a test when making code changes.
  This keeps the code coverage level up, and helps ensure the changes work as intended.
- When fixing a bug, you must add a test that would produce the bug in the master branch, and then show that it is fixed with the new code. 
- New code additions must be well formatted. Changes should pass the pre-commit workflow, which you can set up locally using [pre-commit](https://pre-commit.com/#intro). 
- Docstring additions must render correctly, including escapes and LaTeX.

</details>

PR #418, opened 2023-12-16
Update type hinting to use builtin list, dict, and tuple

N-Wouda:
This PR updates the type hinting to use the builtin `list`, `dict`, and `tuple` classes (rather than their equivalents from `typing`. This is supported since Python 3.9, our current minimum version. 

This PR:

- [ ] Closes #xxxx.

<details>

**Notes**:

</details>


N-Wouda:
Bit of maintenance now that we dropped Python 3.8 in #413.

PR #421, opened 2024-01-13
Add a flag that tracks whether the Route's structure has updated, but statistics have not been synced yet

N-Wouda:
This PR addresses a perhaps minor issue in `pyvrp::search::Route`: after inserting/removing nodes, the associated statistics are not updated until `Route::update()` is explicitly called. The reason is performance: updating is somewhat expensive, and typically the call-site modifying the routes knows it can delay updating until all modifications have been made. That's much more efficient.

The downside is that the `Route` object can be left in a somewhat broken state, where after updating the `Route`'s nodes one forgets to call `update()`. Using the statistics in this case is not something we currently catch in any way. This PR adds a `dirty` flag that tracks the state of `Route`, and causes asserts to fail when the statistics have not yet been updated. All this only applies in debug mode, so no runtime penalty is paid for these checks by our users.

<details>

**Notes**:

</details>


N-Wouda:
An alternative is something like the following in all statistics members:
```cpp
if (dirty)
    update();
```
This way we could even make `update()` private altogether.

The downside is that a lot of these members are used in really tight loops, where even the `if(dirty)` check adds up. Additionally, we might not want simple members to (somewhat unexpectedly) be really expensive because they're the first one called after a structure change.

I think it's reasonable to force the programmer to think about exactly where the call to `update()` goes in this case. The debug statements should at least help point out the issue in case it's forgotten.

pyvrp/search/tests/test_Route.py
```diff
@@ -285,46 +285,6 @@ def test_all_routes_overlap_with_maximum_tolerance_value(ok_small):
     assert_(route2.overlaps_with(route1, 1))
 
 
-def test_data_is_not_updated_until_update_call(ok_small):
```
N-Wouda, line 288: This was already a bit of a weird test, but such behaviour is now no longer allowed: accessing such data before calling uodate() triggers the asserts.

pyvrp/cpp/search/Route.cpp
```diff
@@ -28,7 +28,11 @@ Route::NodeStats::NodeStats(TimeWindowSegment const &tws)
 {
 }
 
-std::pair<double, double> const &Route::centroid() const { return centroid_; }
+std::pair<double, double> const &Route::centroid() const
+{
+    assert(!dirty);
+    return centroid_;
+}
```
leonlan, line 35: Do we also need to add an assertion in `Route::overlapsWith`? I think it's possible there to compare "dirty" routes since the private member `centroid_` is used.

leonlan (approved):

pyvrp/cpp/search/Route.cpp
N-Wouda: Yes, good call. One sec.

PR #423, opened 2024-01-22
Add OX for TSP

### LINKED ISSUE #235: Crossover for TSP
When using PyVRP to solve a TSP (a VRP with a single route), I think the SREX crossover will simply return one of the parents as offspring. As these are already locally optimal, the local search will not improve and the algorithm will get stuck. 

I did think of this issue, not validate it experimentally. Not sure if this is a bug or just unsupported either.

Some thoughts would be to have a different crossover for a single route.

N-Wouda:
This PR:

- [x] Closes #235.

TODO:
- [x] More tests;
- [x] Support for missing clients (prize-collecting);

<details>

**Notes**:

</details>


pyvrp/Model.py
```diff
@@ -298,6 +299,9 @@ def solve(self, stop: StoppingCriterion, seed: int = 0) -> Result:
             for _ in range(pop_params.min_pop_size)
         ]
 
-        gen_args = (data, pm, rng, pop, ls, srex, init)
+        # We use SREX when the instance is a proper VRP; else OX for TSP.
+        crossover = srex if data.num_vehicles > 1 else ox
```
N-Wouda, line 303: `Model.solve` now automatically does the right thing. If one's building their own solve routine, I think it's reasonable to assume they also know how to select an appropriate crossover operator. The exception (in OX) and warning (in SREX) should help guide that choice.

N-Wouda:
@wouterkool @leonlan alright, I think this is good to go!

pyvrp/crossover/tests/test_ordered_crossover.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # When the indices cover the whole range of clients, all clients from the
+    # first route are copied in and there is nothing left to obtain from the
+    # second route. Thus, the offspring solution is indentical to the first
+    # solution.
+    offspring = cpp_ox((sol1, sol2), pr107, (0, pr107.num_clients - 1))
+    assert_equal(offspring, sol1)
+
+    # When nothing is copied from the first route, everything is taken from
+    # the second route. Thus, the offspring solution is indentical to the
```
leonlan, line 41:
```suggestion
    # second route. Thus, the offspring solution is identical to the first
    # solution.
    offspring = cpp_ox((sol1, sol2), pr107, (0, pr107.num_clients - 1))
    assert_equal(offspring, sol1)

    # When nothing is copied from the first route, everything is taken from
    # the second route. Thus, the offspring solution is identical to the
```

pyvrp/crossover/tests/test_ordered_crossover.py
```diff
@@ -0,0 +1,86 @@
...
+    # Thus, we only get the values [2, 3] from the second route, for a total of
+    # four clients in the offspring solution.
+    offspring = cpp_ox((sol1, sol2), data, (0, 2))
+    assert_equal(offspring.num_clients(), 4)
+
+    # From the first route we take [6, 7], which are not in the second route.
+    # From the second route we get [2, 3, 4, 5], for a total of six clients.
+    offspring = cpp_ox((sol1, sol2), data, (2, 4))
+    assert_equal(offspring.num_clients(), 6)
+
```
leonlan, line 66: Can we explicitly check the resulting routes? I.e., `[4, 5, 2, 3]` for the first case and `[6, 7, 2, 3, 4, 5]` for the second (I think?).

pyvrp/crossover/tests/test_ordered_crossover.py
```diff
@@ -0,0 +1,86 @@
...
+
```
leonlan, line 67: It would also be nice to have a test case where the clients from parent 2 have to wrap around like `[p2, p1, p1, p2]`, where `p1` are clients from parent 1 and `p2` are clients from parent 2. (Not necessarily in a prize-collecting setting).

leonlan (approved):
LGTM! My only suggestion is to add one or more explicit tests for the resulting routes.

pyvrp/crossover/tests/test_ordered_crossover.py
N-Wouda: Thanks! Spelling is hard 😆

wouterkool (commented):
LGTM!

One thing I'd like to discuss though is what we think of the case where we have a VRP with > 1 vehicles where all clients can be served by a single vehicle. While not very common this is perfectly fine and can happen in practice. In this case we will get stuck without a warning.

One idea could be to just have a crossover operator (pure python) that dynamically dispatches to OX or SREX based on the number of routes in the parent solution... would this have much downside?

N-Wouda:
> One idea could be to just have a crossover operator (pure python) that dynamically dispatches to OX or SREX based on the number of routes in the parent solution... would this have much downside?

Not as far as I can tell. That'd help, especially for smaller instances. We'd need to take some care to navigate around the warnings and exceptions of standalone OX/SREX, but this should be possible. What to call such an operator, however, is probably a harder task than implementing it.

PR #426, opened 2024-01-24
Add name fields

### LINKED ISSUE #412: Name fields for vehicle types, clients, and depots

It's pretty hard to understand which `Client` object is which, since they do not have any easily identifiable information on them: just data. A free-form, optional `name` field would help a lot, since that could e.g. be used to track a name or address.

A simple `std::string const name` field on these objects. Since we do not really use any of the string's features, might also just be a `char *`.

N-Wouda:
This PR:

- [x] Closes #412.

### Benchmarks on VRPTW

This PR w/ `std::string` (cbd36cebf38cf29739b37a8486cccbe265e9b863; 10x seeds, 1h; job ID 6975160):
```
               MEAN      MIN       MAX   COV
      inst
obj   C1    41365.0  41340.2   41406.6   0.0
      C2    16191.6  16188.8   16194.8   0.0
      R1    47279.6  47203.5   47411.3   0.1
      R2    27774.8  27731.4   27801.8   0.1
      RC1   44246.9  44190.2   44283.4   0.1
      RC2   23269.9  23253.1   23282.5   0.0
iters C1    59861.9  49358.3   66563.5   7.6
      C2    97626.2  77936.6  108538.2   9.6
      R1    32440.1  25192.6   36297.7  12.3
      R2    48749.3  35693.8   55939.6  13.1
      RC1   36444.3  26475.0   40933.9  13.5
      RC2   49916.8  37278.9   56831.7  14.3
```

From #411:
```
                MEAN      MIN       MAX   COV
      inst
obj   C1     41364.4  41339.2   41406.3   0.0
      C2     16190.8  16188.5   16193.2   0.0
      R1     47258.6  47186.1   47302.3   0.1
      R2     27766.2  27727.2   27796.0   0.1
      RC1    44235.4  44204.8   44253.1   0.0
      RC2    23266.7  23252.5   23302.3   0.1
iters C1     63203.5  47433.9   70698.7  12.6
      C2    106604.0  80918.5  119611.2  12.6
      R1     36559.1  27936.8   41311.5  13.1
      R2     54925.2  42722.3   62756.7  12.3
      RC1    39750.0  30405.4   44036.5  11.7
      RC2    56193.6  43259.6   61123.7  12.0
```

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/cpp/ProblemData.cpp
+++ b/pyvrp/cpp/ProblemData.cpp
@@ -1,11 +1,29 @@
 #include "ProblemData.h"
 
+#include <cstring>
 #include <numeric>
 
 using pyvrp::Distance;
 using pyvrp::Duration;
 using pyvrp::Matrix;
 using pyvrp::ProblemData;
+
+namespace
+{
+// Small local helper for what is essentially strdup() from the C23 standard,
+// which my compiler does not (yet) have. See here for the actual recipe:
+// https://stackoverflow.com/a/252802/4316405 (modified to use new instead of
+// malloc).
+static char *duplicate(const char *src)
+{
+    char *dst = new char[std::strlen(src) + 1];  // space for src + null
+    if (!dst)
+        return dst;
+
+    std::strcpy(dst, src);
+    return dst;
+}
+}  // namespace
 
 ProblemData::Client::Client(Coordinate x,
                             Coordinate y,
@@ -16,7 +34,7 @@ ProblemData::Client::Client(Coordinate x,
                             Duration releaseTime,
                             Cost prize,
                             bool required,
-                            std::string name)
+                            char const *name)
     : x(x),
       y(y),
       demand(demand),
@@ -25,7 +43,7 @@ ProblemData::Client::Client(Coordinate x,
       twLate(twLate),
       releaseTime(releaseTime),
       prize(prize),
-      name(name),
+      name(duplicate(name)),
       required(required)
 {
     if (demand < 0)
@@ -44,6 +62,37 @@ ProblemData::Client::Client(Coordinate x,
         throw std::invalid_argument("prize must be >= 0.");
 }
 
+ProblemData::Client::Client(Client const &client)
+    : x(client.x),
+      y(client.y),
+      demand(client.demand),
+      serviceDuration(client.serviceDuration),
+      twEarly(client.twEarly),
+      twLate(client.twLate),
+      releaseTime(client.releaseTime),
+      prize(client.prize),
+      name(duplicate(client.name)),
+      required(client.required)
+{
+}
+
+ProblemData::Client::Client(Client &&client)
+    : x(client.x),
+      y(client.y),
+      demand(client.demand),
+      serviceDuration(client.serviceDuration),
+      twEarly(client.twEarly),
+      twLate(client.twLate),
+      releaseTime(client.releaseTime),
+      prize(client.prize),
+      name(client.name),  // we can steal
+      required(client.required)
+{
+    client.name = nullptr;  // stolen
+}
+
+ProblemData::Client::~Client() { delete[] name; }
+
 ProblemData::VehicleType::VehicleType(size_t numAvailable,
                                       Load capacity,
                                       size_t depot,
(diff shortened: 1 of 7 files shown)


pyvrp/cpp/ProblemData.h
```diff
@@ -94,8 +98,9 @@ class ProblemData
         Duration const twEarly;      // Earliest possible start of service
         Duration const twLate;       // Latest possible start of service
         Duration const releaseTime;  // Earliest possible time to leave depot
-        Cost const prize = 0;        // Prize for visiting this client
-        bool const required = true;  // Must client be in solution?
+        Cost const prize;            // Prize for visiting this client
+        std::string const name;      // Location name (for reference)
```
N-Wouda, line 102: A `std::string` is really big, so I'm running a benchmark to check if that matters. If it does, I'll use a `char *` instead and do a little memory allocation to make that work. We don't really _need_ a `std::string`, but it's convenient and avoids managing memory manually. But I'm not willing to sacrifice any performance for this.

N-Wouda:
Yeah that seems to matter a fair bit for overall performance, like 10% fewer iterations. I'll try the char pointer stuff tomorrow.

pyvrp/cpp/ProblemData.cpp
```diff
@@ -42,19 +62,52 @@ ProblemData::Client::Client(Coordinate x,
...
+      twLate(client.twLate),
+      releaseTime(client.releaseTime),
+      prize(client.prize),
+      name(client.name),  // we can steal
+      required(client.required)
+{
+    client.name = nullptr;  // stolen
+}
+
+ProblemData::Client::~Client() { delete[] name; }
```
N-Wouda, line 94: Because we have to manage these raw C strings ourselves, I need to set up a lot more: explicit memory allocation, copy/move constructors, and the destructor. I'm not setting up the copy or move assignment operators, because these objects are essentially constant after construction.

All this works, but it's a bit hard to test from Python because I do not know exactly where copy/move get used. It's pretty easy to reason about correctness, however, even without explicit tests for these constructors.

[same story applies for VehicleType]

N-Wouda:
Yeah, with `char const *` seems to be faster (higher iterations).

N-Wouda:
@leonlan could you have a look at this? Once this is in I'll continue to work on writing MDVRPTW solutions.

pyvrp/cpp/ProblemData.cpp
```diff
@@ -24,6 +40,7 @@ ProblemData::Client::Client(Coordinate x,
       twLate(twLate),
       releaseTime(releaseTime),
       prize(prize),
+      name(duplicate(name)),
```
leonlan, line 43: Is there a reason why the ordering of prize and name is reversed here?

pyvrp/cpp/ProblemData.cpp
```diff
@@ -42,19 +59,52 @@ ProblemData::Client::Client(Coordinate x,
...
-                                      std::optional<Duration> maxDuration)
+                                      std::optional<Duration> maxDuration,
+                                      char const *name)
     : numAvailable(numAvailable),
       depot(depot),
       capacity(capacity),
       fixedCost(fixedCost),
       twEarly(twEarly),
       twLate(twLate),
+      name(duplicate(name)),
```
leonlan, line 107: Same here, why is the order of name and max duration reversed?

leonlan (approved):

pyvrp/cpp/ProblemData.cpp
N-Wouda: [same holds for Client]

The fields are initialised in the order of their layout in the header file. That layout is:
```cpp
      [8] size_t const numAvailable;  // Available vehicles of this type
      [8] size_t const depot;         // Departure and return depot location
      [4] Load const capacity;        // This type's vehicle capacity
      [4] Cost const fixedCost;       // Fixed cost of using this vehicle type
      [8] std::optional<Duration> const twEarly;  // Start of shift
      [8] std::optional<Duration> const twLate;   // End of shift
      [8] char const *name;                       // Type name (for reference)
      [4] Duration const maxDuration;             // Maximum route duration
```
which is slightly different from argument order in the constructor, which we try to keep backwards compatible.

Each of the items here has a particular size: Cost, Load, Duration, etc. all boil down to `int`, which is 4 bytes on most of the systems we work with. A `size_t` and a pointer type are typically 8 bytes, as is `std::optional`. I've put these sizes in brackets above.

The size of each field determines [alignment requirements](https://en.cppreference.com/w/c/language/object#Alignment). The compiler needs to respect these, and will insert some artificial padding to make that happen. The VehicleType struct now has size `52` in pure data, but `56` in effective size due to a little padding at the end, after `maxDuration`. We can thus add another integer value effectively for 'free' when we need to track more data for future features - adding that to the end now will do the right thing.

Suppose we had done this instead:
```cpp
      [4] Duration const maxDuration;             // Maximum route duration
      [4 bytes padding to align name]
      [8] char const *name;                       // Type name (for reference)
``` 
Then the articial padding would have been 'hidden' between `maxDuration` and `name`, and our struct would have been bigger than it needs to be. 

Since this is all internal we get to change this around a bit whenever we need to. I sometimes tweak it a bit to shave off some bytes.

PR #427, opened 2024-01-25
Write MDVRPTW solutions

N-Wouda:
This PR:

- [x] Is part of #305.

<details>

**Notes**:

</details>


pyvrp/cpp/ProblemData.h
```diff
@@ -263,6 +263,16 @@ class ProblemData
...
 
+    /**
+     * Returns a list of all vehicle types in the problem instance.
+     *
+     * Returns
+     * -------
+     * List[VehicleType]
+     *     List of all vehicle types in the problem instance.
+     */
+    [[nodiscard]] std::vector<VehicleType> const &vehicleTypes() const;
```
N-Wouda, line 274: This was somewhat missing and I figured I'd add it in now.

docs/source/dev/benchmarking.rst
N-Wouda, line 1: @leonlan could you check if these instructions are correct for CVRP, VRPTW, and PC-VRPTW?

pyvrp/stop/TimedNoImprovement.py
N-Wouda, line 1: I don't think this was actually used, but I'll add a note to the change log to state the workaround (use `MultipleCriteria`).

pyvrp/stop/__init__.py
```diff
@@ -1,5 +1,5 @@
 from .MaxIterations import MaxIterations as MaxIterations
 from .MaxRuntime import MaxRuntime as MaxRuntime
+from .MultipleCriteria import MultipleCriteria as MultipleCriteria
```
N-Wouda, line 3: I'm open to suggestions for a better name!

pyvrp/stop/tests/test_MultipleCriteria.py
```diff
@@ -0,0 +1,49 @@
...
+    with assert_raises(ValueError):
+        MultipleCriteria([])
+
+
+def test_stops_if_zero_max_iterations():
+    """
+    Tests that the stopping criterion stops immediately if zero maximum
+    iterations are allowed.
+    """
+    stop = MultipleCriteria([NoImprovement(0), MaxRuntime(0.100)])
```
leonlan, line 21:
```suggestion
    stop = MultipleCriteria([MaxIterations(0), MaxRuntime(0.100)])
```

Suggestion: It's a bit confusing to write max iterations but then to use the no improvement criterion. I cannot see the direct benefit of using no improvement over max iterations so I suggest replacing it (here and in the other tests).

pyvrp/cli.py
```diff
@@ -60,13 +67,39 @@ def tabulate(headers: list[str], rows: np.ndarray) -> str:
...
+        type2vehicle = [
+            (int(vehicle) - 1 for vehicle in vehicle_type.name.split(","))
+            for vehicle_type in data.vehicle_types()
+        ]
+
+        routes = [f"Route {idx + 1}: " for idx in range(data.num_vehicles)]
+        for route in result.best.get_routes():
+            visits = map(str, route.visits())
+            vehicle = next(type2vehicle[route.vehicle_type()])
+            routes[vehicle] += " ".join(visits)
```
leonlan, line 89: Nice 👍

pyvrp/stop/__init__.py
leonlan: I like this one! I have tried before to come up with a name (for ALNS) but I think this one is clear enough. 

(Off-topic: In ALNS it should probably be called MultipleStoppingCriteria to avoid confusion with acceptance criteria.)

pyvrp/stop/tests/test_MultipleCriteria.py
N-Wouda: Fair. I copied these tests from the old `TimedNoImprovement`. I'll update this now.

PR #429, opened 2024-01-25
Enforce distance and duration matrix diagonal to 0

### LINKED ISSUE #428: Evaluation of distance/duration of empty route in local search operators
#### Describe the bug

In `Solution.cpp`, `Route` does not compute any statistics if `visits_.empty()`, hence the distance and duration will be 0. However, when moves that result in an empty route are evaluated in the local search, the distance (at least in `Exchange`) is computed as `distanceMatrix(depot, depot)`. If this value is not 0, it results in a discrepancy in the cost evaluation in local search vs the actual cost computed by the solution. This happens for example in `test_model_and_solve` in `test_Model.py`, when a relocate move to an empty route is evaluated.

#### Code sample

```bash
pytest pyvrp/tests/test_Model.py -k test_model_and_solve
```

When running this test, the following line will evaluate to a large value as the distance depot -> depot is `MAX_VALUE` as this is the default in `Model.py`. https://github.com/PyVRP/PyVRP/blob/94b3954aa2072a1ef259cafe0d3ad9eec0bdc333/pyvrp/cpp/search/Exchange.h#L93-L94 
This results in a large negative delta cost here which is incorrect:
https://github.com/PyVRP/PyVRP/blob/94b3954aa2072a1ef259cafe0d3ad9eec0bdc333/pyvrp/cpp/search/Exchange.h#L102

<details>

**Note**: Please be sure you are using the latest released version of `pyvrp`, or a recent build of `main`.
If your problem has been fixed in an unreleased version, you might be able to use `main` until a new release occurs.

**Note**: If you are using a released version, have you verified that the bug exists in the master branch of this repository?

</details>

If the issue has not been resolved, please file it in the issue tracker.
Make sure to include the version of Python you are using, and the version of `pyvrp` you run.
You can copy-paste the output of `import pyvrp; pyvrp.show_versions()` to achieve this.

#### Expected Output

I expect operators to evaluate moves and define the cost of an empty route consistently with the definition of cost in a `Solution`. Most reasonable, I'd say this means that we should enforce `dist(depot,depot) = 0`. I suggest to enforce this for all locations just in case.

wouterkool:
This PR:

- [x] Closes #428.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/Model.py
+++ b/pyvrp/Model.py
@@ -241,6 +241,8 @@ class Model:
         # not set below are never traversed.
         distances = np.full((len(locs), len(locs)), MAX_VALUE, dtype=int)
         durations = np.full((len(locs), len(locs)), MAX_VALUE, dtype=int)
+        np.fill_diagonal(distances, 0)
+        np.fill_diagonal(durations, 0)
 
         for edge in self._edges:
             frm = loc2idx[id(edge.frm)]
--- a/pyvrp/Model.py
+++ b/pyvrp/Model.py
@@ -180,6 +180,9 @@ class Model:
             """
             warn(msg, ScalingWarning)
 
+        if frm == to:
+            raise ValueError("Cannot have an edge from a node to itself.")
+
         edge = Edge(frm, to, distance, duration)
         self._edges.append(edge)
         return edge
--- a/pyvrp/Model.py
+++ b/pyvrp/Model.py
@@ -180,8 +180,8 @@ class Model:
             """
             warn(msg, ScalingWarning)
 
-        if frm == to:
-            raise ValueError("Cannot have an edge from a node to itself.")
+        if frm == to and (distance != 0 or duration != 0):
+            raise ValueError("A self-loop edge must have 0 distance/duration.")
 
         edge = Edge(frm, to, distance, duration)
         self._edges.append(edge)
--- a/pyvrp/cpp/ProblemData.cpp
+++ b/pyvrp/cpp/ProblemData.cpp
@@ -163,6 +163,16 @@ ProblemData::ProblemData(std::vector<Client> const &clients,
             throw std::invalid_argument("Depot release time must be 0.");
     }
 
+    // Check that diagonal of distance and duration matrix is equal to 0.
+    for (size_t i = 0; i < numLocations(); ++i)
+    {
+        if (dist_(i, i) != 0)
+            throw std::invalid_argument("Distance matrix diagonal must be 0.");
+
+        if (dur_(i, i) != 0)
+            throw std::invalid_argument("Duration matrix diagonal must be 0.");
+    }
+
     for (auto const &client : clients_)
     {
         centroid_.first += static_cast<double>(client.x) / numClients();
--- a/pyvrp/search/tests/test_LocalSearch.py
+++ b/pyvrp/search/tests/test_LocalSearch.py
@@ -430,7 +430,7 @@ def test_local_search_does_not_remove_required_clients():
         ],
         depots=[Client(x=0, y=0)],
         vehicle_types=[VehicleType(1, capacity=50)],
-        distance_matrix=np.full((3, 3), fill_value=10),
+        distance_matrix=np.where(np.eye(3), 0, 10),  # All 10 except diagonal
         duration_matrix=np.zeros((3, 3), dtype=int),
     )
 
--- a/pyvrp/tests/test_ProblemData.py
+++ b/pyvrp/tests/test_ProblemData.py
@@ -261,7 +261,7 @@ def test_problem_data_replace_with_changes():
     new = original.replace(
         clients=[Client(x=1, y=1)],
         vehicle_types=[VehicleType(3, 4), VehicleType(5, 6)],
-        distance_matrix=np.ones((2, 2), dtype=int),
+        distance_matrix=np.where(np.eye(2), 0, 2),  # All 2 except diagonal,
     )
 
     assert_(new is not original)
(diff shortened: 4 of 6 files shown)


pyvrp/Model.py
```diff
@@ -180,6 +180,9 @@ def add_edge(
             """
             warn(msg, ScalingWarning)
 
+        if frm == to and (distance != 0 or duration != 0):
```
wouterkool, line 183: I considered disallowing self-loops at all, however this way we avoid the need for an if statement when adding all edges from a matrix using a nested for-loop, as the self-loop can be added as long as it's 0 valued.

pyvrp/tests/test_ProblemData.py
```diff
@@ -233,7 +261,7 @@ def test_problem_data_replace_with_changes():
     new = original.replace(
         clients=[Client(x=1, y=1)],
         vehicle_types=[VehicleType(3, 4), VehicleType(5, 6)],
-        distance_matrix=np.ones((2, 2), dtype=int),
+        distance_matrix=np.where(np.eye(2), 0, 2),  # All 2 except diagonal,
```
N-Wouda, line 264:
```suggestion
        distance_matrix=np.where(np.eye(2), 0, 2),  # All 2 except diagonal
```
(since almost all the other changes of a similar form go undocumented, these comments might also just be superfluous and removed)

pyvrp/cpp/ProblemData.cpp
```diff
@@ -163,6 +163,16 @@ ProblemData::ProblemData(std::vector<Client> const &clients,
...
 
+    // Check that diagonal of distance and duration matrix is equal to 0.
+    for (size_t i = 0; i < numLocations(); ++i)
+    {
+        if (dist_(i, i) != 0)
+            throw std::invalid_argument("Distance matrix diagonal must be 0.");
+
+        if (dur_(i, i) != 0)
+            throw std::invalid_argument("Duration matrix diagonal must be 0.");
+    }
```
N-Wouda, line 174:
```suggestion
    for (size_t idx = 0; i != numLocations(); ++idx)
    {
        if (dist_(idx, idx) != 0)
            throw std::invalid_argument("Distance matrix diagonal must be 0.");

        if (dur_(idx, idx) != 0)
            throw std::invalid_argument("Duration matrix diagonal must be 0.");
    }
```

pyvrp/Model.py
```diff
@@ -180,6 +180,9 @@ def add_edge(
...
+            raise ValueError("A self-loop edge must have 0 distance/duration.")
```
N-Wouda, line 184:
```suggestion
            raise ValueError("A self loop edge must have 0 distance and duration.")
```
Nit: here it's self-loop, in all other places self loop.

Can we add this case to the error condition in the docstring? Something like
> When either distance or duration is a negative value, or when self loops have nonzero distance or duration values.

It'd then also make sense to move this if/raise block up to just below the other error condition, so all error conditions are grouped together.

pyvrp/tests/test_ProblemData.py
```diff
@@ -178,6 +178,34 @@ def test_problem_data_raises_when_incorrect_matrix_dimensions(matrix):
...
+        )
+
+    # duration matrix with non-zero value on the diagonal
+    distance_matrix = np.zeros((2, 2), dtype=int)
+    duration_matrix = np.array([[0, 0], [0, 1]], dtype=int)
+
+    with assert_raises(ValueError):
+        ProblemData(
+            clients, depots, vehicle_types, distance_matrix, duration_matrix
+        )
```
N-Wouda, line 206: This can be halved using `pytest.mark.parametrize` to pass in the matrices. That'd make the separation between the data causing the issue and the set-up to get to where we can test it a bit clearer.

pyvrp/tests/test_ProblemData.py
```diff
@@ -310,6 +338,8 @@ def test_matrix_access():
 
     dist_mat = gen.integers(500, size=(size, size))
     dur_mat = gen.integers(500, size=(size, size))
+    np.fill_diagonal(dist_mat, 0)
+    np.fill_diagonal(dur_mat, 0)
     clients = [
```
N-Wouda, line 343:
```suggestion
    np.fill_diagonal(dist_mat, 0)
    np.fill_diagonal(dur_mat, 0)

    clients = [
```
nit

pyvrp/cpp/ProblemData.cpp
N-Wouda: I'd move this up a little bit, to just below the matrix dimension checks. Then all "matrix stuff" is nicely grouped together.

pyvrp/search/tests/test_LocalSearch.py
```diff
@@ -430,7 +430,7 @@ def test_local_search_does_not_remove_required_clients():
         ],
         depots=[Client(x=0, y=0)],
         vehicle_types=[VehicleType(1, capacity=50)],
-        distance_matrix=np.full((3, 3), fill_value=10),
+        distance_matrix=np.where(np.eye(3), 0, 10),  # All 10 except diagonal
```
N-Wouda, line 433: Nit. Might be superfluous comment. I think this is pretty clear and many other places that use this form also don't have a comment explaining what's going on.

N-Wouda (approved):
Looks good to me; I only have a small remark about the documentation of an error condition, and some phrasing/ordering stuff.

One of the test cases in `test_ProblemData.py` can be halved or so using the parametrize function. It'd then also be easy to check the last case where both matrices are 'wrong' - that's just one more row of data to feed into parametrize.

N-Wouda:
I'm implementing my very minor feedback to get this PR merged, because it fixes a bug that'd be great to have fixed before we release v0.7.0. Because I cannot push to the ORTEC fork, I have implemented my comments in #436.

PR #430, opened 2024-01-25
Implement consistency checks in local search

wouterkool:
This PR adds a build option for optional internal consistency checks in local search. This helps catching bugs when introducing new features for which local search operators need to be adapted.
With the option enabled:
- Check that cost after applying local search move equals cost before + delta cost
- Check that solution is loaded correctly
- Add compiler checks for undefined behavior when using consistency checks

<details>

**Notes**:

</details>

### DIFF
--- a/build_extensions.py
+++ b/build_extensions.py
@@ -31,6 +31,11 @@ def parse_args():
         help="Double is more precise, integer faster. Defaults to 'integer'.",
     )
     parser.add_argument(
+        "--consistency_checks",
+        action="store_true",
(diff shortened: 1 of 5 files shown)


wouterkool:
@N-Wouda I can imagine we don't want to do this by default for debug builds so I made it optional using a separate flag, but I'm open to suggestions. I think having the option to run with consistency checks is very useful (we've had some bugs that would have been found earlier using these checks).

N-Wouda:
Shall we now close this PR?

PR #432, opened 2024-01-27
Update docs about MDVRPTW

### LINKED ISSUE #419: Add a more advanced example?

We support VRPs with quite a few different attributes, but our examples in the documentation do not show how to use all of those. We currently only show how to solve CVRPs and VRPTWs. Perhaps we should add another small example that solves a problem with prizes, different vehicle types, multiple depots, etc.

Another example showcasing more features.

N-Wouda:
This PR:

- [x] Closes #419.

<details>

**Notes**:

</details>


N-Wouda:
@leonlan could you have a look at this and let me know what you think?

leonlan (approved):
LGTM. I think this works for now but I think we can do better for v0.8.0 when we add more VRP variants. I'll make an issue for that to discuss this further.

PR #433, opened 2024-01-27
Build maintenance

N-Wouda:
This PR updates the CD a bit, and removes some superfluous settings from `pyproject.toml`.

<details>

**Notes**:

</details>


N-Wouda:
Updating the workflows is always an adventure 😂

N-Wouda:
We need to wait for a new poetry release, because it's currently breaking our build (basically, from what I have gathered: poetry wants platformdirs in [3, 4], while virtualenv installs something >4). The next release of poetry will be OK with platformdirs in [3, 5], so that should fix it.

N-Wouda:
Just checked poetry's issue tracker, and I think that release will happen in a few days. I'll keep this PR around until then.

N-Wouda:
Still waiting for Poetry on this.

N-Wouda:
Note to self: update OS to latest in CD.

N-Wouda:
RTD is broken but that's unrelated to this PR (also happens in #487).

PR #436, opened 2024-01-29
Enforce zero matrix diagonal

### LINKED ISSUE #428: Evaluation of distance/duration of empty route in local search operators

N-Wouda:
This PR:

- [x] Closes #428.

(pushed in new PR because I do not have write access to the ORTEC fork)

<details>

**Notes**:

</details>


N-Wouda:
Thanks @wouterkool!

PR #448, opened 2024-01-31
Update pyvrp.repair operator interface to use lists of routes, not solutions

### LINKED ISSUE #389: Allow empty routes in Solution
I would like `Solution -> Solution`, but we need the ability to at least increase the number of routes somehow. Else we're unnecessarily limiting any LNS to never increase the number of routes, which seems weird to me (if the vehicles are available, why not?).

But `Solution` currently isn't equipped to do that. So we'd have to change that first, and then settle on a single interface for the repair operators.

_Originally posted by @N-Wouda in https://github.com/PyVRP/PyVRP/issues/386#issuecomment-1714551995_

N-Wouda:
This PR:

- [x] Closes #389.

<details>

**Notes**:

</details>

### DIFF
--- a/docs/source/api/repair.rst
+++ b/docs/source/api/repair.rst
@@ -6,7 +6,8 @@ Repair operators
 ================
 
 The :mod:`pyvrp.repair` module provides operators that are responsible for repairing a :class:`~pyvrp._pyvrp.Solution` after destruction in a large neighbourhood search (LNS) setting.
-TODO
+These operators take a given list of routes which represent the solution to repair, and insert the unplanned clients into these routes.
+To allow fine-grained control over the number of routes and vehicles used, no new routes are created: clients are only inserted into the given routes.
 
 .. automodule:: pyvrp.repair._repair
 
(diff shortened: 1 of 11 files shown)


N-Wouda:
This PR is either part of #389, or it closes it, depending on whether we want to keep that issue open. I'm leaning close.

leonlan (approved):
LGTM

wouterkool (approved):
LGTM!

Only one remark but not necessarily related to this change: is `repair` the best terminology? It suggests the solution must be damaged/destroyed somehow, whereas I wonder if in general `greedy` and `nearest_route_insert` could also be used to construct a solution/insert new customers into existing solution, etc. If you agree, maybe consider renaming the module to 'construction' or something like that?

N-Wouda:
> is `repair` the best terminology?

The main reason these operators exist is #307. There is a long tradition in LNS to call this step _repair_. That the operators in `pyvrp.repair` can do more things than just LNS is nice, but I'm a little hesitant to change the naming of this module on the basis of it being a big more general than just LNS.

PR #462, opened 2024-02-09
Add asserts that check for overflow

N-Wouda:
This PR adds checks that detect overflow when running debug builds. This will be useful when we start defaulting values to `INT_MAX` in some places.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/constants.py
+++ b/pyvrp/constants.py
@@ -1,2 +1,11 @@
-MAX_USER_VALUE = 2 << 25
-MAX_VALUE = 2 << 28
+MAX_USER_VALUE = 2 << 22
+"""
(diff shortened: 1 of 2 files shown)


N-Wouda:
`__builtin_add_overflow` is something all big compilers support, but I am unsure about where it comes from. GCC has it available without needing a specific header, but that might not be true across all compilers. Let's see what clang thinks, for example.

pyvrp/constants.py
N-Wouda, line 1: Turns out that the defaults here were already too big, and causing overflow. I've lowered them a bit. The biggest user value we take is now `2 << 22`, or `8_388_608`. That seems reasonable for most applications.

leonlan (approved):

Issue #461, opened 2024-02-09
Release time and edge durations result in time warp and infeasible solutions


Stanpol:
#### Describe the bug

Setting `release time` for clients or `duration` for edges results in routes having `has_time_warp()==True` and infeasible solutions.

#### Code sample
This is a working example. Take 1000 random clients, all have their release time at 0 and edge durations at 0.
```python
import numpy as np
from pyvrp import Model
from pyvrp.stop import MaxRuntime
from pyvrp.plotting import plot_solution

np.random.seed(0)

n = 1000
x = (np.random.rand(n) * 1000).astype(int)
y = (np.random.rand(n) * 1000).astype(int)
demand = [0] + [1]*999

m = Model()
m.add_vehicle_type(16, capacity=75)
depot = m.add_depot(x=x[0], y=y[0])
clients = [
    m.add_client(x=x[idx], y=y[idx], demand=demand[idx], release_time=0) # release time is 0
    for idx in range(1, len(x))
]

locations = [depot, *clients]
for frm in locations:
    for to in locations:
        duration = distance = abs(frm.x - to.x) + abs(frm.y - to.y)  # Manhattan
        m.add_edge(frm, to, distance=distance, duration=0) # duration is 0

res = m.solve(stop=MaxRuntime(1))  # one second
print(res)
plot_solution(res.best, m.data())
for id, r in enumerate(res.best.get_routes()):
    print(f"Route {id}: Release time {r.release_time()}, "
          f"distance {r.distance()}, "
          f"duration {r.duration()}, "
          f"end time {r.end_time()}, "
          f"vehicle type {r.vehicle_type()}, "
          f"has excess load {r.has_excess_load()}, "
          f"has time warp {r.has_time_warp()}.")
```
A feasible solution was found in 1 second, all the distances are below 5000:
![image](https://github.com/PyVRP/PyVRP/assets/708533/09f68dcc-81c4-4063-9b5f-c71cda0717af)

Let's modify the code. First set the duration equal to distance. And give it a minute to solve.
```python
import numpy as np
from pyvrp import Model
from pyvrp.stop import MaxRuntime
from pyvrp.plotting import plot_solution

np.random.seed(0)

n = 1000
x = (np.random.rand(n) * 1000).astype(int)
y = (np.random.rand(n) * 1000).astype(int)
demand = [0] + [1]*999

m = Model()
m.add_vehicle_type(16, capacity=75)
depot = m.add_depot(x=x[0], y=y[0])
[...]

PR #463, opened 2024-02-09
Make [0, INT_MAX] the default time windows

N-Wouda:
This PR hopefully makes the issue of #461 a bit easier to navigate.

<details>

**Notes**:

</details>


pyvrp/_pyvrp.pyi
```diff
@@ -41,7 +41,7 @@ class Client:
         demand: int = 0,
         service_duration: int = 0,
         tw_early: int = 0,
-        tw_late: int = 0,
+        tw_late: int = ...,
```
N-Wouda, line 44: https://docs.astral.sh/ruff/rules/argument-default-in-stub/ - Ruff disagrees with using `np.iinfo(np.int32).max` here as the default.

pyvrp/read.py
```diff
@@ -149,11 +149,13 @@ def read(
         durations = distances
         time_windows: np.ndarray = round_func(instance["time_window"])
     else:
-        # No time window data. So the time window component is not relevant,
-        # and we set all time-related fields to zero.
+        # No time window data. So the time window component is not relevant.
         durations = np.zeros_like(distances)
         service_times = np.zeros(dimension, dtype=int)
```
leonlan, line 154: Setting durations/service times to zero when there are no time windows no longer makes sense with vehicle max duration and shifts. We don't have any instances where it can go wrong, but it'd be convenient if we changed this now. Can we:
- always set `distances = durations` (VRPLIB-style)
- and leave `service_times` unmodified here?

pyvrp/cpp/TimeWindowSegment.h
```diff
@@ -131,10 +131,13 @@ TimeWindowSegment::merge(Matrix<Duration> const &durationMatrix,
     Dur const edgeDuration = durationMatrix(idxLast_, other.idxFirst_);
     Dur const atOther = duration_ - timeWarp_ + edgeDuration;
 
-    // Wait duration increases if we arrive at the other's first client before
-    // opening, whereas time warp increases if we arrive there after closing.
-    Dur const diffWait = std::max<Dur>(other.twEarly_ - atOther - twLate_, 0);
+    // Time warp increases if we arrive at the other's first client after its
+    // time window closes, whereas wait duration increases if we arrive there
+    // before opening.
     Dur const diffTw = std::max<Dur>(twEarly_ + atOther - other.twLate_, 0);
```
leonlan, line 137: Question: why no ternary for `diffTw`? Here's also a subtraction by `other.twLate` that could cause underflow.

leonlan: Related: in line 174 we also have an expression which subtract `twLate`:
```
         + std::max<Duration>(releaseTime_ - twLate_, 0)
```

pyvrp/Model.py
```diff
@@ -113,7 +113,7 @@ def add_client(
         demand: int = 0,
         service_duration: int = 0,
         tw_early: int = 0,
-        tw_late: int = 0,
```
leonlan, line 116: I don't understand why we initialize `tw_late` and others with `INT_MAX` instead of `MAX_USER_VALUE`. More generally, why do we not check and warn about all user data to be smaller than `MAX_USER_VALUE`? Is that only an issue for duration/distances?

leonlan (approved):

pyvrp/cpp/TimeWindowSegment.h
N-Wouda: We have:
- `twEarly_ >= 0`
- `atOther >= 0`
- `releaseTime_ >= 0`

And then a subtraction by everything up to and including `INT_MAX` is well-defined. So these cases are not problematic.

pyvrp/Model.py
N-Wouda: Don't have strong opinions on this. I think it was mostly meant as a convenience for the user when setting duration/distance data. We might want to build something like this in more generally when constructing a `ProblemData` instance, but that's not very high on my TODO list.

N-Wouda: I'm also OK with not warning about anything, and removing the warnings we currently have, if we can make things work nicely with potential overflow.

PR #467, opened 2024-02-09
Move tests out of the source tree

### LINKED ISSUE #373: Out of tree tests?
Right now, we have our tests directly next to the source code they test. That's nice, but does mean we also need to put our fixtures in the source tree. We might want to switch to a separate `tests/` tree, such that testing and source code are completely separated. 

A particular case that motivates this is the `pyvrp/conftest.py` file, which should not really be there. Not super high-priority, but we might want to do this at some point.

N-Wouda:
I think it's a good idea to do this now. There is a bit of test infrastructure that overlaps with the actual package, and I would like to separate the two a bit better. The new structure is as follows:
```
<root>/
  pyvrp/
    <actual package>
  tests/
    <tests for package>
```

This PR:

- [x] Closes #373.

<details>

**Notes**:

</details>


tests/conftest.py
N-Wouda, line 1: This file used to count for coverage, but is only used for tests, so the change in coverage due to this is not worrisome.

pyproject.toml
```diff
@@ -26,7 +25,8 @@ exclude = [
 ]
 packages = [
     { include = "pyvrp" },
-    { include = "docs" },
+    { include = "tests", format = "sdist" },
+    { include = "docs", format = "sdist" },
```
N-Wouda, line 29: These should both only be in the source distribution, not the final wheel.

pyproject.toml
```diff
@@ -26,7 +34,6 @@ exclude = [
 ]
 packages = [
```
N-Wouda, line 35: I've been doing a bit of reading on how poetry likes to see things set up, and as it turns out `docs/` and `tests/` should not be in this attribute, but rather just be made part of the `include` attribute :).

leonlan (approved):

Issue #420, opened 2024-01-13
Add copy/move constructors and assignment operators to classes that need them


N-Wouda:
I don't think we have been very consistent with this, so it's good to check and where they are missing either add or explicitly delete these operators.

PR #471, opened 2024-02-12
Simultaneous pickup and delivery

N-Wouda:
This PR adds support for mixed backhaul, that is, clients that both demand from, and return supply back to, the depot. As such, this PR:

- [x] Closes #434, part of #331.
- [x] Adds and passes relevant tests.
- [x] Has well-formatted code and documentation.

### Benchmark

This PR (fb4a881a8b84461d6d59dcbf3cf0fe52ca1d873f; job ID 7252913; 1h VRPTW):
```
                MEAN      MIN       MAX  COV
      inst
obj   C1     41363.3  41339.0   41406.7  0.0
      C2     16191.2  16188.8   16193.2  0.0
      R1     47258.3  47189.3   47316.4  0.1
      R2     27770.5  27728.7   27793.1  0.1
      RC1    44235.6  44186.7   44256.4  0.0
      RC2    23265.6  23252.8   23273.3  0.0
iters C1     65254.3  62045.2   69890.7  3.7
      C2    103122.0  97443.7  115895.7  5.3
      R1     35679.7  33392.1   38409.3  4.5
      R2     52035.2  49347.3   55149.7  3.7
      RC1    39353.9  37277.7   41681.6  3.9
      RC2    54220.6  50831.1   57578.4  3.9
```

Current main (a21c07f5c7d19e8692b9bd9ca49d5f581c783b2d; job ID 7237889; 1h VRPTW):
```
                MEAN      MIN       MAX   COV
      inst
obj   C1     41364.6  41339.1   41406.6   0.0
      C2     16191.1  16188.8   16193.2   0.0
      R1     47267.9  47189.9   47322.0   0.1
      R2     27772.3  27732.0   27797.8   0.1
      RC1    44243.1  44214.2   44295.8   0.1
      RC2    23269.1  23252.8   23301.7   0.1
iters C1     61378.4  43211.8   71776.5  14.8
      C2    102347.7  74514.4  116911.4  13.7
      R1     34207.2  24470.1   39459.3  14.2
      R2     49770.3  37928.6   60922.6  13.3
      RC1    37750.6  27831.0   47317.1  15.0
      RC2    52670.3  37043.1   59174.0  13.2
```

<details>

**Notes**:

- It is essential that you add a test when making code changes.
  This keeps the code coverage level up, and helps ensure the changes work as intended.
- When fixing a bug, you must add a test that would produce the bug in the master branch, and then show that it is fixed with the new code. 
[...]

PR #473, opened 2024-02-12
Proxy for segments of route statistics

### LINKED ISSUE #472: Simplify writing operators
In #471 I'm adding a new concatenation scheme for load segments, since return shipments make simple load calculations a bit harder. As a consequence, I find myself writing the whole of `Route::tws`, `Route::twsBetween`, `Route::twsBefore`, and `Route::twsAfter` again, but now for the load segment. That is annoying and hard to read. This issue serves to track the problem, and propose a solution. The solution is a bit more difficult technically, but should make it easier to write operators without as much boilerplate.

Consider the following case from #471:
```cpp
        auto const tws = TWS::merge(data.durationMatrix(),
                                    R->twsBefore(U->idx() - 1),
                                    R->twsAfter(U->idx() + 1));

        auto const ls = LoadSegment::merge(R->lsBefore(U->idx() - 1),
                                           R->lsAfter(U->idx() + 1));
```
This computes the `TimeWindowSegment` and `LoadSegment` assoicated with the route where `U` is removed. Such computations happen all over the local search - this is very common.

Would it not be much nicer if instead we could write the following:
```cpp
        auto const tws = TWS::merge(data.durationMatrix(),
                                    R->before(U->idx() - 1),
                                    R->after(U->idx() + 1));

        auto const ls = LoadSegment::merge(R->before(U->idx() - 1),
                                           R->after(U->idx() + 1));
```
and have that automatically do the right thing? I think that's possible if we set up generic `before`, `between`, `at`, and `after` members and have those return a proxy class that can be cast to the right type.

N-Wouda:
This PR implements the proxy method described in #472, and thus:

- [x] Closes #472, part of #420.
- [x] Has a benchmark to show there is no negative performance impact due to this change.

### Benchmark

This PR (86231e476f7abfa9669de655104923fca351ffc3; job ID 7237889; 1h VRPTW):
```
                MEAN      MIN       MAX   COV
      inst
obj   C1     41364.6  41339.1   41406.6   0.0
      C2     16191.1  16188.8   16193.2   0.0
      R1     47267.9  47189.9   47322.0   0.1
      R2     27772.3  27732.0   27797.8   0.1
      RC1    44243.1  44214.2   44295.8   0.1
      RC2    23269.1  23252.8   23301.7   0.1
iters C1     61378.4  43211.8   71776.5  14.8
      C2    102347.7  74514.4  116911.4  13.7
      R1     34207.2  24470.1   39459.3  14.2
      R2     49770.3  37928.6   60922.6  13.3
      RC1    37750.6  27831.0   47317.1  15.0
      RC2    52670.3  37043.1   59174.0  13.2
```
Current main (89dce334a4a650a58d68354bea5e0497d97383dc; job ID 7196352; 1h VRPTW):
```
               MEAN      MIN       MAX   COV
      inst
obj   C1    41367.6  41338.9   41407.1   0.0
      C2    16190.9  16188.0   16193.3   0.0
      R1    47264.6  47184.2   47364.6   0.1
      R2    27772.8  27737.4   27810.3   0.1
      RC1   44242.7  44189.9   44298.6   0.1
      RC2   23268.7  23252.8   23281.4   0.0
iters C1    60231.5  42535.5   65310.2  12.2
      C2    99196.3  70568.1  113006.6  13.2
      R1    34476.6  23499.1   40098.5  14.7
      R2    50044.2  35462.8   59678.6  14.9
      RC1   37444.6  25670.3   44032.3  14.5
      RC2   52624.6  35604.3   60778.3  15.2
```
So essentially no difference.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/cpp/search/Route.h
+++ b/pyvrp/cpp/search/Route.h
@@ -81,23 +81,54 @@ private:
         NodeStats(TimeWindowSegment const &tws);
     };
 
-    // TODO template different cases?
-    class Proxy
+    class ProxyAt
+    {
+        Route const *route;
+        size_t const idx;
+
+    public:
+        inline ProxyAt(Route const &route, size_t idx);
+        inline operator TimeWindowSegment const &() const;
+    };
+
+    class ProxyAfter
+    {
+        Route const *route;
+        size_t const start;
+
+    public:
+        inline ProxyAfter(Route const &route, size_t start);
+        inline operator TimeWindowSegment const &() const;
+    };
+
+    class ProxyBefore
+    {
+        Route const *route;
+        size_t const end;
+
+    public:
+        inline ProxyBefore(Route const &route, size_t end);
+        inline operator TimeWindowSegment const &() const;
+    };
+
+    class ProxyBetween
     {
         Route const *route;
         size_t const start;
         size_t const end;
 
     public:
-        inline Proxy(Route const &route, size_t start, size_t end);
+        inline ProxyBetween(Route const &route, size_t start, size_t end);
 
-        // TODO specialise with const & for known data.
         inline operator TimeWindowSegment() const;
         inline operator Load() const;
         inline operator Distance() const;
     };
 
-    friend class Proxy;
+    friend class ProxyAt;
+    friend class ProxyAfter;
+    friend class ProxyBefore;
+    friend class ProxyBetween;
 
     ProblemData const &data;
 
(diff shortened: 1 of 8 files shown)


pyvrp/cpp/search/bindings.cpp
N-Wouda, line 1: I'm keeping the old names in the bindings for backwards compatibility, and since we would have to use explicit names here anyway as Python probably does not understand the proxy object.

pyvrp/cpp/search/Route.h
```diff
@@ -81,6 +81,24 @@ class Route
         NodeStats(TimeWindowSegment const &tws);
     };
 
+    // TODO template different cases?
+    class Proxy
```
N-Wouda, line 85: This object is returned by calls to `Route::at`, `Route::between`, `Route::after`, and `Route::before`. It can then be casted to the type most appropriate for the call site - currently, one of `TWS`, `Load`, or `Distance`. This happens implicitly, making the `Proxy` a completely transparent implementation detail.

N-Wouda: (I'll probably move this into its own header if we continue with this, because I need to template it for several special cases - and then it gets big fast)

N-Wouda:
I'm going to try to finish this tomorrow so I can use it to simplify #471.

N-Wouda:
This is a few percent slower than current main. So I'm first going to try and make it so that this does not cost us any performance, and then, once we get there, work towards something a bit cleaner.

pyvrp/cpp/search/Route.h
N-Wouda, line 1: I don't know how to do this much simpler. I tinkered with templates a bit to tidy it up, but that ends up with about as much code that's harder to understand. I'm inclined to leave it as-is, and maybe return at some point in the future if it really gets out of hand.

N-Wouda:
@leonlan assuming the benchmark above works out, could you review this?

leonlan (approved):

PR #474, opened 2024-02-13
Remove unnecessary inlines

### LINKED ISSUE #328: Revisit our use of `inline`
We `inline` a lot of things, citing performance. I'm sure some of those are needed, but others are probably over the top. We should experiment a bit to see if we can remove more of these inlines.

N-Wouda:
This PR:

- [x] Closes #328.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/cpp/CostEvaluator.cpp
+++ b/pyvrp/cpp/CostEvaluator.cpp
@@ -3,6 +3,25 @@
 using pyvrp::Cost;
 using pyvrp::CostEvaluator;
 
+Cost CostEvaluator::loadPenalty(Load excessLoad) const
+{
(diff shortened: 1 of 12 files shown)


N-Wouda:
This first commit removes all inlines from the codebase, and moves the relevant function definitions to the cpp files. I'm now going to check with the profiler which ones we should really bring back.

N-Wouda:
Alright I think this should cover most of it. I'm going to run a benchmark to see how much this matters.

Benchmark (de9d52d90e308cf4daf3945220a22a4c89987127; job ID 7193630; 1h VRPTW):
```
RUNNING
```

N-Wouda:
I don't see any meaningful difference in benchmark performance, so I'm merging this in.

PR #477, opened 2024-02-19
New README badges

leonlan:
This PR replaces the README badges with ones from www.shields.io. Those badges have style options that have a more modern look and fit our logo better.


N-Wouda:
These look nice! 🚀

PR #481, opened 2024-02-21
Separate client and depot structs

### LINKED ISSUE #480: Separate Client and Depot structs
Right now we use a `Client` struct for both clients and depots. Clients have many more attributes than depots, and the `Client` struct supports this. As a consequence, we have to check in the `ProblemData` constructor that these attributes are not set for depots, which is (a) odd, and (b) confusing because the name `Client` does not make much sense for depots.

I suggest we separate the two structs. There'll be a little duplication, but that should be OK. We can probably still use the `ProblemData.location` member if we return a `std::variant<Client, Depot>`. We should benchmark to determine if that's fast enough.

N-Wouda:
This PR:

- [x] Closes #480.

### Benchmark

This PR (93b69a410c3801bf5e0aa53b66462a36f2f86c61; job ID 7283174; 1h VRPTW):
```
                MEAN      MIN       MAX   COV
      inst
obj   C1     41366.1  41338.7   41406.9   0.0
      C2     16191.2  16188.8   16193.2   0.0
      R1     47265.4  47189.1   47340.5   0.1
      R2     27771.1  27731.8   27799.4   0.1
      RC1    44241.0  44216.9   44266.5   0.0
      RC2    23270.2  23251.0   23305.7   0.1
iters C1     62083.6  40398.0   72053.5  14.5
      C2    100938.7  69613.3  117967.6  13.2
      R1     34600.1  22766.4   43833.9  16.3
      R2     50151.1  35937.7   56638.0  13.2
      RC1    38454.9  25687.2   46378.8  14.5
      RC2    51770.7  35407.6   61539.9  13.7
```

Current main (3e4ad95ca4a8284eb20a43214b726e330ba925fe; job ID 7252913; 1h VRPTW):
```
                MEAN      MIN       MAX  COV
      inst
obj   C1     41363.3  41339.0   41406.7  0.0
      C2     16191.2  16188.8   16193.2  0.0
      R1     47258.3  47189.3   47316.4  0.1
      R2     27770.5  27728.7   27793.1  0.1
      RC1    44235.6  44186.7   44256.4  0.0
      RC2    23265.6  23252.8   23273.3  0.0
iters C1     65254.3  62045.2   69890.7  3.7
      C2    103122.0  97443.7  115895.7  5.3
      R1     35679.7  33392.1   38409.3  4.5
      R2     52035.2  49347.3   55149.7  3.7
      RC1    39353.9  37277.7   41681.6  3.9
      RC2    54220.6  50831.1   57578.4  3.9
```

<details>

**Notes**:

</details>


N-Wouda:
This is functional, but needs more work before it's ready. And then a performance check.

N-Wouda:
> This is functional, but needs more work before it's ready. And then a performance check.

Apparently the CI thinks otherwise. This all runs on my machine so it'll be tricky to debug.

pyvrp/read.py
```diff
@@ -100,15 +100,12 @@ def read(
 
     instance = vrplib.read_instance(where, instance_format=instance_format)
 
-    # A priori checks
-    if "dimension" in instance:
-        dimension: int = instance["dimension"]
-    else:
-        if "demand" not in instance:
-            raise ValueError("File should either contain dimension or demands")
-        dimension = len(instance["demand"])
```
N-Wouda, line 109: There wasn't really a need for this dance to get the problem size: if it's available, we get it from `dimension`, and if it's not, we get it from the number of rows in the dist/dur matrices.

pyvrp/cpp/ProblemData.cpp
```diff
@@ -104,6 +104,41 @@ ProblemData::Client::Client(Client &&client)
...
+}
+
+ProblemData::Depot::Depot(Depot &&depot)
+    : x(depot.x),
+      y(depot.y),
+      twEarly(depot.twEarly),
+      twLate(depot.twLate),
+      name(depot.name)  // we can steal
+{
+    depot.name = nullptr;  // stolen
```
N-Wouda, line 137: There is no test for this since it's fundamentally a C++ mechanic that isn't really available to test from Python, but it should be clear from reading it that this is indeed correct.

Issue #434, opened 2024-01-28
Simultaneous pickup and delivery


N-Wouda:
For the next release cycle (so, 0.8.0 or so), I would like to try to get some form of mixed backhaul into PyVRP. Right now we only support deliveries from the depots to the clients, and I would like to also add return shipments from the clients to the depots.

[João Santos et al. (2023)](https://doi.org/10.1016/j.omega.2023.102864) provide an overview of many VRPs with pickup and delivery, including various forms of backhaul.

_Originally posted by @N-Wouda in https://github.com/PyVRP/PyVRP/issues/331#issuecomment-1913578985_

leonlan:
After #471, we should find a set of benchmark instances for a variant of VRPSPD. Some options are:
- [Montané and Galvão (2006)](https://www.sciencedirect.com/science/article/abs/pii/S0305054804001741): 18 VRPSPD instances with [100-400] clients
	- Issue: can't find these instances. I have contacted Anand Subramanian, since he also used this to benchmark ILS-RVND-SP.
	- If we use these instances then we can compare with Vidal et al. (2014).
- [Queiroga et al. 2019](https://www.sciencedirect.com/science/article/pii/S0377221720304008)
	- Instances can be found here: http://www.vrp-rep.org/datasets/item/2019-0003.html
	- We could use these instances instead and compare with a bunch of metaheuristics from https://link.springer.com/article/10.1007/s11590-020-01564-5.

N-Wouda:
I'm keeping this open just until we get some benchmark results in.

N-Wouda:
#482 is related to this - we should figure that one out to get the full benchmark runs in.

N-Wouda:
> * It's 300 instances (3 variants for all 100 X instances), so that's a bit much for a benchmark.

What's the stopping criterion? Intuitively I agree that this is a bit much and we might want to take a subset of these instances for PyVRP, but if the runtime limit is small we might also be able to keep all of them. If not, maybe we should take the ~30 biggest instances for each variant, for a total of 90 instances?

leonlan:
The stopping criterion was unspecified, so I went for the same convention as CVRP/X instances: $2.4 * n$, where $n$ is the number of clients.

That’s an average of 20 minute per instance * 300 instances = 100h. Multiply by 10 seeds and we have 1000h. That’s about the same as VRPTW and PCVRPTW (60 * 2 * 10 = 1200h).

I think taking the 30 biggest instances per category is a good idea!

[...]

PR #485, opened 2024-02-23
Support VRPLIB backhaul instances

leonlan:
This PR adds support for VRPLIB instances with backhaul data. 

Part of #434.


pyvrp/read.py
```diff
@@ -162,6 +167,16 @@ def read(
 
     prizes = round_func(instance.get("prize", np.zeros(dimension, dtype=int)))
 
+    if instance.get("type", None) == "VRPB":
```
N-Wouda, line 170:
```suggestion
    if instance.get("type") == "VRPB":
```

N-Wouda: (this is the default value)

pyvrp/read.py
```diff
@@ -116,6 +116,11 @@ def read(
     if max_duration != _INT_MAX:
         max_duration = round_func(np.array([max_duration])).item()
 
+    if "backhaul" in instance:
+        backhauls: np.ndarray = instance["backhaul"]
+    else:
+        backhauls = np.zeros(dimension, dtype=int)
+
     if "demand" in instance:
         demands: np.ndarray = instance["demand"]
```
N-Wouda, line 125: Should we also add support for a linehaul section?

docs/source/dev/supported_vrplib_fields.rst
```diff
@@ -55,6 +55,10 @@ PyVRP supports the following data sections:
 .. glossary::
    :sorted:
 
+   ``BACKHAUL_SECTION``
+      Array of backhaul quantities, one for each location.
+      The depot locations *must* have a backhaul quantity of zero.
+
    ``DEMAND_SECTION``
       Array of demands, one for each location.
       The depot locations *must* have zero demand.
```
N-Wouda, line 64: Maybe make the difference between backhaul (client back to depot) and demand/linehaul (depot to client) a bit clearer. Some thing like

```suggestion
   ``BACKHAUL_SECTION``
      Array of backhaul quantities, one for each location.
      This is the amount picked up at the location and transported back to the depot.
      The depot locations *must* have a backhaul quantity of zero.

   ``DEMAND_SECTION``
      Array of demands, one for each location.
      This is the delivery amount from the depot to the client.
      The depot locations *must* have zero demand.
```

Additionally, I think all references to depot locations needing to have zero values (here, and in various other places) can be removed - the `Depot` type no longer gets those arguments anyway. I missed updatting this when separating the structs.

tests/data/RC101.vrp
```diff
@@ -0,0 +1,29 @@
+NAME: RC101
+COMMENT: Converted to TSPLIB by Accorsi & Cavaliere. Then converted to PyVRP's VRPLIB dialect.
```
N-Wouda, line 2: Maybe change the name (and file name) to something like "SmallVRPSPD.vrp", and update the comment to "Based on the RC101 file <converted to ....>.

It's a bit confusing that this is in fact, not, the RC101 instance. Just a small subset of it.

tests/test_read.py
```diff
@@ -353,3 +353,86 @@ def test_mdvrptw_instance():
...
+    assert_equal(vehicle_type.num_available, 4)
+    assert_equal(vehicle_type.capacity, 200)
+
+    # The first client is a linehaul client (only delivery, no pickup), and
+    # the second client is a backhaul client (only pickup, no delivery). All
+    # other clients have both delivery and pickup.
+    clients = data.clients()
+
+    assert_(clients[0].delivery > 0 and clients[0].pickup == 0)
+    assert_(clients[1].delivery == 0 and clients[1].pickup > 0)
```
N-Wouda, line 382: Can we make this exact? Now that we only have a few clients we can probably write out what these values need to be exactly.

tests/test_read.py
```diff
@@ -353,3 +353,86 @@ def test_mdvrptw_instance():
...
+
+    for client in data.clients()[2:]:
+        assert_(client.pickup > 0 and client.delivery > 0)
```
N-Wouda, line 385: Yeah just wrap this into one large exact check. Maybe something like:
```python
deliveries = [...]
pickups = [...]

for idx, client in enumerate(data.clients()):
  assert_allclose(client.delivery, deliveries[idx])
  assert_allclose(client.pickup, pickups[idx])
```

tests/test_read.py
```diff
@@ -353,3 +353,86 @@ def test_mdvrptw_instance():
...
+
+    # Test that distance/duration are not set to a large value, as in VRPB.
+    assert_(np.all(data.distance_matrix() <= 142))  # max distance in instance
+    assert_(np.all(data.duration_matrix() <= 142))  # max duration in instance
```
N-Wouda, line 389:
```suggestion
    # Test that distance/duration are not set to a large value, as in VRPB.
    assert_allclose(np.max(data.distance_matrix()), 142)
    assert_allclose(np.max(data.duration_matrix()), 142)
```

tests/test_read.py
```diff
@@ -353,3 +353,86 @@ def test_mdvrptw_instance():
...
+
+    for client in clients[50:]:
+        assert_(client.pickup > 0)
+        assert_(client.delivery == 0)
+
+    # Tests that distance/duration from depot to backhaul clients is set to
+    # a large value, as well as for backhaul to linehaul clients.
+    linehauls = set(range(1, 51))
+    backhauls = set(range(51, 101))
+    max_value = 1_000_000
```
N-Wouda, line 426: This is supposed to be exactly `MAX_USER_VALUE`, right? We can import that from `pyvrp.constants` right? E.g. `from pyvrp.constants import MAX_USER_VALUE`, and then make the checks below exact using `assert_allclose` with `MAX_USER_VALUE` (where appropriate). All checks that should be smaller can remain inequalities, but this should provide some guarantees that we are indeed setting the correct value in the correct places.

N-Wouda (approved):
LGTM. Most of my comments are minor updates to make the tests a bit better and clarify some stuff. Can be merged right after!

PR #490, opened 2024-02-25
Switch over to int64_t

### LINKED ISSUE #468: Switch to int64_t?
Our current use of `int32_t` is sufficient for most simple routing problems, but makes it hard to balance multiple different cost aspects in the objective. Suppose, for example, that the distance factors in at a cost of 0.21 euro/km, and the driver's hourly costs amount to some 30 euros. Taking both into account in the objective would amount to a route cost of $(0.21 / 1000) d + (30 / 3600) t $, where $d$ and $t$ are the route's distance and duration in meters and seconds, respectively.

Turning this expression into something integral requires the multiplication by $1000 \cdot 3600$, resulting in an integral objective function of $756 d + 30000 t$. With this objective, the limits of a `int32_t` are such that we cannot have routes longer than about 20h, or 2800km. Either of those maxima seem very easy to reach in practical applications (e.g., long haul).

Switching to a larger integer type would help with this, but we'll need to be careful to check the impact on performance of this change.

N-Wouda:
This PR:

- [x] Closes #468.

<details>

**Notes**:

</details>


pyvrp/constants.py
N-Wouda, line 1: @leonlan this keeps the max user value and internal max value, but makes them much much bigger. $2^{48} \approx 2.8 \cdot 10^{14}$, and $2^{52} \approx 4.5 \cdot 10^{15}$. That should be plenty large, and leave sufficient size to spare to avoid overflows.

leonlan: Question: Why do we have a separate `MAX_USER_VALUE` and `MAX_VALUE` instead of a single maximum value?

leonlan (approved):
LGTM. No other comment besides the max user value/max value distinction, which we can just make one value.

PR #509, opened 2024-03-01
Drop support for Solomon-style instances

### LINKED ISSUE #506: Remove `instance_format` argument
We no longer support Solomon-style instances in our benchmarks, so we should remove the `instance_format` argument in `read.py`. This also means we have to replace some old Solomon-style data files in tests.

leonlan:
This PR:

- [x] Closes #506.

<details>

**Notes**:

</details>


tests/search/test_neighbourhood.py
```diff
@@ -98,6 +98,7 @@ def test_neighbourhood_params_does_not_raise_for_valid_arguments(
         # fmt: on
     ],
 )
+# TODO
```
leonlan, line 101: This test fails but I don't know (yet) why.

```
______________________________________________________________________________ test_compute_neighbours[18-20-34-True-False-99-expected_neighbours_check3] ______________________________________________________________________________
...

    )
    def test_compute_neighbours(
        rc208,
        weight_wait_time: int,
        weight_time_warp: int,
        nb_granular: int,
        symmetric_proximity: bool,
        symmetric_neighbours: bool,
        idx_check: int,
        expected_neighbours_check: set[int],
    ):
        """
        Tests ``compute_neighbours`` on several well-understood cases.
        """
        params = NeighbourhoodParams(
            weight_wait_time,
            weight_time_warp,
            nb_granular,
            symmetric_proximity,
            symmetric_neighbours,
        )
        neighbours = compute_neighbours(rc208, params)
    
        assert_equal(len(neighbours), rc208.num_locations)
        assert_equal(len(neighbours[0]), 0)
    
        # We compare sets because the expected data (from the old C++
        # implementation) sorts by client ID (ascending), not proximity.
>       assert_equal(set(neighbours[idx_check]), expected_neighbours_check)
E       AssertionError: 
E       Items are not equal:
E        ACTUAL: {9, 10, 11, 12, 13, 14, 15, 16, 17, 20, 22, 24, 47, 52, 53, 55, 56, 57, 58, 59, 64, 65, 66, 69, 74, 80, 82, 83, 86, 87, 88, 90, 91, 98}
E        DESIRED: {9, 10, 11, 12, 13, 14, 15, 16, 20, 22, 24, 47, 52, 53, 55, 56, 57, 58, 59, 60, 64, 65, 66, 69, 74, 80, 82, 83, 86, 87, 88, 90, 91, 98}
```

leonlan: Turns out that the VRPLIB Solomon instances (from LKH-3) were different in 1) the number of vehicles and 2) the rounding function. I have changed the instances to match the original instances.

N-Wouda: Can we then also get rid of the new `# TODO` comment? :-)

tests/test_read.py
```diff
@@ -159,13 +159,11 @@ def test_reading_En22k4_instance():  # instance from CVRPLIB
         assert_equal(data.location(loc).required, True)
 
 
-def test_reading_RC208_instance():  # Solomon style instance
+def test_reading_RC208_instance():
```
N-Wouda, line 162: Do we still want to ahve this test now that we only do VRPLIB? I think the test above this already covers VRPLIB; not sure if another test is valuable.

N-Wouda (approved):
LGTM

tests/search/test_neighbourhood.py
leonlan: Oops 😄

tests/test_read.py
leonlan: Yea I'll remove this one and update the above test a bit.

PR #510, opened 2024-03-01
Split 2-OPT into ReverseSegment and SwapTails

### LINKED ISSUE #504: Split 2-OPT
Right now 2-OPT does one of two things (hah!), depending on whether nodes $U$ and $V$ are from the same route or not:
- If $U$ and $V$ are both in the same route, it evaluates the reversal of the segment between $U$ and $V$. This is essentially the "OG" 2-OPT for TSP.
- Otherwise, it attempts to insert the segment $V \rightarrow \cdots$ in place of $U$, and the segment $U \rightarrow \cdots$ in place of $V$. This essentially swaps the tails of both routes.

Let's split this into two operators. The first, in-route 2-OPT can be renamed to `ReverseSegment`, since that is what it does. The second can be either `SwapTails` or `SwapRouteTails`, and be upgraded to a route operator. At that point we can add caching, integrate SwapRoutes as a special case, and do a whole lot more prettiness to make things better.

I think this is a substantial improvement in clarity that may also be faster.

N-Wouda:
This PR:

- [x] Closes #504.

### Benchmarks

This PR (13f8800a6e86fca18bd3f9b7f287970a1512d9aa; job ID 7474761; 1h VRPTW)
```
               MEAN      MIN      MAX   COV
      inst
obj   C1    41381.2  41367.2  41399.3   0.0
      C2    16195.6  16194.1  16197.6   0.0
      R1    47332.3  47293.2  47392.1   0.1
      R2    27792.7  27773.0  27818.7   0.0
      RC1   44326.8  44292.6  44371.6   0.1
      RC2   23282.4  23264.8  23307.9   0.1
iters C1    49953.8  36474.2  57625.8  11.0
      C2    77654.1  56661.7  92990.9  12.1
      R1    23704.5  17219.9  31143.7  14.2
      R2    48094.9  34347.2  59268.9  12.8
      RC1   25269.4  19147.6  29493.1  10.8
      RC2   51344.8  35158.9  61453.1  13.0
```

Current main (0e735b370debb12bdd683ddfe0ee7023582ab5e0; job ID 7339165; 1h VRPTW):
```
               MEAN      MIN      MAX   COV
      inst
obj   C1    41385.9  41341.2  41412.2   0.1
      C2    16196.8  16192.9  16201.3   0.0
      R1    47361.5  47329.2  47402.5   0.1
      R2    27796.2  27766.9  27863.8   0.1
      RC1   44328.7  44289.3  44379.9   0.1
      RC2   23280.6  23254.7  23305.1   0.1
iters C1    48148.1  37462.8  53355.0   9.9
      C2    74614.4  65722.2  79767.4   5.8
      R1    22659.1  19286.8  24750.3   7.2
      R2    45476.6  34275.2  51136.2  10.8
      RC1   23657.6  17187.2  26393.1  11.2
      RC2   48343.5  33959.3  58083.1  13.6
```

<details>

**Notes**:

</details>


N-Wouda:
This should be functional, but I want to think about caching opportunities a bit more before declaring it done.

N-Wouda:
OK, so I don't have a good solution for making `SwapTails` a (powerful) route operator. It's much better at the node level. I'll rewrite this PR to move it back down to the node level.

N-Wouda:
Similar performance to current main, but the split to me seems conceptually simpler. @leonlan could you review this?

pyvrp/cpp/search/SwapTails.h
```diff
@@ -0,0 +1,29 @@
...
+namespace pyvrp::search
+{
+/**
+ * SwapTails(data: ProblemData)
+ *
+ * Given two nodes :math:`U` and :math:`V`, tests whether replacing the arc of
+ * :math:`U` to its successor :math:`n(U)` and :math:`V` to :math:`n(V)` by
+ * :math:`U \rightarrow n(V)` and :math:`V \rightarrow n(U)` is an improving
+ * move.
+ */
```
leonlan, line 15: Can we add a note here similar that this is known as `2-OPT*`?

leonlan (approved):
LGTM!

pyvrp/cpp/search/SwapTails.h
N-Wouda: Sure thing!

PR #511, opened 2024-03-01
Remove unnecessary local search operators

### LINKED ISSUE #501: Removing unnecessary local search operators
We briefly discussed in #94 whether we should remove MoveTwoClientsReversed or not, but we need data to support this decision. This issue investigates this further in a more general setting: which local search operators can we remove without affecting PyVRP's performance? The simpler we can make PyVRP, the better!

leonlan:
This PR:

- [x] Closes #501.

<details>

**Notes**:

</details>


N-Wouda (approved):
LGTM. The tests in #501 confirm that this should be OK, and the resulting codebase becomes a bit simpler to manage. Looks good.

PR #514, opened 2024-03-04
Generalised VRP

### LINKED ISSUE #470: Clustered VRP

This issue tracks clustered VRP and generalised VRP. The two are somewhat related, since they both specify clusters (groups) of clients. The difference is in what's required for these clusters:
- [ ] Clustered VRP requires all clients in the same group to be visited in sequence, one after another, before the route may continue with clients outside the cluster. We can already do this by modifying the distance matrix, but that does not work well with e.g. a max distance constraint.
- [x] Generalised VRP requires only a single client in the group is visited, rather than all of them. We already support optional clients, so we do not have to worry about not all locations being visited. But we do still need to build in some sort of support for this.

We do not have any facility in place to deal with such higher-level constraints currently.

Support for clustered and generalised VRP.

#### Additional context

Came up in #441.

N-Wouda:
This PR:

- [x] Is the algorithmic component of #470.

### Benchmark on VRPTW

This PR (2b018b83a633fd7cf3ad0957696d5be1a6d11811; job ID 7649985; 1h VRPTW):
```
                MEAN       MIN       MAX   COV
      inst
obj   C1     41364.9   41335.0   41385.2   0.0
      C2     16201.7   16197.5   16205.2   0.0
      R1     47286.5   47254.4   47351.7   0.1
      R2     27803.7   27775.8   27842.3   0.1
      RC1    44285.2   44251.5   44309.0   0.0
      RC2    23284.1   23265.8   23302.4   0.0
iters C1     88478.7   76109.5  100240.0   7.8
      C2    120671.4  103937.1  133096.9   7.3
      R1     40880.3   33675.1   48299.5  10.3
      R2     76301.3   68023.4   85290.7   7.3
      RC1    41913.2   35430.8   46755.3   8.3
      RC2    81935.6   72636.1   93840.3   7.4
```

Main (0aef08db304b9277778319e609a44716258de08f; job ID 7656251; 1h VRPTW):
```
                MEAN      MIN       MAX   COV
      inst
obj   C1     41366.6  41335.0   41382.6   0.0
      C2     16202.2  16197.8   16206.0   0.0
      R1     47296.9  47257.0   47350.0   0.1
      R2     27810.0  27771.0   27889.0   0.1
      RC1    44294.7  44272.1   44323.5   0.0
      RC2    23288.4  23269.7   23306.8   0.1
iters C1     81435.7  65376.3   88378.2   9.7
      C2    111474.8  92576.1  121856.0   8.7
      R1     38397.6  29765.3   41729.7  10.7
      R2     70378.4  54651.9   82771.7  12.8
      RC1    38911.3  30863.4   42913.5  11.9
      RC2    75934.3  61226.7   82553.1  10.1
```

no meaningful difference

<details>

**Notes**:

</details>


N-Wouda:
Very early PR that I'll work on over the coming weeks.

N-Wouda:
This was easier than I thought it'd be. I have something barely functional now. Next up are many more tests, and a hunt for some benchmark instances?

N-Wouda:
Somewhat blocked on the issue of how to make VRPLIB understand mutually exclusive groups.

N-Wouda:
Unblocked because Leon and I had a good discussion about this and we figured out a way to make this work with VRPLIB. So I'll be able to finish a barebones version of this PR now.

However, I do want to let this cook a bit longer. The whole `client_group` rather than `mutually_exclusive_group` might be better long-term, because doing clusters via large in/out cluster distances does not mesh well with a max distance constraint (of #500). Generalising it a bit upfront is probably not a bad idea. We can start with a mandatory `mutually_exclusive=True` and go from there.

N-Wouda:
To figure out: why is running the new instance not deterministic? What's going on?

N-Wouda:
I think I fixed the issue. It was an "improving cycle" where:
- Client U was inserted after V;
- Client U was swapped for a different client from its group;
- Client U was inserted after V again;

I fixed this by having the insert check that U can only be inserted if its client group is not already in the solution - else U will be dealt with when checking the group moves.

pyvrp/cpp/search/primitives.cpp
```diff
@@ -85,3 +85,49 @@ pyvrp::Cost pyvrp::search::removeCost(Route::Node *U,
 
     return deltaCost;
 }
+
+pyvrp::Cost pyvrp::search::inplaceCost(Route::Node *U,
```
leonlan, line 89: Not the PR for this, but this operator may also be interesting for PCVRP because we only remove/insert clients now. What do you think?

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -251,47 +254,110 @@ void LocalSearch::applyEmptyRouteMoves(Route::Node *U,
...
     }
+
+    if (mustInsert || bestCost < 0)
+    {
+        UAfter->route()->insert(UAfter->idx() + 1, U);
+        update(UAfter->route(), UAfter->route());
+    }
+}
+
+void LocalSearch::applyGroupMoves(Route::Node *U,
```
leonlan, line 312: Just to summarize for myself:
- Assumption: if U has a group, then at least one client of the group must be in the solution.
	- This assumption is satisfied, because `applyOptionalClientMoves` makes sure that if a required group has no clients in the solution, then it will be inserted.
- Find all inserted group clients and remove all but the one with the lowest delta cost. This is V.
- Swap U for V if U != V and if improving.

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -251,47 +254,110 @@ void LocalSearch::applyEmptyRouteMoves(Route::Node *U,
...
+        auto const groupInSol = std::any_of(group.begin(), group.end(), pred);
+
+        if (groupInSol)  // then we'll have to wait for the group moves to
+            return;      // determine what to do.
+
+        // U must be inserted when it is in a required mutually exclusive client
+        // group that currently is not in the solution at all.
+        assert(group.mutuallyExclusive);
+        mustInsert = group.required && !groupInSol;
+    }
```
leonlan, line 273: Just to summarize for myself: The reason why this part is here instead of `GroupMove` is to prevent clients from being inserted when another group client is already in the solution. The original `applyOptionalClientMoves` would just insert the optional client if improving, which is redundant if they are being removed afterwards again.

But the main reason seems to be that this always ensures that all groups are part of the solution before applying group moves.

leonlan (commented):
Some comments about the neighborhood. I also left some notes in the new LS procedure for optional and group moves. I found it a bit difficult to get through the first time, because of the implicit assumption that "a group is always present in the solution in `applyGroupMove`". I don't have any concrete suggestions right now (due to time), but maybe these notes can already help to clarify some things.

pyvrp/search/neighbourhood.py
```diff
@@ -82,13 +82,21 @@ def compute_neighbours(
...
-    k = min(params.nb_granular, data.num_clients - 1)  # excl. self
+    for group in data.groups():
+        if group.mutually_exclusive:
+            # Clients in mutually exclusive groups cannot neighbour each other,
+            # since only one of them can be in the solution at any given time.
+            # We use max float, not infty, to ensure these clients are ordered
+            # before the depots: we want to avoid same group neighbours, but it
+            # is not problematic if we need to have them.
+            idcs = np.ix_(group.clients, group.clients)
+            proximity[idcs] = np.finfo(np.float64).max
```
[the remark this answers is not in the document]
N-Wouda: > In other words, should we exclude neighbors that have a proximity of `np.inf`? I think we have to do this, because when both the depot and clients of the same group have a proximity of `np.inf`, then what can happen is that the depot becomes a neighbor.

`np.inf` is not the same as `np.info(np.float64).max`. In particular, `np.info(np.float64).max < np.inf`, and a depot can thus never become a neighbour.

pyvrp/cpp/search/LocalSearch.cpp
N-Wouda: Si!

leonlan (approved):
LGTM! Separation of optional client and group moves is much easier to follow.

Issue #491, opened 2024-02-25
Should we keep double precision?


N-Wouda:
With 64-bit integers we can support very large integer values. That calls into question whether the continued support for double-precision computations is still worth it. It adds complexity to the implementation through a feature flag, and doubles our CI requirements because we test everything both in double and integer precision mode.

N-Wouda:
I am leaning no, but we should probably test if we can do without by running e.g. the MDVRPTW benchmark instances in integer precision. Those did not used to work before, but with int64 and some scaling they likely will be fine now.

leonlan:
I'm also in favor of removing this. 

> We might want to add an "exact" round func to read() that defaults to what the EXACT_2D specification in VRPLIB does: multiply everything by 1000, and round down to the nearest integer.

We can also just change this in the VRPLIB format and let `vrplib` handle this (Together with other rounding covnentions such as DIMACS).

N-Wouda:
@leonlan do you know if we can remove this? Basically I'm asking if int precision now works for MDVRPTW or not.

leonlan:
Haven't checked this yet. I assigned it to myself for next week.

N-Wouda:
> @N-Wouda Will you remove double precision mode? I'm not so familiar with the `Measure` type :-)

Sure but this is also a great learning opportunity 😄

PR #517, opened 2024-03-06
Exact round function

leonlan:
This PR adds the `exact` round function for benchmark instances that require double precision. Also `trunc1` is removed and replaced with `dimacs`.


pyvrp/read.py
```diff
@@ -67,8 +70,8 @@ def read(
 
             * ``'round'`` rounds the values to the nearest integer;
             * ``'trunc'`` truncates the values to be integral;
-            * ``'trunc1'`` or ``'dimacs'`` scale and truncate to the nearest
-              decimal;
+            * ``'dimacs'`` scales by 10 and rounds down to an integer;
+            * ``'exact'`` scales by 1000 and rounds to the nearest integer.
```
leonlan, line 74: In view of #491, there may be instances that require double precision for which the `exact` round function is slightly off. Solution costs are rounded by 2 decimals, so routes with more than 10 clients can result in slightly different costs.

tests/test_read.py
```diff
@@ -209,6 +192,28 @@ def test_round_func_round_nearest():
...
+    # We're going to test dist(0, 1) and dist(1, 0), which should be the same
+    # since the distances are symmetric/Euclidean.
+    assert_allclose(data.location(0).x, 40000)
+    assert_allclose(data.location(0).y, 50000)
+
+    assert_allclose(data.location(1).x, 25000)
+    assert_allclose(data.location(1).y, 85000)
+
+    # Compute the distance, and assert that it is indeed correctly rounded.
+    dist = sqrt((40 - 25) ** 2 + (85 - 50) ** 2) * 1000
```
N-Wouda, line 212: Could we perhaps add a thousand separator like `40_000`?

pyvrp/read.py
```diff
@@ -21,12 +20,16 @@ def round_nearest(vals: np.ndarray):
     return np.round(vals).astype(int)
 
 
-def convert_to_int(vals: np.ndarray):
+def round_down(vals: np.ndarray):
     return vals.astype(int)
```
N-Wouda, line 24:
```suggestion
def truncate(vals: np.ndarray):
    return vals.astype(int)
```
Maybe? Since we also call this function `"trunc"` in the `ROUND_FUNCS` options.

pyvrp/read.py
```diff
@@ -67,8 +70,8 @@ def read(
 
             * ``'round'`` rounds the values to the nearest integer;
             * ``'trunc'`` truncates the values to be integral;
-            * ``'trunc1'`` or ``'dimacs'`` scale and truncate to the nearest
-              decimal;
+            * ``'dimacs'`` scales by 10 and rounds down to an integer;
```
N-Wouda, line 73: Maybe also truncates rather than "rounds down"

pyvrp/cli.py
```diff
@@ -67,11 +67,16 @@ def tabulate(headers: list[str], rows: np.ndarray) -> str:
     return "\n".join(header + content)
 
 
-def write_solution(where: Path, data: ProblemData, result: Result):
+def write_solution(
+    where: Path, data: ProblemData, result: Result, round_func: str
+):
     with open(where, "w") as fh:
+        factor = ROUND_FUNCS[round_func](1)
+        cost = round(result.cost() / factor, 2)
```
N-Wouda, line 75: I'm not so sure about this factor business here - see Slack for some ideas!

N-Wouda (approved):
LGTM, a few comments

Issue #454, opened 2024-02-07
Parameter presets for short and long runs


N-Wouda:
#### Is your feature request related to a problem? Please describe

Our current default parameter and algorithm configuration settings are optimised for long run-times - in many cases, for runs of 1-2h. There they find very good results, often at or near the BKS.

A common use case, however, is to try to find the best result in much shorter time frames. Say, one or two minutes. We currently have no good presets for such short runtimes, and also offer no way to easily switch between different presets depending on the use case - other than writing their own `solve` routine and setting all parameters up themselves.

#### Describe the solution you'd like

That we tune PyVRP with short runtimes, collect a good default preset for such runtimes, and allow a user to easily switch between presets. The latter might be something like `Model.solve(..., config=preset.FAST)`, or `preset.BEST` like we already have today.

#### Describe alternatives you have considered

A reasonable alternative is to let the user figure all this out, but that seems burdensome. If we can provide a good default for this use case, that might help a lot already.

leonlan:
After figuring out how to load configs in `Model.solve`, we can use `Model.solve` also in the CLI.

leonlan:
Reminder that we should focus on minimizing number of used routes with short time limits. Having too many routes slows the convergence.

leonlan:
Might no longer be necessary if we go for #533. The current runtime-based convergence of acceptance and destroy rates already perform well and take care of short/long runtimes.

N-Wouda:
Don't know if this'll still be relevant with ILS. I suppose we'll need to experiment once it's almost done.

N-Wouda:
Also related to #944.

N-Wouda:
Subsumed by #944.

PR #523, opened 2024-03-08
SolveParams class and separate solve function

leonlan:
This PR introduces a SolveParams class for solver parameter configuration and separates the `solve` function, which both Model and CLI use. This is useful for addressing #454.


leonlan:
@N-Wouda Can you give me your opinion on this?

pyvrp/search/__init__.py
```diff
@@ -17,7 +17,7 @@
 from .neighbourhood import NeighbourhoodParams as NeighbourhoodParams
 from .neighbourhood import compute_neighbours as compute_neighbours
 
-NODE_OPERATORS = [
+NODE_OPERATORS: list[NodeOperator] = [
```
leonlan, line 20: Mypy complains about this but I don't understand why.

N-Wouda: These are classes, not instantiated objects. This should work: `list[Type[NodeOperator]]` (using `typing.Type`).

pyvrp/ParamConfig.py
```diff
@@ -0,0 +1,81 @@
...
+from pyvrp.search import (
+    NODE_OPERATORS,
+    ROUTE_OPERATORS,
+    NeighbourhoodParams,
+    NodeOperator,
+    RouteOperator,
+)
+
+
+class ParamConfig:
```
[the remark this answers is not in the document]
N-Wouda: Not sure about the name, let's discuss that a bit more.

leonlan:
Will continue working on this later this week!

pyvrp/cli.py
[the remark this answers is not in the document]
leonlan: It takes about 2 seconds to construct a 1000-client instance.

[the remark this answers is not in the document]
N-Wouda: Ah I assumed `MaxRuntime`'s clock started ticking the moment it got constructed, but that's not the case. This should be OK then.

pyvrp/cpp/SubPopulation.h
N-Wouda, line 1: The whole `PopulationParams` needs some love, which I'll try to get around to sometime soon.

pyvrp/cli.py
```diff
@@ -310,6 +254,9 @@ def main():
     msg = "Whether to scale stopping criteria values by the number of clients."
     stop.add_argument("--per_client", action="store_true")
 
+    msg = "Whether to display information about the solver progress."
+    parser.add_argument("--display", action="store_true", help=msg)
```
N-Wouda, line 258: Does this work without passing the `stats_dir` argument as well? And how does this interact with the progress bar when solving a set of instances? I am unsure if we should support this from the CLI.

If we do, the argument should probably be added above the `stop` argument group. It's much easier to separate the different parts of the CLI that way.

docs/source/conf.py
```diff
@@ -30,6 +30,7 @@
 autoclass_content = "class"
 autodoc_member_order = "bysource"
 autodoc_typehints = "signature"
+autodoc_preserve_defaults = True
```
N-Wouda, line 33: Looks OK. I didn't see anything weird when browsing the API docs!

pyvrp/SolveParams.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        with open(loc, "rb") as fh:
+            data = tomli.load(fh)
+
+        gen_params = GeneticAlgorithmParams(**data.get("genetic", {}))
+        pen_params = PenaltyParams(**data.get("penalty", {}))
+        pop_params = PopulationParams(**data.get("population", {}))
+        nb_params = NeighbourhoodParams(**data.get("neighbourhood", {}))
+        node_ops = [getattr(pyvrp.search, op) for op in data["node_ops"]]
+        route_ops = [getattr(pyvrp.search, op) for op in data["route_ops"]]
```
N-Wouda, line 92: If not passed, should we then not default to `NODE_OPERATORS`, or `ROUTE_OPERATORS`? That'd match the constructor.

pyvrp/__init__.py
```diff
@@ -6,6 +6,7 @@
 from .Population import Population as Population
 from .Population import PopulationParams as PopulationParams
 from .Result import Result as Result
+from .SolveParams import SolveParams as SolveParams
```
N-Wouda, line 9: I think we should also import `solve` here. See also the other comment.

tests/test_solve.py
N-Wouda, line 1: It's a bit tricky to have these tests when they essentially duplicate the tests that also execute `Model.solve` (so essentially everything in `test_Model.py`). That might not be very useful. Some tests that quickly get us much more bang for buck is to focus on a few small smoke tests to check that e.g. running longer/more iters gets a better solution, and that different seeds get different solutions but also that running twice with the same seed gets the same solution. That could be two or three pretty short tests just to check that this should mostly be OK.

N-Wouda (commented):
Implementation-wise I think this is essentially OK, but I'm not sure about the place it's currently in, and have some concerns about defaults. The comments below should not take more than say 15min to resolve, I think.

pyvrp/cli.py
leonlan: I find it very useful to have `display` available in the CLI when e.g. solving new VRP variants and see if the search is progressing or not. 

I have added a warning about this when solving a set of instances in parallel. When it's not in parallel, then the output is still OK in the sense that the progress bar doesn't update in place. Does that work for you?

pyvrp/solve.py
```diff
@@ -0,0 +1,191 @@
...
+from pyvrp.crossover import ordered_crossover as ox
+from pyvrp.crossover import selective_route_exchange as srex
+from pyvrp.diversity import broken_pairs_distance as bpd
+from pyvrp.search import LocalSearch, compute_neighbours
+
+if TYPE_CHECKING:
+    import pathlib
+
+    from pyvrp.Result import Result
+    from pyvrp.stop import StoppingCriterion
```
N-Wouda, line 18: This can go all the way down to just above `class SolveParams:`, right?

N-Wouda: (The imports below and above seem to be a bit out of whack due to this conditional in the middle)

pyvrp/cli.py
```diff
@@ -222,12 +167,18 @@ def benchmark(instances: list[Path], num_procs: int = 1, **kwargs):
...
     if len(instances) == 1 or num_procs == 1:
         res = [func(arg) for arg in tqdm(args, unit="instance")]
     else:
+        if kwargs.get("display"):
+            warnings.warn(
+                "Solving instances in parallel does not go along well with "
+                "the --display option. Consider disabling this option."
+            )
+
         res = process_map(func, args, max_workers=num_procs, unit="instance")
```
N-Wouda, line 182: See Slack!

N-Wouda (approved):
One small thing about `if TYPE_CHECKING:`

Issue #522, opened 2024-03-08
Inform users that PyVRP only supports integral data


leonlan:
We should explicitly state somewhere that PyVRP only supports integral data. I think this should be included in the quickstart example and somewhere else in the documentation. In both cases, we can advise users to scale their data to deal with floating-point numbers.

**Related**: should we make an exception for `Client.x` and `Client.y` and allow these to be floats? I assume most users will use long/lats as coords. Coordinates are not involved in cost/duration calculations anyway and `Route.centroid` internally is also represented using floats. 

--- 
Example: two points in Amsterdam (West and East)
- 52.3597897384259, 4.79539534000189 
- 52.35536406262091, 4.94998766369004

This needs a 1000x scaling factor to distinguish between the first two points. Could be even more when the points are closer to each other.

N-Wouda:
Some ramblings:

> We should explicitly state somewhere that PyVRP only supports integral data. I think this should be included in the quickstart example and somewhere else in the documentation.

More explicit than the fact that all the types everywhere are `int`?

> **Related**: should we make an exception for `Client.x` and `Client.y` and allow these to be floats?

I'm leaning no because special-casing for this makes the (already somewhat hard) template stuff in `Measure.h` even harder. We could, but "everything's integral everywhere" is pretty easy to explain and for users to remember.

> Coordinates are not involved in cost/duration calculations anyway and `Route.centroid` internally is also represented using floats.

We need the centroid now for two very specific things: in SREX for the angle sort, and for the route neighbourhoods for the SWAP* operator. I think we can remove both uses in the near future, and replace them with something more robust. Then we could get rid of the centroid calculation altogether, which would move this stuff back to the users to compute if they want it (who can then decide themselves what to do).

leonlan:
> More explicit than the fact that all the types everywhere are int?

Yea, I think we can make this more explicit because the fact that we only use or type hint ints is not obvious to beginner Python users. OR-Tools has a similar restriction because of constraint programming, and quite a number of [questions](https://www.google.com/search?client=firefox-b-d&q=or+tools+floating) have been asked about it.

> I'm leaning no because special-casing for this makes the (already somewhat hard) template stuff in Measure.h even harder. We could, but "everything's integral everywhere" is pretty easy to explain and for users to remember.

Sure.

[...]

PR #527, opened 2024-03-10
Return false when Measure type caster cannot properly load source data

N-Wouda:
This PR:

- [x] Part of #522.

<details>

**Notes**:

</details>

### DIFF
--- a/pyvrp/cpp/bindings.h
+++ b/pyvrp/cpp/bindings.h
@@ -72,6 +72,9 @@ template <pyvrp::MeasureType T> struct type_caster<pyvrp::Measure<T>>
     bool load(pybind11::handle src,
               [[maybe_unused]] bool convert)  // Python -> C++
     {
+        if (!PyLong_Check(src.ptr()))
+            return false;
+
         auto const raw = PyLong_AsLong(src.ptr());
         value = pyvrp::Measure<T>(raw);
         return !PyErr_Occurred();
--- a/pyvrp/cpp/bindings.h
+++ b/pyvrp/cpp/bindings.h
@@ -72,11 +72,16 @@ template <pyvrp::MeasureType T> struct type_caster<pyvrp::Measure<T>>
     bool load(pybind11::handle src,
               [[maybe_unused]] bool convert)  // Python -> C++
     {
-        if (!PyLong_Check(src.ptr()))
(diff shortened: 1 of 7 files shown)


pyvrp/cpp/bindings.h
```diff
@@ -72,8 +72,16 @@ template <pyvrp::MeasureType T> struct type_caster<pyvrp::Measure<T>>
     bool load(pybind11::handle src,
               [[maybe_unused]] bool convert)  // Python -> C++
     {
-        auto const raw = PyLong_AsLong(src.ptr());
+        if (!convert && !PyLong_Check(src.ptr()))  // strict int when conversion
+            return false;                          // not allowed.
+
+        PyObject *tmp = PyNumber_Long(src.ptr());
+        if (!tmp)
+            return false;
```
N-Wouda, line 80: If `convert` is `false` the Python object needs to be an `int` - no conversion is allowed. When `convert` is `true`, we accept any object `a` for which `int(a)` does not raise.

pyvrp/cpp/bindings.h
```diff
@@ -69,10 +69,18 @@ template <pyvrp::MeasureType T> struct type_caster<pyvrp::Measure<T>>
...
-              [[maybe_unused]] bool convert)  // Python -> C++
+    bool load(pybind11::handle src, bool convert)  // Python -> C++
     {
-        auto const raw = PyLong_AsLong(src.ptr());
+        if (!convert && !PyLong_Check(src.ptr()))  // only int when conversion
+            return false;                          // is not allowed.
+
+        PyObject *tmp = PyNumber_Long(src.ptr());  // any argument for which
+        if (!tmp)                                  // Python's int() succeeds.
+            return false;
```
N-Wouda, line 79: A consequence of this is that floating point data is all cast to ints, because
```python
x = 3.5
y = int(x)
```
is perfectly fine Python code. As this is in Python and is also how e.g. OR-Tools does it, I don't think we need to raise for floating-point inputs.

leonlan (approved):
Conversion is fine with me!

Could you also add a note about this to the quickstart, e.g., just above the [VRPTW example](https://pyvrp.org/examples/quick_tutorial.html#VRP-with-time-windows). Then we can close #522.

> Note: PyVRP automatically converts all numeric input values to integers. If you have data with decimal values, you must scale them to integers first to get the desired behavior.

PR #528, opened 2024-03-11
Add ClientGroup

N-Wouda:
This PR doesn't do anything functional. It only serves to introduce the `ClientGroup` object, a lot more data validation, and setting everything up to implement the logic of #470 in a follow-up PR.

This PR:

- [x] Is part of #470, broken off from #514.

<details>

**Notes**:

</details>


pyvrp/cpp/ProblemData.cpp
```diff
@@ -249,67 +298,131 @@ size_t ProblemData::numClients() const { return clients_.size(); }
...
+            auto const *msg = "Client not in the group it references.";
+            throw std::invalid_argument(msg);
+        }
+
+        if (client.required && group.mutuallyExclusive)
+        {
+            auto const *msg = "Required client in mutually exclusive group.";
+            throw std::invalid_argument(msg);
+        }
+    }
```
leonlan, line 333: This check feels too stringent. It's not wrong to have a required client inside a mutually exclusive group. We only need to check that there are no more than two required clients per group. I suggest a `num_required` counter in the group checks below and whenever `num_required > 1`, this should raise.

pyvrp/cpp/ProblemData.h
```diff
@@ -162,6 +185,62 @@ class ProblemData
...
+     * required
+     *     Whether visiting this client group is required.
+     * mutually_exclusive
+     *     When ``True``, exactly one of the clients in this group must be
+     *     visited if the group is required, and at most one if the group is
+     *     not required.
+     */
+    class ClientGroup
+    {
+        std::vector<size_t> clients_;  // clients in this group
```
leonlan, line 218: Some questions about the `clients_` variable:
- Does it matter for the implementation if it's ordered?
- Does it matter if there are client duplicates?
- Would we rather use a `set` instead of a vector here?

pyvrp/cpp/ProblemData.cpp
leonlan: I read in `Model` the following explanation:

```python
       if required and group is not None and group.mutually_exclusive:
            # Required clients cannot be part of a mutually exclusive client
            # group, since then there's nothing to decide about.
            raise ValueError("Required client in mutually exclusive group.")
```

I don't know the implications of allowing required clients in a mutually exclusive group on the implementation, so correct me there if needed.

pyvrp/Model.py
```diff
@@ -273,12 +323,18 @@ def data(self) -> ProblemData:
             distances[frm, to] = edge.distance
             durations[frm, to] = edge.duration
 
+        members = [set(group.clients) for group in self._groups]
+        for idx, client in enumerate(self._clients, len(self._depots)):
+            if client.group is not None and idx not in members[client.group]:
+                self._groups[client.group].add_client(idx)
```
leonlan, line 329: The state of the ClientGroup is a bit confusing now: it only perfectly captures the client indices directly after calling `data()`. After calling `add_client` and `add_depot`, the client group's indices are wrong.

Since we are exposing client groups to users, we should ensure that the client group data is always correct.

Suggestion:

- The selected code can be refactored to be called `_update_groups`.
- When adding a client with a group, this client can be simply added to the corresponding group.
- When adding a depot, we clear all groups and then call `_update_groups` to populate them again with new indices. (we might not even need a separate `_update_groups` method)
- In `data()` we don't need to update the groups anymore.

pyvrp/cpp/ProblemData.h
N-Wouda: - No.
- A little cause we iterate over these and if they're bigger than they need to be we do extra work for no gain.
- Iterating a set is (much) slower than iterating a vector. We spend most time iterating, not checking membership. I think we can add a check to `addClient` that throws when the client is already in the group.

pyvrp/cpp/ProblemData.h
```diff
@@ -18,6 +18,7 @@ namespace pyvrp
  *     vehicle_types: List[VehicleType],
  *     distance_matrix: List[List[int]],
  *     duration_matrix: List[List[int]],
+ *     groups: List[ClientGroup] = [],
```
leonlan, line 21: Nit: we can type these now as `list[...]` right?

leonlan (approved):

pyvrp/cpp/ProblemData.h
N-Wouda: Sí!

PR #532, opened 2024-03-16
Print "-" as cost when population is empty

leonlan:
This PR prints a dash as cost in the progress printer whenever the subpopulation is empty. This fixes a ValueError that is raised because of rounding NaN values.

<details>

**Notes**:

</details>

### DIFF
--- a/tests/test_ProgressPrinter.py
+++ b/tests/test_ProgressPrinter.py
@@ -1,3 +1,4 @@
+import pytest
 from numpy.testing import assert_, assert_equal
 
 from pyvrp import (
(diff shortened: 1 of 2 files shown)


tests/test_ProgressPrinter.py
```diff
@@ -147,3 +148,47 @@ def test_should_print_false_no_output(ok_small, capsys):
...
+        pop.add(Solution.make_random(ok_small, rng), cost_eval)
+
+    stats.collect_from(pop, cost_eval)
+    stats.num_iterations = 500  # force printing
+    printer.iteration(stats)
+
+    # The "-" should not be printed anymore, but there should be some output.
+    out = capsys.readouterr().out
+    assert_("-" not in out)
+    assert_(out != "")
```
N-Wouda, line 194: I think this part of the test is somewhat superfluous when we also have the other tests above (which essentially check the same thing). Do you have an example of what the exact output is when one of the subpops is empty? Not necessarily for a test, just for my own curiosity.

N-Wouda (approved):
LGTM

tests/test_ProgressPrinter.py
leonlan: I think I added it mainly as a sanity check. This is what it looks like when feasible is empty:

```
PyVRP v0.8.0a0

Solving an instance with:
    1 depot
    436 clients
    1 vehicle (1 vehicle type)

      500      6s |   0        -        - |  33 1400514125 348530410
     1000     12s |   0        -        - |  41 17041546427 6233398105
```

N-Wouda: Yeah that looks nice!

PR #535, opened 2024-03-19
Clip the penalty parameters to [1, 100K]

N-Wouda:
<details>

**Notes**:

</details>


N-Wouda:
@leonlan like we discussed! 100K is the HGS-CVRP default as well.

leonlan (approved):

PR #546, opened 2024-03-23
Support vehicle-specific distance and duration objectives

### LINKED ISSUE #543: Add support for mixed distance and duration objectives

We currently only support an objective that minimises distance and some penalty terms for the hard constraints. Minimising duration, or more specifically, a weighted combination of distance and duration, is also very common and desirable. We should probably make the weights vehicle type specific. This should be pretty easy to add in now that we have centralised the relevant cost computations.

None, this is unsupported.

N-Wouda:
This PR:

- [x] Implements #543.

<details>

**Notes**:

</details>


N-Wouda:
Should be done now, except for some additional tests.

pyvrp/cpp/CostEvaluator.h
```diff
@@ -132,7 +139,6 @@ class CostEvaluator
      */
     template <bool exact = false,
               bool skipLoad = false,
-              bool skipDuration = false,
```
N-Wouda, line 135: We don't actually use this anywhere, so for now let's remove it. We might add it back in once we have a suitable use case.

N-Wouda:
This is ready for review. @leonlan could you have a look?

pyvrp/cpp/ProblemData.h
```diff
@@ -321,6 +323,9 @@ class ProblemData
      *     tw_late: int = np.iinfo(np.int64).max,
      *     max_duration: int = np.iinfo(np.int64).max,
      *     max_distance: int = np.iinfo(np.int64).max,
+     *     unit_distance_cost: int = 1,
+     *     unit_duration_cost: int = 0,
```
N-Wouda, line 327: I'm not a huge fan of these names, but at least it seems clear. Any suggestions for improvement are welcome :).

pyvrp/cpp/ProblemData.h
```diff
@@ -348,6 +353,12 @@ class ProblemData
      *     Maximum route duration. Unconstrained if not explicitly provided.
      * max_distance
      *     Maximum route distance. Unconstrained if not explicitly provided.
+     * unit_distance_cost
+     *     Cost per unit of distance traveled by vehicles of this type. Default
```
leonlan, line 357: Idem below.
```suggestion
     *     Cost per unit of distance travelled by vehicles of this type. Default
```

pyvrp/cpp/ProblemData.h
```diff
@@ -348,6 +353,12 @@ class ProblemData
...
+     *     1.
+     * unit_duration_cost
+     *     Cost per unit of duration on routes serviced by vehicles of this
```
leonlan, line 360: "serviced" is a bit confusing because we also have service duration. Idem below.

```suggestion
     *     Cost per unit of duration on routes using vehicles of this
```

pyvrp/cpp/CostEvaluator.h
leonlan, line 1: So nice to see that only 1 file needs to be changed instead of 4+ 😄

leonlan (approved):
LGTM! I can't really think of a better name for the unit costs. They are clear so I think that's fine!

pyvrp/cpp/CostEvaluator.h
N-Wouda: Agree! I was holding off on stuff like this a bit until we had figured out how to do evaluation in a single place. Now that we have mostly gotten that going it's so much easier to change the objective as needed.

PR #549, opened 2024-03-25
Fix test to also work on Mac

N-Wouda:
A bit annoying that the RNG is subtly different on Mac than Linux/Windows

<details>

**Notes**:

</details>


N-Wouda:
Fingers crossed that this test now passes the CD builds on different platforms :).

PR #557, opened 2024-03-30
Simplify code coverage builds

### LINKED ISSUE #68: Code coverage builds
> ```meson
> if get_option('buildtype') == 'debug'
>     if compiler.has_argument('-fprofile-abs-path')
>         # clang does not have this particular flag, so we only add it when
>         # compiling with gcc. It's helpful in determining code coverage.
>         add_project_arguments('-fprofile-abs-path', language: 'cpp')
>     endif
> 
>     add_project_arguments('--coverage', language: 'cpp')
>     add_project_link_arguments('--coverage', language: 'cpp')
> endif
> ```

This can be handled by the `-Db_coverage=true` option, more or less. It hooks up `--coverage` and produces build targets for `ninja coverage`.

See https://mesonbuild.com/Feature-autodetection.html#coverage for discussion of coverage support.

...

I do not think that fprofile-abs-path is needed since Meson only uses a single working directory -- anyway, the coverage report generator should set things up to work. I guess that flag couldn't hurt...

_Originally posted by @eli-schwartz in https://github.com/N-Wouda/pyvrp/pull/58#discussion_r1096926488_

N-Wouda:
This PR uses Meson's `b_coverage` setting to handle code coverage builds. That now also works with clang (rather than just gcc before). So, this PR:

- [x] Closes #68.

<details>

**Notes**:

</details>

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -89,11 +89,13 @@ jobs:
         run: |
           poetry install --only-root
           poetry run python build_extensions.py --build_type debug --clean
-      - name: Run tests and generate reports coverage
-        run: |
-          poetry run pytest
-          poetry run gcovr --filter pyvrp/cpp/ -o coverage.txt
-      - uses: codecov/codecov-action@v4
+      - name: Run tests
+        run: poetry run pytest
+      - if: matrix.compiler == 'gcc'
+        name: Generate coverage reports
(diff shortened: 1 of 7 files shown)


N-Wouda:
This'll probably involve a bit of tinkering with Codecov because I don't fully understand what they need to see to generate accurate coverage reports.

Issue #475, opened 2024-02-18
Vehicle-dependent distances and durations


N-Wouda:
#### Is your feature request related to a problem? Please describe

Right now PyVRP assumes there is a single distance and duration matrix that applies to all vehicle types. This is unrealistic for heterogeneous fleets.

#### Describe the solution you'd like

A way to set different distances and durations for different vehicle types.

#### Additional context

Brought up in #441 by @Stanpol.

N-Wouda:
A big open question here is interface. I think the following could work:
```python
m = Model()
veh_type1 = m.add_vehicle_type()

m.add_edge(..., vehicle_type=None)  # default, applies to all vehicle types
m.add_edge(..., vehicle_type=veh_type1)  # overrides default - only applies to veh_type1
```
Additionally, we'll need to look into how to make this fast. Maybe find a little literature?

N-Wouda:
I'm tentatively assigning this to the milestone for release v0.10.0, with the understanding that I am not very good at planning two releases ahead :-).

N-Wouda:
To do this we may want to remove/deprecate `pyvrp::ProblemData::dist` and `pyvrp::ProblemData::duration`, and instead require access via `pyvrp::ProblemData::distanceMatrix` (and `durationMatrix`). That way we can parameterise the matrix accessors with a `size_t vehicleType` argument to return the correct matrix. [Later, if and when we also want to allow time-dependent VRP, we can add another argument to indicate the time.]

N-Wouda:
This should also get us #329: for vehicle types not allowed to visits certain locations, simply set all arcs to those locations to a very high cost. That should guarantee those locations are not visited by those vehicle types.

N-Wouda:
> Additionally, we'll need to look into how to make this fast.

I'm thinking of somehow tagging the underlying matrices somehow. If vehicles use the same matrix, we can use the cached data on `pyvrp::search::Route` when evaluating moves. Only when they don't we would need to recompute. This can be as simple as storing a distance/duration matrix index on the `VehicleType`, which can be used to as the `ProblemData` instance for the actual matrix. If the indices are the same on two different vehicle types we can just reuse the cached data; else we recompute.

N-Wouda:
Moving this up to 0.9.0.

leonlan:
Moidfied X instances for HVFRP (with heterogeneous costs and distances) can be found in 
[...]

PR #566, opened 2024-04-05
Add routing profiles

N-Wouda:
This PR:

- [x] Is part of #475.

<details>

**Notes**:

</details>


N-Wouda:
I think I'll first do a PR that only removes `ProblemData::dist` and `ProblemData::duration`, so this PR doesn't blow up so much.

N-Wouda:
I need to think a bit about whether I like this approach or not. I'll probably continue in a different PR now that I've seen what this is going to look like.

N-Wouda:
There's some new work over at https://github.com/PyVRP/PyVRP/tree/routing-profiles, which supersedes this branch.

PR #580, opened 2024-05-03
PyVRP devcontainer

leonlan:
This PR adds a `devcontainer` folder that can be used to conveniently setup a functional development environment with Github Codespaces (and other containers). Part of #384.

## Notes
- Based on https://github.com/scipy/scipy/blob/main/.devcontainer/devcontainer.json


.devcontainer/.clang-format
N-Wouda, line 1: Does it not understand the top-level clang-format file?

N-Wouda:
I haven't tried the whole codespaces thing before but I can imagine it's pretty nice for new users that do not want to go through the hassle of setting up a local installation of everything.

.devcontainer/setup.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Install poetry
+curl -sSL https://install.python-poetry.org | python3 -
+
+# Initialize submodules
+git submodule update --init
+
+# Don't ask for superuser acces when debugging extensions
+# See https://github.com/benibenj/vscode-pythonCpp/issues/18
+# echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope # errors
```
leonlan, line 17: This no longer works - is this due to the new jammy image? 

```
tee: /proc/sys/kernel/yama/ptrace_scope: Read-only file system
0
```

.devcontainer/setup.sh
```diff
@@ -0,0 +1,18 @@
...
+# Install poetry
+curl -sSL https://install.python-poetry.org | python3 -
+poetry config virtualenvs.in-project true
+
+# Initialize submodules
+git submodule update --init
+
+# Don't ask for superuser access when debugging extensions
+# See https://github.com/benibenj/vscode-pythonCpp/issues/18
+echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
```
leonlan, line 18: Without this command I couldn't get the debugger to work (see related issue). I honestly don't know what this does and if this is a security issue. It would be nice to get the debugger to work on Codespaces but it's not a must-have for me.

leonlan: Never mind. The change in [5e37607](https://github.com/PyVRP/PyVRP/pull/580/commits/5e37607d4c208f5bf12f1017b99f467672a130cf) works.

N-Wouda:
@leonlan LGTM, I'll merge this in now. Could you have a look tomorrow to see if everything also works as intended for you?

PR #586, opened 2024-05-16
Reparameterize duration segment

wouterkool:
This PR proposes an alternative parameterization for `DurationSegment`.

Currently we have:
* `twEarly`
* `twLate`
* `duration`
* `timeWarp`

However, it should hold that `(twEarly == twLate) || timeWarp = 0`, as there cannot be both slack and timewarp in a route with minimum duration. Currently, this is not checked in the code, and some tests violate this assumption.

This PR proposes to change the parameterization of `DurationSegment` to implicitly enforce this, by replacing `twLate` and `timeWarp` by a single `latestFinish = twLate + duration - timeWarp`. Given only `latestFinish`, we can recover `twLate` and `timeWarp` as follows:
* `timeWarp = max(0, twEarly + duration - latestFinish)`
* `twLate = max(latestFinish - duration, twEarly)`

For consistency, interpretability I propose to rename `twEarly` to `earliestStart` which I think better explains what it actually is.

With this parameterization, I think the merge becomes (slightly?) more intuitive:
* the `earliestStart` for the merged segment is either the earliest start of the first segment, or a later earliest start computed backwards from the earliest start of the second segment
* the `latestFinish` for the merged segment is either the latest finish of the second segment, or an earlier latest finish computed forwards from the latest finish of the first segment

In both of these computations, we should actually consider existing time warp, but you can prove you can omit this as this is already implicit in the `latestFinish`. See a proof here below in 'details':

<details>

```cpp
    // edgeDuration is the travel duration from our last to the other's first
    // client
    Dur const edgeDuration = durationMatrix(idxLast_, other.idxFirst_);

    // We must wait if we arrive before the earliest start time of the other
    // segment. Note that the earliest start time of the other segment may not
    // be the opening of the time window of the first client, but rather the
    // time we should start to avoid unnecessary waiting time. We may still
    // start earlier at that other segment, but this suggests that we should
    // wait at another moment for the same total duration since we cannot
    // finish the other route segment earlier.
    Dur const waitDuration
        = latestFinish_ < other.earliestStart_ - edgeDuration
              ? other.earliestStart_ - latestFinish_ - edgeDuration
              : 0;  // ternary rather than max avoids underflow

    // We compute the new earliest start time for the merged segment backwards
    // from the earliest start time for the second segment.
    // Formally, we should also include timeWarp_ in the first segment in the
    // backwards computation, however:
    // - If timeWarp = 0, we can omit it from the computation
    // - If timeWarp > 0, then earliestStart_ + duration_ - timeWarp_ =
    // latestFinish_
    //   - If waitDuration > 0, it holds that:
    //     other.earliestStart_ - waitDuration - edgeDuration - (duration_ -
    //     timeWarp_)
    //     = latestFinish_ - (duration_ - timeWarp_) = earliestStart_
    //   - If waitDuration = 0, it holds that:
    //     latestFinish_ >= other.earliestStart_ - edgeDuration
    //     so earliestStart_ + duration_ - timeWarp_ >= other.earliestStart_
    //     - edgeDuration
    //     so earliestStart_ >= other.earliestStart_ - edgeDuration - duration_
    //     + timeWarp_
    //  so in both cases the maximum is equal to earliestStart_, and this
    //  remains true if we ignore timeWarp_ in the computation as that will only
    //  decrease the left operand of the max.
    Dur const mergedEarliestStart = std::max(other.earliestStart_ - waitDuration
                                                 - edgeDuration - duration_,
                                             earliestStart_);

    // We compute the new latest finish time for the merged segment forward
    // from the latest finish time of the first segment.
    // Formally, we should also include timeWarp_ in the second segment in the
    // forward computation, however:
    // - If other.timeWarp_ = 0, we can omit it from the computation
    // - If other.timeWarp_ > 0, then other.earliestStart_ + other.duration_ -
    //   other.timeWarp_ = other.latestFinish_
    //   - If waitDuration > 0, it holds that:
    //     latestFinish_ + edgeDuration + waitDuration + other.duration_ -
    //     other.timeWarp_
    //     = other.earliestStart_ + other.duration_ - other.timeWarp_
    //     = other.latestFinish_
    //   - If waitDuration = 0, it holds that:
    //     latestFinish_ + edgeDuration >= other.earliestStart_
    //     so latestFinish_ + edgeDuration + other.duration_ - other.timeWarp_
    //     >= other.earliestStart_ + other.duration_ - other.timeWarp_
    //     = other.latestFinish_
    //  so in both cases the maximum is equal to latestFinish_, and this remains
    //  true if we ignore timeWarp_ in the computation as that will only
    //  increase the left operand of the min.
    Dur const diffDuration = edgeDuration + waitDuration + other.duration_;
    Dur const mergedLatestFinish
        = latestFinish_ < other.latestFinish_ - diffDuration
              ? latestFinish_ + diffDuration
              : other.latestFinish_;  // Avoid overflow
```

</details>

In addition, this PR fixes the tests that have both slack and timewarp, and introduces one additional test case which merges two segments both with time warp, whereas there is waiting time in between to make sure the waiting time and timewarp don't cancel out. I was not completely sure if this was covered already but an extra test doesn't hurt.

- [ ] Closes #xxxx.


wouterkool:
@N-Wouda currently I have kept both the original constructor and a new constructor with the new parameterization to avoid changing all the tests. I wanted to discuss this with you first but we can change everything to only use the constructor with the new parameterization.

N-Wouda:
@wouterkool thanks, I'll have a look tomorrow!

pyvrp/cpp/DurationSegment.h
```diff
@@ -32,24 +32,21 @@ namespace pyvrp
...
  *     Total duration, including waiting time.
- * time_warp
- *     Total time warp on the route segment.
- * tw_early
+ * earliest_start
  *     Earliest visit moment of the first client.
- * tw_late
- *     Latest visit moment of the first client.
+ * latest_finish
+ *    Latest finish moment of the last client.
```
N-Wouda, line 38: Is this the time at which we depart from the last client in the segment, after having serviced them?

pyvrp/cpp/DurationSegment.h
```diff
@@ -132,25 +137,48 @@ DurationSegment DurationSegment::merge(Matrix<Duration> const &durationMatrix,
...
-                             ? other.twEarly_ - atOther - twLate_
-                             : 0;  // ternary rather than max avoids underflow
+    // We must wait if we arrive before the earliest start time of the other
+    // segment. Note that the earliest start time of the other segment may not
+    // be the opening of the time window of the first client, but rather the
+    // time we should start to avoid unnecessary waiting time. We may still
+    // start earlier at that other segment, but this suggests that we should
+    // wait at another moment for the same total duration since we cannot
+    // finish the other route segment earlier.
+    Dur const waitDuration
```
N-Wouda, line 150: This is still a diff/delta in wait duration, right?

pyvrp/cpp/DurationSegment.h
```diff
@@ -132,25 +137,48 @@ DurationSegment DurationSegment::merge(Matrix<Duration> const &durationMatrix,
...
+        = latestFinish_ < other.earliestStart_ - edgeDuration
+              ? other.earliestStart_ - latestFinish_ - edgeDuration
+              : 0;  // ternary rather than max avoids underflow
```
N-Wouda, line 153: For each case that has a ternary rather than a max: is this needed? I.e., does this cause over or underflow if we use max?

And for the cases that use max rather than a ternary: do those never overflow?

I wrote the cases out pretty carefully for the previous implementation, but I do not know yet what's needed for this proposal.

pyvrp/cpp/DurationSegment.h
```diff
@@ -132,25 +137,48 @@ DurationSegment DurationSegment::merge(Matrix<Duration> const &durationMatrix,
...
+    // finish time for the second/other segment or an earlier finish time
+    // computed forward from the latest finish time of the first/current
+    // segment. In this computation, we can ignore existing time warp in the
+    // second segment, since if there is existing time warp, there is no slack
+    // in the second segment and the latest finish time will not change.
+    Dur const diffDuration = edgeDuration + waitDuration + other.duration_;
+    Dur const mergedLatestFinish
+        = latestFinish_ < other.latestFinish_ - diffDuration
+              ? latestFinish_ + diffDuration
+              : other.latestFinish_;  // Avoid overflow
```
N-Wouda, line 175: This is
```cpp
std::max(latestFinish_ + diffDuration, other.latestFinish_);
```
Can `latestFinish_ + diffDuration` overflow?

pyvrp/cpp/DurationSegment.h
```diff
@@ -132,25 +137,48 @@ DurationSegment DurationSegment::merge(Matrix<Duration> const &durationMatrix,
...
+
+    // The new earliest start time for the merged segment is either the
+    // earliest start time for the first/current segment or a later start time
+    // computed backwards from the earliest start time of the second/other
+    // segment. In this computation, we can ignore existing time warp in the
+    // first segment, since if there is existing time warp, there is no slack
+    // in the first segment and the earliest start time will not change.
+    Dur const mergedEarliestStart = std::max(other.earliestStart_ - waitDuration
+                                                 - edgeDuration - duration_,
+                                             earliestStart_);
```
N-Wouda, line 163: Can this underflow?

pyvrp/cpp/DurationSegment.h
```diff
@@ -178,8 +206,9 @@ Duration DurationSegment::duration() const { return duration_; }
 Duration DurationSegment::timeWarp(Duration const maxDuration) const
 {
     // clang-format off
-    return timeWarp_
-         + std::max<Duration>(releaseTime_ - twLate_, 0)
+    Duration const latestStart = std::max<Duration>(latestFinish_ - duration_, earliestStart_);
```
N-Wouda, line 209: If we inline `DurationSegment::latestStart()`, we can call the method here. That saves a bit on having to understand the computation in two places.

N-Wouda (commented):
I think this PR is a nice improvement. What I am not sure about yet is how defensive we need to be - can things over and underflow? If so, in which cases? If not, can we prove that (easily)?

The naming and comments in `DurationSegment::merge` is a bit much that I didn't really find helpful while reviewing (the code itself was clearer), but we can improve that later. It might be more about sequencing than actual content. I think we should first remove the old constructor and update the tests before having another look at this.

pyvrp/cpp/DurationSegment.h
wouterkool: Yes

pyvrp/cpp/DurationSegment.h
wouterkool: Yes, in a sense. It is the duration you have to wait between the two segments that are merged, just like `edgeDuration` is the travel duration between the segments. You could consider `waitDuration` a delta for the total wait duration, but similarly `edgeDuration` is a delta for the total edge duration (or travel duration). Hence I think it makes sense to simply call it `waitDuration` in the context of the merge.

pyvrp/cpp/DurationSegment.h
wouterkool: Well it relies on the assumptions:
* That the edge duration is 'well behaved', i.e. never `max_value` but rather a large well behaved user value, which is checked and warned for by `Model.py`. If a user directly uses `ProblemData` this can give problems but I think the same is true for the current parameterization?
* That `earliestStart` is never infinite or a large value
* That `latestFinish`, replacing `twLate`, can be 'infinite'/`max_value`, which is why I use the ternary here to first check if that is the case.
* That `serviceDuration` for a client is a reasonable value and as such `duration` and `waitDuration` will all be reasonable values.

Summarizing, only `latestFinish` should be problematic as it can be 'infinite' which is handled explicitly, i.e. we either:
* subtract from `latestFinish` which should be ok
* we add to it if we have checked first the addition does not overflow
* we subtract it from something if we have checked first the result is positive

Regarding #590, I merged the test which needed to be slightly adapted as the input contained both timewarp and slack which is not feasible (I removed the slack by setting `tw_early = 18`). I validated that the modified test still reproduces the previous bug, but the proposed implementation passed the test without changes.

pyvrp/cpp/DurationSegment.h
wouterkool: Yes, `latestFinish_` can be infinite if for the depot there is no time window (correct me if wrong). Then adding a positive duration would overflow.

pyvrp/cpp/DurationSegment.h
wouterkool: I don't think so, all values that are subtracted should be 'well behaved' / less than 'infinite' if edge duration is at most the 'user max value'.

pyvrp/cpp/DurationSegment.h
wouterkool: Good idea, thanks :)

N-Wouda (commented):
I don't have more time today to read through the implementation unfortunately. I did find a few small things that we need to fix so I'm posting this early. I'll hopefully have time soon to read the rest: I still need to look at the code in `DurationSegment::merge` in more detail, and go through the tests. My main concerns with `merge` are over/underflow and I think we have most of that covered now, so I don't expect a whole lot more there.

pyvrp/cpp/DurationSegment.cpp
```diff
@@ -3,19 +3,31 @@
...
+      duration_(0),
+      earliestStart_(std::max(depot.twEarly, vehicleType.twEarly)),
+      latestFinish_(std::min(depot.twLate, vehicleType.twLate)),
+      releaseTime_(0)
+{
+    // The start and finish times are limited by both the depot open and
+    // closing times, and the vehicle's start and end of shift, whichever is
+    // tighter. Note that since the duration is 0, the latest finish is equal
+    // to the latest start as given by twLate for the depot and vehicle type.
+}
```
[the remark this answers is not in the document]
wouterkool: I think it's important that both in `Search::Route` and `Solution::Route` we use the same definition, and we also have a dedicated constructor to create a `DurationSegment` for a `Client` so I think it makes sense to have a `DurationSegment` constructor for a `Depot` (in combination with a `VehicleType`) as well to centralise the definition in one place. You really want me to revert it?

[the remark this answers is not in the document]
wouterkool: Done

pyvrp/cpp/DurationSegment.cpp
```diff
@@ -3,19 +3,31 @@
...
 
 DurationSegment::DurationSegment(size_t idx, ProblemData::Client const &client)
     : idxFirst_(idx),
       idxLast_(idx),
       duration_(client.serviceDuration),
-      timeWarp_(0),
-      twEarly_(client.twEarly),
-      twLate_(client.twLate),
+      earliestStart_(client.twEarly),
+      latestFinish_(client.twLate + client.serviceDuration),
```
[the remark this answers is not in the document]
wouterkool: Good catch, fixed as we discussed. `latestFinish_` will now be capped at `INT_MAX`.

pyvrp/cpp/DurationSegment.cpp
```diff
@@ -3,19 +3,19 @@
...
       idxLast_(idx),
       duration_(client.serviceDuration),
-      timeWarp_(0),
-      twEarly_(client.twEarly),
-      twLate_(client.twLate),
+      earliestStart_(client.twEarly),
+      latestFinish_(client.twLate > std::numeric_limits<Duration>::max()
+                                        - client.serviceDuration
+                        ? std::numeric_limits<Duration>::max()
+                        : client.twLate + client.serviceDuration),
```
leonlan, line 18: I just took a look at this PR in more detail after our Slack discussion. I originally it was only about renaming, but why do we need to reparametrize the duration segment? The net change is one less parameter, but we now have to handle an overflow and the implementation diverges from the user interface.

wouterkool: There is no strict need but the current implementation is redundant, i.e. it uses two parameters with a restriction on their combinations so in a sense it is just simpler to make that explicit using a single parameter.

Additionally I think it makes the merge conceptually simpler as its mostly just updating the earliest start and latest finish, which both follow from the earliest start / latest finish from either the first or second segment, depending on which one is more restricting.

Finally (a personal argument) this parameterization is easier to generalize: i.e. with time dependent travel times, time warp is a function of start time while the latest finish is constant, greatly simplifying things.

wouterkool:
I've updated this to be in sync with main. @N-Wouda can you have another look?

N-Wouda:
For multi trip with release times, we need to track latest finish times (or, at least, the latest finish time of the previous segment). It might be worthwhile to revisit this as an alternative parametrisation once we release multi trip.

PR #607, opened 2024-06-11
Computed initial penalty values

### LINKED ISSUE #594: Computed (rather than static) initial penalty values

The objective consists of the following terms:
```
<costs> + <uncollected prizes> + <lots of penalties for infeasibilities>
```
see [here](https://pyvrp.org/api/pyvrp.html#pyvrp._pyvrp.CostEvaluator.cost) for all details.

The only way each of these terms is going to have an effect is if they're roughly on the same scale - if all distances are in the hundred thousands, but e.g. load is just in the dozens, then a small (initial) penalty for load violations won't make the solver consider that constraint: it's too busy optimizing distances further. Only when the penalty values slowly increase over many iterations will it start to devote more search moves to improving the load violations.

Right now our initial penalty values are defined as follows:

https://github.com/PyVRP/PyVRP/blob/d942c9c0ac008e73beb778b0e4f45b928823aae6/pyvrp/PenaltyManager.py#L84-L86

These are hardcoded values that do not take into account the scale of the problem at all. Of course, eventually the penalties will adjust in such a way that we find feasible solutions sufficiently often, but initially the balance can be completely lost. For short optimisation runs, that is problematic.

Smarter initial penalty selection, ideally based on the actual problem data. This should probably take into account that the distance and duration matrices might have substantially differently sized values, which all need to be unified into a single objective.

N-Wouda:
This PR adds a method to compute the initial penalty values from data, rather than using hardcoded values. This should help a lot with performance early on, when the previously hardcoded values might not have been appropriately scaled. As such, it:

- [x] Closes #594.

I should have some time to work out formulas for initial parameter values later this week.

### Benchmarks

Short, 1min benchmarks on VRPTW.

Main (5dbf094f52c422ced29e9fb3ff8729b6a2d6da68; job ID 9927611; 1min / 10 seeds):
```
               MEAN      MIN      MAX  COV
      inst
obj   C1    42005.1  41869.6  42062.7  0.1
      C2    16425.9  16398.8  16488.1  0.2
      R1    49671.5  49438.0  50087.2  0.5
      R2    29294.7  29167.6  29383.8  0.2
      RC1   46235.5  46156.3  46311.0  0.1
      RC2   24517.4  24437.0  24603.2  0.2
iters C1      943.9    880.0    993.7  3.2
      C2     1088.6   1041.0   1147.8  3.5
      R1      401.2    390.5    414.1  2.0
      R2      677.5    615.3    746.1  5.8
      RC1     422.2    399.9    441.8  3.0
      RC2     703.7    666.3    726.6  2.8
```

This PR (6c0ad805a2f18367021d80cbb7fe73fc5a675869; job ID 9927977; 1min / 10 seeds):
```
               MEAN      MIN      MAX  COV
      inst
obj   C1    41980.7  41931.6  42027.8  0.1
      C2    16385.1  16329.8  16459.2  0.3
      R1    48544.5  48412.9  48689.7  0.2
      R2    29281.0  29111.6  29454.9  0.3
      RC1   46151.1  46069.7  46390.2  0.2
      RC2   24434.2  24295.4  24503.6  0.3
iters C1      966.3    766.0   1045.5  8.1
      C2     1125.2   1036.3   1223.1  5.5
      R1      398.5    374.1    425.1  4.7
      R2      670.1    623.0    752.2  5.3
      RC1     428.3    400.8    484.3  5.5
      RC2     730.0    674.4    818.4  6.8
```

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/PenaltyManager.py
+++ b/pyvrp/PenaltyManager.py
@@ -100,15 +100,19 @@ class PenaltyManager:
     recent history, and can be used to provide a temporary penalty booster
     object that increases the penalties for a short duration.
 
+    .. note::
+
+       Consider initialising using :meth:`~init_from_data` to compute initial
+       penalty values from the data instance directly.
+
     Parameters
     ----------
     params
         PenaltyManager parameters. If not provided, a default will be used.
     initial_penalties
         Initial penalty values for unit load (idx 0), duration (1), and
-        distance (2) violations. Defaults to the values 20, 6, and 6. See also
-        :meth:`~init_from_data` to compute potentially better values from
-        a data instance.
+        distance (2) violations. Defaults to ``(20, 6, 6)`` for backwards
+        compatibility.
     """
 
     MIN_PENALTY = 1
@@ -136,12 +140,12 @@ class PenaltyManager:
     ) -> PenaltyManager:
         """
         Initialises from the given data instance and parameter object. The
-        initial penalty parameter values are computed from the problem data.
+        initial penalty values are computed from the problem data.
 
         Parameters
         ----------
         data
-            Data instance to use when computing penalty parameters.
+            Data instance to use when computing penalty values.
         params
             PenaltyManager parameters. If not provided, a default will be used.
         """
--- a/pyvrp/PenaltyManager.py
+++ b/pyvrp/PenaltyManager.py
@@ -149,10 +149,30 @@ class PenaltyManager:
         params
             PenaltyManager parameters. If not provided, a default will be used.
         """
-        # TODO use data
-        init_load = 20
-        init_tw = 6
-        init_dist = 6
+        distances = data.distance_matrices()
+        durations = data.duration_matrices()
+        edge_costs = [  # edge costs per vehicle type
+            veh_type.unit_distance_cost * distances[veh_type.profile]
+            + veh_type.unit_duration_cost * durations[veh_type.profile]
+            for veh_type in data.vehicle_types()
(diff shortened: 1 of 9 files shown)


pyvrp/PenaltyManager.py
```diff
@@ -119,29 +101,53 @@ class PenaltyManager:
...
             [],  # track recent time feasibility
             [],  # track recent distance feasibility
         ]
 
+    @classmethod
+    def init_from_data(
+        cls,
+        data: ProblemData,
+        params: PenaltyParams = PenaltyParams(),
+    ) -> PenaltyManager:
```
N-Wouda, line 133: Using a `classmethod` for this is a bit clunky, but the added benefit is that (a) we keep a way to specify the initial penalties directly, which is nice for testing, and (b) this keeps the default initialisation of
```python
pm = PenaltyManager()
```
valid, with the same outcome. So any dependent code that did not explicitly set the initial penalty values would not be affected by the change.

N-Wouda:
@leonlan could you have a look at this? And, should we benchmark this?

leonlan (approved):
LGTM

N-Wouda:
Benchmarks don't show much downside to this either. And these are the ones we're already tuned on. I'm excited to see how this affects my real world data :-).

N-Wouda:
My new default penalty parameters are `[13040    11     1]`, which is a lot better than what I previously got.

Issue #329, opened 2023-08-06
Vehicle-client restrictions


N-Wouda:
#### Describe the solution you'd like

Currently, all vehicles can be used to visit all clients. This might be OK in static problem settings like the ones we currently support, but falls apart when e.g. re-optimizing routes that are already underway: the deliveries have already been loaded onto the truck, and the clients those deliveries go to cannot be serviced by another vehicle.

I *might* need this for a case I'm solving, but I can also solve my particular instance in a less efficient manner by tweaking the time windows. But that does not hold in general.

#### Describe alternatives you have considered

There aren't any, really. This is currently unsupported.

#### Additional context

We could support this by restricting the vehicle types to which a client may be assigned. There can only be one vehicle of such a type (how do we enforce that cleanly?). Such a client should be assigned to that vehicle type, but I am not sure if we should require that or deal with it via soft penalties like load or time window violations.

N-Wouda:
I think soft violations and a new "swap vehicle" route operator. We'll need such a route operator anyway for #188, and likely for #305 as well.

leonlan:
The VRP with vehicle-client assignment restrictions is called "site-dependent VRP" in the literature. Instead of soft violations, I think you could also do this with a bitmaps.

leonlan:
Just a note that the MDVRPTW instances from Vidal et al. (2013) are also used for the site-dependent VRPTW.

lstegner:
Hello! I am wondering if there is a plan/timeline for site-dependent VRP support in the near future, or if there's some advice/guidance to get a "quick and dirty" solution up and running for some small scale testing for a research project I'm working on.

N-Wouda:
Hi @lstegner! 👋 A realistic timeline for this is before the summer, say in June. We'll add support for site-dependent VRP via #475, which will be part of PyVRP 0.9.0 (the next release). That'll likely take around 2-3 months, as I am currently quite busy finishing up my dissertation.

lstegner:
Hi @N-Wouda! Thanks so much for the quick update. Good luck with your dissertation, and looking forward to the next release 😄

N-Wouda:
Hi @lstegner, with #475 now completed, it is possible to model vehicle-client restrictions using a workaround with profiles. In particular, by making a routing profile per vehicle type, you could model the vehicle-client restrictions as follows:
```python
from pyvrp import Model

m = Model()

# here add your depots and clients

prof1 = m.add_profile()
[...]

PR #550, opened 2024-03-26
Read HFVRP instances

leonlan:
This PR:

- [ ] Closes #xxxx.
- [ ] Adds and passes relevant tests.
- [ ] Has well-formatted code and documentation.

<details>

**Notes**:

- It is essential that you add a test when making code changes.
  This keeps the code coverage level up, and helps ensure the changes work as intended.
- When fixing a bug, you must add a test that would produce the bug in the master branch, and then show that it is fixed with the new code. 
- New code additions must be well formatted. Changes should pass the pre-commit workflow, which you can set up locally using [pre-commit](https://pre-commit.com/#intro). 
- Docstring additions must render correctly, including escapes and LaTeX.

</details>

PR #617, opened 2024-06-21
Read SDVRP instances

leonlan:
This adds support for SDVRP instances. Relates to #329.


leonlan:
I'll work on #550 next after this PR.

docs/source/dev/supported_vrplib_fields.rst
```diff
@@ -26,6 +26,8 @@ PyVRP supports the following specifications:
    ``CAPACITY``
       Homogeneous capacity of all vehicles in the instance.
       The number of vehicles is provided using the :term:`VEHICLES` specification.
+      To specify heterogeneous capacities, see :term:`CAPACITY_SECTION`.
+      When :term:`CAPACITY` and :term:`CAPACITY_SECTION` are both provided, the data from :term:`CAPACITY_SECTION` is used.
```
N-Wouda, line 30: I think we need to do this like with `SERVICE_TIME_SECTION` - just describe the section, and leave it implicit that a specification is also OK (and assumed to then hold for all clients/vehicles/as appropriate).

I am not sure what we should do if both a specification and a section is given. That is not unambiguous. I think that should be an error (possibly already in VRPLIB; because it's overwriting an existing field).

N-Wouda (commented):
A few things I'm not really clear on yet. Let's discuss over Slack.

docs/source/dev/supported_vrplib_fields.rst
leonlan: - I've included `SERVICE_TIME` as a specification. I think it's beter to be implicit about it.
- When a field allows both specification and section, the specification refers to the section.
- We no longer talk about instances that include both specifications and sections for a single field - this will be handled by `vrplib`.

N-Wouda (approved):
LGTM now. @leonlan could you have another brief look?

leonlan:
LGTM, merging this in!

Issue #366, opened 2023-08-31
Multi-trip vehicle routing


leonlan:
In multi-trip vehicle routing, vehicles may serve multiple routes. That is, vehicles can return to the depot after serving one route, and then serve customers from another route.

Two papers propose a solution method for the multi-trip VRP with time windows:
1. [Cattaruzza et al. (2016)](https://pubsonline.informs.org/doi/abs/10.1287/trsc.2015.0608): this paper describes an HGS algorithm.
2. [François et al. (2019)](https://pubsonline.informs.org/doi/abs/10.1287/trsc.2019.0909): this paper describes an ALNS algorithm.

With the current setup, it should not be super difficult to add such an extension. Most importantly, we need:
- [x] Local search operators that insert/remove an intermediate depot (a.k.a. "trip delimiters") in a route.
- [x] The concatenation scheme needs to be modified slightly (see references) because of release times.
	- [x] The references by Cattaruzza and François don't discuss how depot/trip delimiters "renew" the capacity constraints. This needs to be investigated as well.
- [x] How to allow for multi-trip vehicle routing? We should keep single-trip vehicle routing as default. One way to incorporate this is to introduce a `max_number_trips` argument to `ProblemData`. The default is one. Setting this value to None means no restrictions.
      Decision: `max_trips: int = 1` argument on the `VehicleType`.
- [x] Benchmark instances can be obtained from the references.
	- I sent a mail to Kai Wang to ask for multidepot MVRPTWR instances (see [paper](https://www.sciencedirect.com/science/article/pii/S1366554519301486#s0110)).
          Update [23/6/24]: haven't heard back.
- [x] Representation of a multi-trip solution. How do we include trip delimiters? I think the intermediate depots should be made explicit e.g. `Route #1: 1 2 0 3 0 4`. Or we can use "|" as trip delimiter: `Route #1: 1 2 | 3 | 4`.
      Decision: use `|` for now, it's clearer with multi-depot.

N-Wouda:
That will make it a bit more difficult to evaluate swapping, I guess, because one vehicle type might have a different `max_trips` than another.

leonlan:
> That will make it a bit more difficult to evaluate swapping routes, I guess, because one vehicle type might have a different `max_trips` than another.

One way I can think of doing this is by removing a number of the intermediate depots. As a simple heuristic we can remove the last number of intermediate depots.

[...]

PR #627, opened 2024-06-27
Add support for multiple trips to pyvrp.Route

N-Wouda:
Multiple trips takes a few PRs because it's a pretty big change. This PR handles the user-facing part, with `pyvrp.Route` and `pyvrp.Solution`. To do this, we need:
- [x] Multi-trip feasibility checks on `pyvrp.Route`;
- [x] `pyvrp.Route` reload depot accessors, trip iteration support;
- [x] Updated computations to accurately compute distance/duration etc. with multiple trips.

We might want to expose more statistics on `pyvrp.Route` for individual trips, but that's not necessarily a priority for this PR.

This PR:

- [x] Is part of #366.

<details>

**Notes**:

In particular:

</details>


N-Wouda:
Very early PR, but it's a start.

N-Wouda:
Broken, but that's OK. I'll try to finish the feasibility and access functions tomorrow. The computations are probably for next week or so.

pyvrp/cpp/Route.h
```diff
@@ -35,7 +79,6 @@ class Route
     Duration timeWarp_ = 0;        // Total time warp on this route
     Duration travel_ = 0;          // Total *travel* duration on this route
     Duration service_ = 0;         // Total *service* duration on this route
-    Duration wait_ = 0;            // Total *waiting* duration on this route
```
N-Wouda, line 38: This field wasn't actually used. Not sure why we even had this to begin with - it's trivially computable (and, that's indeed what happens when using `wait_duration()`).

N-Wouda:
This is now functional. I first want to expose a `Trip` class before I consider this PR done, because we'd for example like to have some way to get the trip-specific release times out.

N-Wouda:
Adding a separate `Trip` class is more work than I want to put in now. I'll have a look at cleaning this up a bit later this week, and then I think it's largely done.

PR #646, opened 2024-08-27
Pickle the data instance

### LINKED ISSUE #565: Pickle ProblemData
Our `Solution` objects are pickleable, ensuring a `Result` object can be stored to disk easily. But our `ProblemData` instances are not. We should probably also support pickling those objects to ensure they can be stored and re-used.

My motivation for this is the discussion I posted in https://github.com/PyVRP/PyVRP/discussions/564 - if a `ProblemData` instance becomes pickleable, a user can simply provide that pickled object to us, instead of a wide variety of data formats that they use to store their instance data.

N-Wouda:
This PR:

- [x] Closes #565.

<details>

**Notes**:

In particular:

</details>


N-Wouda:
I need to add tests for various new `__eq__` implementations, but this is otherwise good to go.

PR #666, opened 2024-10-27
Release GIL in heavy operations

N-Wouda:
This PR is an experiment. I think we can release the GIL when calling into pure C++-side code, which is particularly relevant for heavy operations like crossover and local search. But I need to do some experimentation first to determine if that's actually valid and whether this works as intended. If it does, one can use multithreading from Python to enable parallel PyVRP work, and things should just work. Should be a lot lighter than default multiprocessing since the GIL will no longer be an issue.

This is a step towards parallelising PyVRP, although I am not entirely sure if we should (nor what such an algorithm should look like exactly).

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/crossover/bindings.cpp
+++ b/pyvrp/cpp/crossover/bindings.cpp
@@ -13,6 +13,7 @@ PYBIND11_MODULE(_crossover, m)
           py::arg("parents"),
           py::arg("data"),
           py::arg("indices"),
+          py::call_guard<py::gil_scoped_release>(),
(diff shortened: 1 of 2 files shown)


N-Wouda:
I just tested this a bit by spinning up a few threads and calling multiple solves concurrently. That obviously doesn't scale (because solve needs the GIL all the time), but it did confirm that the LS calls now release the GIL - and that there are no downsides to doing so. Further, with just one thread, there appears to be no change in performance (as expected).

PR #667, opened 2024-10-28
Migrate to uv

### LINKED ISSUE #659: Migrate to uv
[uv](https://docs.astral.sh/uv/) is a new Python package and project manager. It is extremely fast and maintained by Astral, which is the company that also maintains [ruff](https://docs.astral.sh/ruff/). I find uv to work really well and we should consider replacing poetry by uv.

The main Poetry features that we rely on are supported by uv:
- Managing project dependencies 
- Publishing packages
- Building extensions

See https://docs.astral.sh/uv/concepts/projects/#projects for details.

Not yet supported:
- Group development dependencies (https://github.com/astral-sh/uv/issues/5632)
- Custom task (https://github.com/astral-sh/uv/issues/5903)

We could already migrate now but I suggest waiting for group development dependencies.

leonlan:
This PR:

- [x] Closes #659.

<details>

**Notes**:

In particular:

</details>


leonlan:
@N-Wouda do you want to tinker with the build-backend from here? I can continue with the other stuff TODOs later.

N-Wouda:
@leonlan yeah sure, I'll have a look tonight!

N-Wouda:
RTD seems unrelated (probably wrong command in there now; not sure what's going on). 

I still need to figure out build isolation at some point. But for now this seems work-ish?

leonlan:
Seems that hatchling is not correctly installed in the RTD setup?

```
[rtd-command-info] start-time: 2024-10-28T20:45:22.344736Z, end-time: 2024-10-28T20:45:22.723971Z, duration: 0, exit-code: 2
uv sync --group docs --group examples
Using CPython 3.12.3 interpreter at: /home/docs/.asdf/installs/python/3.12.3/bin/python
Creating virtual environment at: .venv
Resolved 156 packages in 1ms
error: Failed to prepare distributions
  Caused by: Failed to build `pyvrp @ file:///home/docs/checkouts/readthedocs.org/user_builds/pyvrp/checkouts/667`
  Caused by: Build backend failed to build editable through `build_editable` (exit status: 1)

[stderr]
Traceback (most recent call last):
  File "<string>", line 8, in <module>
ModuleNotFoundError: No module named 'hatchling'
```

Note to self: use `uv sync --verbose` to check what happens

meson.build
```diff
@@ -1,7 +1,7 @@
 project(
     'pyvrp',
     'cpp',
-    version: run_command('poetry', 'version', '--short', check: true).stdout(),
+    # version: run_command('poetry', 'version', '--short', check: true).stdout(),
```
leonlan, line 4: Can be called with `uv version` from uv 0.7.0 onwards.

https://github.com/astral-sh/uv/releases/tag/0.7.0

pyproject.toml
```diff
@@ -1,168 +1,222 @@
...
+# musllinux-based builds we assume users can compile the thing themselves.
+build = "cp310-* cp311-* cp312-* cp313-*"
+skip = "pp* *_ppc64le *_i686 *_s390x *-win32 *-musllinux*"
+build-frontend = "build"
+test-requires = "pytest>=6.0.0 pytest-cov>=2.6.1"
+test-command = "pytest {package}/tests"
+
+
+[build-system]
+requires = ["hatchling"]
```
leonlan, line 194: I haven't really followed the development of uv's build backend (it's still in preview), but uv v0.7.0 now supports poetry-core as build backend. 

https://github.com/astral-sh/uv/pull/12781

We could opt for this to migrate to UV since we already have the poetry build backend working.

N-Wouda:
The workflows and docs have now been updated and all seem to work. Locally, this also works well enough. There's just one minor snag left to fix: repeatedly calling `uv sync` does not work, because the meson build dir is currently being re-used, which does not work between isolated builds.

N-Wouda:
This looks like a nice enough start. A few more things to iron out:
- Can we get rid of the wrapper that creates a temp build directory for Meson? It's fairly ugly; I'd like to (re)move it if we can.
- Can we remove the call to `uv version --short` from our meson build? This is now the only reason `uv` is pulled into our build system.

N-Wouda:
I think this is good to go. @leonlan care to have a look/give it a try?

docs/source/setup/installation.rst
```diff
@@ -31,11 +31,11 @@ To run the example notebooks locally, first clone the repository:
 
    git clone https://github.com/PyVRP/PyVRP.git
 
-Then, make sure your Python version has ``poetry``:
+Then, make sure your Python version has ``uv``:
```
leonlan, line 34: I think we can simplify this by just requiring the user to have `uv` installed:

```suggestion
Then, make sure to have [``uv``](https://docs.astral.sh/uv/getting-started/installation/) installed.
```
(and code block below)

.github/workflows/CI.yml
```diff
@@ -39,19 +39,13 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
```
leonlan, line 41: Question: why do we use the `setup-python` action instead of letting uv manage Python versions here?

pyproject.toml
```diff
@@ -1,183 +1,185 @@
...
+    "Programming Language :: Python :: 3.13",
+]
+requires-python = ">=3.10"
+dependencies = [
+    "numpy>=1.15.2; python_version < '3.12'",
+    "numpy>=1.26.0; python_version >= '3.12'",
+    "matplotlib>=2.2.0",
+    "vrplib>=1.4",
+    "tqdm>=4.64",
+    "tomli>=2.0",
```
leonlan, line 37: Question: do we need to specify the patch versions here as well? Looking back it the loss of the patch version is something that happened here: [5c950a1](https://github.com/PyVRP/PyVRP/commit/5c950a147f71f2113ec56fea12fffab229de1ff8)

pyproject.toml
```diff
@@ -1,183 +1,185 @@
...
+[project]
+name = "pyvrp"
+version = "0.12.0a0"
+description = "A state-of-the-art vehicle routing problem solver."
+authors = [
+    { name = "Niels Wouda", email = "nielswouda@gmail.com" },
+    { name = "Leon Lan", email = "leon.lanyidong@gmail.com" },
+    { name = "Wouter Kool", email = "wouter.kool@ortec.com" },
+]
+license = { file = "LICENSE.md" }
```
leonlan, line 10: New format for licenses: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

```suggestion
license = "MIT"
license-files = ["LICENSE.md"]
```

leonlan (commented):
LGTM! Just a few small comments. Excited to use uv :)

pyproject.toml
N-Wouda: I don't really know. Probably not? But I'm not looking to change our deps here so I'll double check to be sure.

.github/workflows/CI.yml
N-Wouda: Cause that does not work well with CodSpeed/Valgrind. Valgrind could be fixed when we run it by specifying `uv run` again in the CI, but we do not have access to CodSpeed's run command, so there's no way to fix that.

Besides, uv is a bit slower because GitHub caches these Python downloads (uv needs to bring them in from outside). I'm sure they could perhaps be cached but I don't really know how, and the CodSpeed/Valgrind thing was already a bit of a dealbreaker.

PR #669, opened 2024-10-30
Require at least one vehicle type

leonlan:
This PR introduces a check that ProblemData instances have at least one vehicle type. Noticed this when reviewing #645. 

Having no vehicle type currently also raises because `compute_neighbourhood` requires at least one vehicle type. Since it doesn't make sense to have zero vehicle types, it's just easier to require one vehicle type and add a more descriptive message.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/tests/test_Model.py
+++ b/tests/test_Model.py
@@ -660,7 +660,7 @@ def test_to_data_client_group():
     """
     m = Model()
     m.add_depot(1, 1)
-    m.add_vehicle_type(1)
+    m.add_vehicle_type()
 
     group = m.add_client_group()
     m.add_client(1, 1, required=False, group=group)
@@ -827,7 +827,7 @@ def test_profiles_build_on_base_edges():
 
     depot = m.add_depot(x=1, y=1)
     client = m.add_client(x=2, y=2)
-    m.add_vehicle_type(1)
+    m.add_vehicle_type()
 
     # Add base edges. These edges are used to construct base matrices that the
     # profiles build on. Essentially, if an edge is not specifically provided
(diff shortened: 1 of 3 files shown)


tests/test_ProblemData.py
```diff
@@ -158,6 +158,30 @@ def test_problem_data_raises_when_no_depot_is_provided():
...
+    Tests that the ``ProblemData`` constructor raises a ``ValueError`` when
+    no vehicle types are provided.
+    """
+    with assert_raises(ValueError):
+        ProblemData(
+            clients=[],
+            depots=[Depot(x=0, y=0)],
+            vehicle_types=[],
+            distance_matrices=[np.asarray([[]], dtype=int)],
+            duration_matrices=[np.asarray([[]], dtype=int)],
```
N-Wouda, line 172: This also raises, no? (inconsistent size)

I think we should have the exact same dimensions here as in the non-raise case below.

tests/test_Model.py
```diff
@@ -660,6 +660,7 @@ def test_to_data_client_group():
     """
     m = Model()
     m.add_depot(1, 1)
+    m.add_vehicle_type(1)
```
N-Wouda, line 663:
```suggestion
    m.add_vehicle_type()
```

tests/test_Model.py
```diff
@@ -826,6 +827,7 @@ def test_profiles_build_on_base_edges():
 
     depot = m.add_depot(x=1, y=1)
     client = m.add_client(x=2, y=2)
+    m.add_vehicle_type(1)
```
N-Wouda, line 830:
```suggestion
    m.add_vehicle_type()
```

N-Wouda (approved):
LGTM, with a tiny comments

tests/test_ProblemData.py
leonlan: good catch 👍

Issue #416, opened 2023-12-11
Performance stuff


N-Wouda:
Discussion thread for ideas to improve PyVRP's performance. Could be low-level stuff, like better code optimisation, smaller data types, and/or caching. Or a smarter data structure, better algorithm, etc. Anything's in scope.

Edit: there's also perhaps some stuff in #181 that we might want to implement as part of this.

N-Wouda:
One idea that might help a bit relates to how we skip move evaluation when nothing's changed since the last time we considered the client pairs $U$ and $V$:

https://github.com/PyVRP/PyVRP/blob/7604244262d342f14b649f1845ecd30eda56bc7e/pyvrp/cpp/search/LocalSearch.cpp#L92-L93

This works fine, but any change in the routes of $U$ and $V$ are now reason to re-evaluate. Even when that change happens many nodes after $U$ or $V$, which has no consequences for the evaluation.

We might make this more efficient by taking into account the size of our move operators (e.g., 1, 2, 3, etc. nodes), and checking whether there has been a recent modification that affected nodes sufficiently near either $U$ or $V$ to warrant a new evaluation.

wouterkool:
> This works fine, but any change in the routes of U and V are now reason to re-evaluate. Even when that change happens many nodes after U or V, which has no consequences for the evaluation.

I like the idea but I'm not sure if this is correct? At least with time windows, removing a node towards the end of the route may create slack which enables inserting a customer early in the route that would previously result in a large penalty, even if the affected parts of the route are far apart.

This situation may be hypothetical or not happen frequently so we may decide heuristically to skip those options but that's a functional change of the algorithm, not a more efficient implementation with the same result. Or am I missing something?

N-Wouda:
> This situation may be hypothetical or not happen frequently so we may decide heuristically to skip those options but that's a functional change of the algorithm, not a more efficient implementation with the same result. Or am I missing something?

Not really. This conclusion seems correct to me.

N-Wouda:
Much more aggressive caching is also in scope here. Routes are queried at least two orders of magnitudes more than that they are changed. Every time something between two indices is calculated, can we somehow store the result? Is that worthwhile?

N-Wouda:
Also relevant:
- Restarts: how often, after what (no improv local best in this restart run, or global best so far over all restarts?)
- SREX: can we limit the time spent in the subsequent local search? (see also #521)
[...]

PR #679, opened 2024-11-08
Do not deref route_ when determining when node is a depot node

N-Wouda:
Part of our push for performance / #416. This PR removes the dereferencing that happens inside `Route::Node::isDepot()` to determine if the node is a depot node. Currently, we do this by index: a node is a depot node if its index in a route is `0` or `route.size() + 1`. For this, we need to access `route.size()`. We can avoid this with some simple pointer checks: the node is a depot if its address corresponds with `startDepot_` or `endDepot_` on the route it's in. Those addresses are fixed offsets from the `route_` address, which is more efficient.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/search/Route.h
+++ b/pyvrp/cpp/search/Route.h
@@ -470,7 +470,8 @@ bool Route::Node::isDepot() const
 {
     // We need to be in a route to be the depot. If we are, then we need to
     // be either the route's start or end depot.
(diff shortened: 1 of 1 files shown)


N-Wouda:
CodSpeed suggests this is 2% faster on average, which is roughly in line with what I'd expect from this small change.

PR #692, opened 2024-11-14
Support latest start on vehicle type

### LINKED ISSUE #639: Rethink vehicle shifts

Our `VehicleType` now have `tw_early` and `tw_late` properties that model a full shift: the shift starts at or after `tw_early`, and finishes at or before `tw_late`. This already limits route durations, but we also have a `max_duration` property to further restrict this in case starting later is better.

This particular formulation does not, however, let us constrain starting times, which is something that I've seen come up a few times now. 

An alternative formulation based on `earliest_start_time`, `latest_start_time`, and `max_duration` would allow this: the vehicle starts between `earliest_start_time` and `latest_start_time`, and then returns after at most `max_duration` time units. If we call this `earliest_start_time` and `latest_start_time` we should also rename this on `Route` (and possibly then remove `end_time`? It's not unambiguous and does not have much value since we also have `duration`).

joopvdh:
This PR:
Adds support for constraining start times for routes.

- [x] Closes #639  .

Implementation details:
- Renamed `twEarly` to `earliestStart` and `twLate` to `latestFinish` in `VehicleType`
- Introduced attribute `latestStart` to `VehicleType`
- Not enforced that `latestStart <= latestFinish`. If `latestStart > latestFinish`, then it is possible to start the route after the latest finish (but this will always result in time warp). 
- Distinction between creating a `DurationSegment` for the start depot or end depot of the `VehicleType`. When creating a duration segment for start depot, then `vehicleType.latestStart` is used for `twLate` of the `DurationSegment`. For the end depot, `vehicleType.latestFinish` is used for `twLate`.

Open questions:
1. Renaming `twEarly` and `twLate` on `VehicleType` is a breaking change, but I do prefer to update the names to `earliestStart` and `latestFinish` for clarity. Should the attributes `twEarly` and `twLate` on `Client` also be renamed for consistency and to make clear that the service should start in this time window but not necessarily end in this window?

2. What should be the default value for `latestStart` in the constructor of `VehicleType`? The most logical default value would be "infinite", just like for `latestFinish` when it is not specified. However, the question is what to do when a value is provided for `latestFinish` but not for `latestStart`. Consider the following options:
    - Set `latestStart` to default value 'infinite' => route can start after latest finish. This was previously not possible, so this leads to the following change in behavior:
        - New behavior: We set the `latestStart` to 'infinite' by default, then the route can start after the route latest finish time. This prevents waiting time to get to the first client (if the client has a time window starting after the latest finish).
        - Old behavior: route cannot start after the latest finish, therefore it is not possible to delay the route start to prevent waiting time at the first client. Solution with old behavior will have higher duration and equal time warp, so it is seen as worse compared to the new behavior.
        - Note that this change in behavior only occurs in unlikely situations where the routes are infeasible anyway. So likely the effect is negligible and if there is an effect, it would be improving the objective.
    - Set `latestStart` to same value as `latestFinish` => backwards compatible (`latestStart` does not have to be provided when `latestFinish` is provided). However, this does require the `latestStart` argument to become of type `Optional` with default `None` in the constructor, which is inconsistent with the other arguments. Option 1 is chosen in this PR for that reason.

3. Should we support `earliestFinish` as well on a `VehicleType`? This would make it fully general/flexible, but it is maybe less common to have a restriction on the earliest finish of a route?
    - Note that in the issue description (https://github.com/PyVRP/PyVRP/issues/639) it is proposed to remove `latestFinish` such that the latest end of the route would be defined by `latestStart + maxDuration`. However, I would argue that `latestFinish` and `maxDuration` are two different concepts, so I propose to keep `latestFinish`. For example. a vehicle belonging to a vehicle type might start between 7-9AM, has a latest finish of 4PM and a max duration of 8 hours. If the route assigned to this vehicle takes up 8 hours (max duration not violated), then the route must start no later than 8AM to finish in time. If we would not consider `latestFinish`, then the route could have started at 9AM and ended at 5PM, while not violating the max duration.

<details>
</details>


N-Wouda:
Codecov failures are probably related to this: https://github.com/codecov/codecov-action/issues/1643. Let's ignore coverage for now until they patch the issue.

N-Wouda:
Codecov fixed itself 🎉. I'm having a look now!

pyvrp/cpp/ProblemData.cpp
```diff
@@ -233,11 +235,16 @@ ProblemData::VehicleType::VehicleType(size_t numAvailable,
     if (std::any_of(capacity.begin(), capacity.end(), isNegative<Load>))
         throw std::invalid_argument("capacity amounts must be >= 0.");
 
-    if (twEarly > twLate)
-        throw std::invalid_argument("tw_early must be <= tw_late.");
+    // Note that latestStart <= latestFinish is not required, so a route can
+    // start after its latest finish time (but this will result in time warp).
```
N-Wouda, line 239: Should we then not just require it when validating the input data?

pyvrp/cpp/DurationSegment.cpp
```diff
@@ -30,11 +30,13 @@ DurationSegment::DurationSegment(ProblemData::Client const &client)
...
-DurationSegment::DurationSegment(ProblemData::VehicleType const &vehicleType)
+DurationSegment::DurationSegment(ProblemData::VehicleType const &vehicleType,
+                                 bool const isStartSegment)
     : duration_(0),
       timeWarp_(0),
-      twEarly_(vehicleType.twEarly),
-      twLate_(vehicleType.twLate),
+      twEarly_(vehicleType.earliestStart),
+      twLate_(isStartSegment ? vehicleType.latestStart
+                             : vehicleType.latestFinish),
```
N-Wouda, line 39: #I'd replace the `isStartSegment` argument with a `Duration twLate` argument. Looking at the places where this constructor is used, we can do (for example)
```cpp
{vehicleType_, vehicleType_.latestStart};
```
instead of
```cpp
{vehicleType_, true};
```
Passing in the relevant data avoids the branch being needed here.

tests/test_DurationSegment.py
```diff
@@ -107,8 +107,12 @@ def test_OkSmall_with_time_warp(ok_small):
...
             duration=loc.service_duration if idx != 0 else 0,
             time_warp=0,
-            tw_early=loc.tw_early if idx > 0 else vehicle_type.tw_early,
-            tw_late=loc.tw_late if idx > 0 else vehicle_type.tw_late,
+            tw_early=loc.tw_early
+            if idx >= ok_small.num_depots
+            else vehicle_type.earliest_start,
+            tw_late=loc.tw_late
+            if idx >= ok_small.num_depots
+            else vehicle_type.latest_finish,
```
N-Wouda, line 115: Ternaries like these are hard to read, especially because they are directly below each other. Maybe first assign the values to variables before calling `DurationSegment()`?

examples/quick_tutorial.ipynb
N-Wouda, line 1: For the changes to the text: can we do one sentence per line? I.e., a newline after each period.

N-Wouda (commented):
This looks like a fine PR to me, but there's something that isn't clear to me yet. Can we not do the following, much smaller and non-breaking change instead?

image

It's not as pretty in terms of naming, but nothing has to break, and the `startLate` argument being optional only matters for the constructor/Python layer. We do not need to store it in the `VehicleType`, since we can do something like the following in its constructor:
```cpp
startLate(startLate.value_or(twLate)),
```

pyvrp/cpp/ProblemData.cpp
joopvdh: It depends on preference. The question is whether we should enforce the `latestStart` to be no later than the `latestFinish`. Of course, it can never be feasible to start a route after the `latestFinish/twLate`, but in that situation there will always be time warp anyway for finishing after the `latestFinish/twLate`. So, in that sense it is not needed to enforce that `latestStart <= latestFinish`. Please let me know what you prefer.

joopvdh:
@N-Wouda  I have updated this PR according to your comments. The only remaining open question is whether it should be allowed to have a `startLate` later than the `twLate` on a `VehicleType` (see one of the comments). What do you think on this topic? Would be great if you can have another look at this PR!

N-Wouda:
@joopvdh I'll have a look tomorrow morning! 🎉

pyvrp/cpp/ProblemData.h
```diff
@@ -346,9 +347,12 @@ class ProblemData
...
      *     Fixed cost of using a vehicle of this type. Default 0.
      * tw_early
-     *     Start of the vehicle type's shift. Default 0.
+     *     Earliest start of the vehicle type's shift. Default 0.
+     * start_late
+     *     Latest start of the vehicle type's shift. Default None.
      * tw_late
-     *     End of the vehicle type's shift. Unconstrained if not provided.
+     *     Latest finish of the vehicle type's shift. Unconstrained if not
+     *     provided.
```
N-Wouda, line 355: Can we revert the `tw_early` and `tw_late` docstrings? In general, for things that are not directly relevant to the PR, I try not to touch them to reduce the diff size as much as possible.

pyvrp/read.py
```diff
@@ -383,6 +383,7 @@ def _vehicle_types(self) -> list[VehicleType]:
                 # The literature specifies depot time windows. We do not have
                 # depot time windows but instead set those on the vehicles.
                 tw_early=time_windows[depot_idx][0],
+                start_late=time_windows[depot_idx][1],
```
N-Wouda, line 386: We don't need to pass this right? The default already does the right thing. So there should be no need to modify this file.

tests/search/test_Route.py
```diff
@@ -322,23 +322,29 @@ def test_str_contains_route(ok_small, locs: list[int]):
...
 def test_route_duration_access(ok_small):
     """
-    Tests access to a client's or depot's duration segment, as tracked by
-    the route.
+    Tests access to a client's or depot's duration segment, as tracked by the
+    route. The start depot's segment differs from the end depot's segment due
+    to different latest start and latest finish on the vehicle type.
     """
+    vehicle_type = ok_small.vehicle_type(0).replace(start_late=10_000)
+    ok_small = ok_small.replace(vehicle_types=[vehicle_type])
```
N-Wouda, line 330: Is this test no longer valid without changes? I think it should still work? If so, can we not modify it, and instead only add new tests like the one below?

tests/search/test_SwapRoutes.py
```diff
@@ -171,8 +171,18 @@ def test_evaluate_shift_time_window_differences(ok_small):
...
+                tw_early=10_000,
+                start_late=15_000,
+                tw_late=15_000,
+            ),
+            VehicleType(
+                capacity=[10],
+                tw_early=15_000,
+                start_late=20_000,
+                tw_late=20_000,
+            ),
```
N-Wouda, line 185: These changes seem unnecessary to me.

tests/test_Model.py
```diff
@@ -495,10 +497,10 @@ def test_model_solves_small_instance_with_shift_durations():
...
 
     for idx in range(5):
-        # Vehicles of the first type can visit two clients before having to
-        # return to the depot. The second vehicle type can be used to visit
-        # a single client before having to return to the depot. So we need
-        # at least three routes.
+        # Vehicles of the first type can visit a single client before having to
+        # return to the depot. The second vehicle type can be used to visit two
+        # clients before having to return to the depot. So we need at least
+        # three routes.
```
N-Wouda, line 503: Good catch!

tests/test_read.py
N-Wouda, line 1: If we do not update `read.py`, there should also not be a pressing need to touch `test_read.py`. I think we can revert this.

N-Wouda (commented):
This PR looks really nice. I have a few small comments that should be easy to address, and one bigger thing: `start_late` needs to be the last argument to the `VehicleType` (before `name`, which is keyword only), to ensure the change is backwards compatible.

pyvrp/cpp/ProblemData.cpp
N-Wouda: Sorry, I forgot to return to this. I think we should require `startLate <= twLate`, because:
- It comes from the user; now that the value is default optional. If it's not passed we already do the right thing.
- The user is clearly doing something nonsensical if they provide an explicit `startLate` such that `startLate > twLate`.

pyvrp/Model.py
```diff
@@ -317,6 +317,7 @@ def add_vehicle_type(
         end_depot: Depot | None = None,
         fixed_cost: int = 0,
         tw_early: int = 0,
+        start_late: int | None = None,
```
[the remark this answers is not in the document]
joopvdh: Sharp!

pyvrp/cpp/ProblemData.h
joopvdh: I do think that this change in the docstrings makes it more clear what the distinction is between `tw_early/tw_late/start_late` and how it works. But I will revert the change as you wish.

tests/search/test_Route.py
joopvdh: The test is still valid without changes. However, personally I prefer extending/modifying an existing test over adding an entire new test which is very similar to the existing test. But It is fine for me to leave the current test and create a new one for the new logic.

N-Wouda:
LGTM!

joopvdh:
Amazing! So, I suppose the PR is now ready to be merged, right?

N-Wouda:
Yes!

N-Wouda:
Thanks for the PR @joopvdh, this is a nice additional feature to have 🎉

PR #702, opened 2024-11-18
Strong index types

### LINKED ISSUE #603: Strong types for different indices
Right now we have strong types for `Duration`, `Distance`, `Load`, etc. This is useful because it avoids a number of programming mistakes at compile time.

We do not do the same for indices - those are still all `size_t`. But a location index is entirely different from a profile index, which is different from a vehicle (type) index. I don't expect this to go wrong often, but it might be nice for good measure to also define those indices as strong types, with an underlying `uint32_t` or `size_t` or similar (`uint32_t` should be plenty big I guess).

N-Wouda:
Still an experiment/WIP, because I don't know if this is possible or worth it. Let's find out.

This PR:

- [x] Closes #603.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/Measure.h
+++ b/pyvrp/cpp/Measure.h
@@ -21,10 +21,11 @@ enum class MeasureType
     DURATION,
     COST,
     LOAD,
+    INDEX,
 };
 
 // Forward declaration so we can define the relevant type aliases early.
-template <MeasureType _> class Measure;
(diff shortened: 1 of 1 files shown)


N-Wouda:
OK, so this can be done. I tinkered with it a bit yesterday and today. But it is so much work for so little gain I don't really feel like it.

PR #716, opened 2024-12-13
Route schedule statistics

### LINKED ISSUE #715: Retrieve Client Visit Times
#### Is your feature request related to a problem? 
I am solving a capacitated VRP with time windows and would like to efficiently retrieve the time when a customer is visited. This would include the time when the vehicle arrives and leaves.

A function like solution.client_information(), which would display the solution's client visit information in a e.g. format:
Route | Client | Start_visit | End_visit

Currently, I'm trying to retrieve this information through the plot_route_schedule() function, which is hard to read when multiple points are in a route and tiresome as you have to check the order in which points are visited.

#### Additional context
-

N-Wouda:
This PR:

- [x] Closes #715.

<details>

**Notes**:

In particular:

</details>


N-Wouda:
This needs to be cleaned up, added to a notebook, and I need to write tests. But I think the basic design is just about there now.

N-Wouda:
> needs to be cleaned up

For this we'll probably want to revert to what we did before #241.

pyvrp/cpp/Route.cpp
```diff
@@ -81,6 +95,26 @@ Route::Route(ProblemData const &data, Visits visits, size_t const vehicleType)
...
+
+        now += wait;
+        now -= timeWarp;
+
+        schedule_.emplace_back(
+            now, now + clientData.serviceDuration, wait, timeWarp);
+
+        now += clientData.serviceDuration;
+        prevClient = client;
+    }
```
N-Wouda, line 117: Ideally this'd be moved inside the main loop above.

N-Wouda:
@leonlan could you have a look at this? I'm not sure about naming (e.g., `VisitDatum`- ehhhh) and I don't like the extra loop inside `Route`'s constructor, but I also don't immediately see how we can get this data from the `DurationSegment` we're constructing there.

examples/basic_vrps.ipynb
leonlan, line 1: - I think it would be helpful to add a brief note here that the schedule starts at `route.start_time()` plus a print statement.
- Instead of using the shortest route, I suggest simply taking the first route explicitly.
- `plot_route_schedule` is not assuming that the route schedule starts at `route.start_time()`. It's a bit of a pain in the ass to rewrite it but it would avoid any confusion between `plot_route_schedule` and `route.schedule()` data. Otherwise... my suggestion is to remove the plot in this notebook.

pyvrp/cpp/Route.cpp
N-Wouda: Yeah. I think we can merge this into one loop but it's a bit of work figuring out how that I don't know if it's worth spending time on right now.

examples/basic_vrps.ipynb
N-Wouda: >Instead of using the shortest route, I suggest simply taking the first route explicitly.

That's a table of 30 or so visits. It's really long and clutters the notebook.

pyvrp/cpp/Route.h
```diff
@@ -18,12 +18,49 @@ namespace pyvrp
...
+     * service_duration
+     *     Duration of the service.
+     * wait_duration
+     *     If the vehicle arrives early, this is the duration it has to wait
+     *     until it can begin service.
+     * time_warp
+     *     If the vehicle arrives late, this is the duration it has to 'travel
+     *     back in time' to begin service. Non-zero time warp indicates an
+     *     infeasible route.
+     */
```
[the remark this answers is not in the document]
N-Wouda: This is true for all attributes (see e.g. Client, VehicleType). Adding a type hint here renders as
image
which is.. not great.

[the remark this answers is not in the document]
N-Wouda: I haven't yet figured out how to inline this with sphinx/autodoc/sphinx-immaterial, so I think we'll just have to settle for the ugly hints for now.

leonlan (approved):

Issue #533, opened 2024-03-18
Iterated local search


leonlan:
#### Is your feature request related to a problem? Please describe
PyVRP's current solver is a hybrid genetic search (HGS) algorithm. We have already built many components that allow us to build an iterated local search (ILS) algorithm. The latter is conceptually much simpler in my opinion, and it would be nice to see if we can build one that is as performant as HGS.

#### Describe the solution you'd like
An iterated local search algorithm that is as performant as the current hybrid genetic search algorithm.

#### Additional context
See the following papers for successful implementations of ILS:
- [Subramanian et al. (2013)](https://www.sciencedirect.com/science/article/pii/S0377221712002093) 
- [Maximó et al. (2021)](https://www.sciencedirect.com/science/article/pii/S037722172100117X)
- [Maximó et al. (2023)](https://www.sciencedirect.com/science/article/pii/S0305054822002052)

N-Wouda:
Can we give this a serious attempt? HGS has many different components that all just about work, and I am reasonably convinced we don't need all of that just to get good solutions out: it's the local search that does most of the heavy lifting.

N-Wouda:
I think this will be the big part of 0.10.0 (the features are more of a stretch goal; the algorithm update is a more important component). We can probably get going with this over the summer.

leonlan:
I have a very rough implementation of ILS at https://github.com/PyVRP/PyVRP/tree/ils. I'll try to clean it up a bit in the coming week.

leonlan:
Let's see what we can make of #660. 

1. The first thing I want to do is implement SISR, which we should expect to perform very well. This does not rely on local search (but does rely on `pyvrp.search` components) so it should be straightforward to implement.
2. While developing SISR, I want to create a benchmarking workflow (see, e.g., https://github.com/PyVRP/Bench/issues/1). This should make it easy for me to evaluate the performance and compare against HGS.
3. Once SISR performs reasonably well, I want to continue with ILS, that is, add local search. The reason why I don't want to include local search directly is because my previous experience implementing perturb + local search cycle did work well for large instances. I couldn't figure out why and changing our local search module requires substantial effort. I should also check AILS' implementation in a bit more detail to see if we're missing some important insights. Speed is definitely one thing I need to look at: I recall AILS running about 4-5x faster on X-n1001 (CVRP).

leonlan:
I think that our local search procedure is inefficient in an iterated local search when the perturbation only affects a small part of the search space.

[...]

PR #719, opened 2024-12-27
Load modified routes only in `LocalSearch::loadSolution`

leonlan:
This PR changes `LocalSearch::loadSolution` to only load routes that have been modified. This is particularly useful for #533 because ILS does not modify large parts of the solution between iterations and ILS runs many more iterations than HGS. 

Currently, loading solutions in ILS takes 15% of the total runtime for 1K client instances. This change brings that number down significantly. Furthermore, the change does not impact HGS' runtime because HGS does not run many iterations so loading solutions is not a bottleneck.

I profiled on CVRP/VRPTW 1K instances for both algorithms and here are the runtime percentages for load solution before and after change:

| Setting | Before change | After change |
|-----------|---------|--------|
| CVRP ILS | 14.2% | 1.6% (65% time spent on updating modified routes) |
| VRPTW ILS | 15.2% | 3.0% (80% time spent on updating modified routes) |
| CVRP HGS | 0.1% | 0.1% |
| VRPTW HGS | 0.1% | 0.1% |

Time spent updating is the time spent on clearing/inserting/updating the modified routes. 

<details>

**Notes**:

In particular:

</details>


leonlan:
@N-Wouda could you review this? I don't know if we should merge this in main or if I just merge it into the `ils` branch first.

N-Wouda:
> @N-Wouda could you review this? I don't know if we should merge this in main or just merge it into the `ils` branch first.

I think we can merge this into `main`. Could you profile this again? It should hopefully be a little faster now.

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -390,20 +409,64 @@ void LocalSearch::loadSolution(Solution const &solution)
...
+                             : routes.begin() + vehicleOffset[vehicleType + 1];
+
+        auto const equal = [&](auto const &route) { return route == solRoute; };
+        auto const route = std::find_if(start, end, equal);
+        if (route != end)
+        {
+            isNewRoute[idx] = false;
+            keepExistingRoute[std::distance(routes.begin(), route)] = true;
+        }
+    }
```
N-Wouda, line 434: What worries me is that this loop is quadratic in the number of vehicles. The other loops are all linear. Can we somehow make this faster too?

leonlan:
> I think we can merge this into `main`. Could you profile this again? It should hopefully be a little faster now.

- VRPTW went from 3.0% -> 2.8% (80% update).
- CVRP went from 1.6% -> 2.3% (55% update).


The new implementation is still quadratic time and with homogeneous vehicle types but it should be slightly faster esp. with heterogeneous instances. I quickly ran MDVRPTW 1K instance which went from 1.3% -> 1.1%. It seems more future-proof with larger and heterogeneous fleet so I'm OK with merging this version.

leonlan:
Superseded by #720. Closing this PR.

PR #729, opened 2025-01-20
Multi-trip vehicle routing

### LINKED ISSUE #366: Multi-trip vehicle routing

joopvdh:
This PR is work in progress, but it does already provide a working (basic) implementation of multi-trip vehicle routing. However, note that at the moment this implementation has some limitations and is not yet fully optimized for finding the best multi-trip solutions.

- [ ] Closes #366 .

Assumptions / Scope
- Load "resets" for every new trip => excess load should be checked per trip
- There is limit on the number of trips which can be performed by a vehicle (which is specified on `VehicleTy[e`)  => default 1
- A trip always starts at the `startDepot` of the `VehicleType` and ends at the `endDepot` of the `VehicleType`
- Multi-trip (`maxTrips > 1`) is only allowed when `startDepot == endDepot`
- Multi-trip in combination with non-zero client release times is not yet supported
- No loading duration (yet) between trips

Implementation details
- ProblemData: `maxTrips` attribute introduced on `VehicleType` => default 1
- Maximum number of trips is treated as a hard constraint
- Route (as part of Solution) consists of a vector of trips, where trip is a vector of clients
- `greedy_repair` and `nearest_route_insert` modified to support inserting client in new trip (if not violating maximum trips)
- Local search Route
    - Introduced `NodeType` to denote whether a node belongs to a client or a start/end depot
    - Two depot nodes/delimiters to indicate the start of a trip (`StartDepotNode`) and the end of a trip (`EndDepotNode`). To make the trips explicit.
    - `TripIdx` stored on `Node`
    - `DepotNodes` => lazily insert/remove depot nodes (alternative: to create fixed depot nodes for every possible trip in the route at the start)
    - Load caches are now on trip level: `tripLoadBefore` and `tripLoadAfter`
- Proposal
    - `StartDepotSegment` and `EndDepotSegment` to be used when constructing a Proposal which includes a newly introduced trip
    - Excess load is calculated per trip
- Local search operators
    - Modified the existing operators (`Exchange`, `SwapRoutes`, `SwapTails`, `SwapStar`) to support moves concerning multiple trips.  

Limitations / Open questions
- Initial solutions are always single-trip
- Crossover does not support multi-trip => always a single-trip solution is returned by the crossover
- Current implementation of SwapTails feels too complex
- Missing local search operators for specifically optimizing multi-trip solutions =>introduce `SwapTrips` operator?
- Should we introduce statistics on trip level (duration, num tasks planned, excess load, etc.)?
- Unit tests are not complete yet
- Plotting functionality not yet updated for multi-trip solutions

<details>

**Notes**:

In particular:

</details>


N-Wouda (commented):
I had a bit of a look at the PR just yet and while I could comment on all sorts of details, I first have a few clarifying questions and comments about the higher-level assumptions:

> * There is limit on the number of trips which can be performed by a vehicle (which is specified on `VehicleType`)  => default 1

I am not sure this a constraint we need to support. Do you have a use case where this isn't either 1 or +inf?

> * A trip always starts at the `startDepot` of the `VehicleType` and ends at the `endDepot` of the `VehicleType`

This will be difficult to reconcile with e.g. re-optimisation. It also doesn't offer an easy extension towards (multiple) reload depots (cf. #734). Perhaps we can add an optional `reload_depot` argument that determines where a reload must take place?

> * Multi-trip (`maxTrips > 1`) is only allowed when `startDepot == endDepot`

Why do we need to restrict multi trip in this way?

> * Multi-trip in combination with non-zero client release times is not yet supported

Why do we need to restrict multi trip in this way?

joopvdh:
Hi @N-Wouda, thanks for the initial look at my PR!
> > * There is limit on the number of trips which can be performed by a vehicle (which is specified on `VehicleType`)  => default 1
> 
> I am not sure this a constraint we need to support. Do you have a use case where this isn't either 1 or +inf?

Yes, I need to consider cases where the maximum number of trips on a route is limited.

> > * A trip always starts at the `startDepot` of the `VehicleType` and ends at the `endDepot` of the `VehicleType`
> 
> This will be difficult to reconcile with e.g. re-optimisation. It also doesn't offer an easy extension towards (multiple) reload depots (cf. #734). Perhaps we can add an optional `reload_depot` argument that determines where a reload must take place?

It is a matter of how flexible you want to be with supporting multiple depots in combination with multiple trips. Supporting a different start and end depot, and optionally reload depot makes the implementation more complex for moves that result in the addition or removal of a trip. My use case does not require having different different start and end depot in combination with multi-trip, so I have chosen for this basic implementation.

> > * Multi-trip (`maxTrips > 1`) is only allowed when `startDepot == endDepot`
> 
> Why do we need to restrict multi trip in this way?

See my answer above. It was not needed for my use case and it is simpler to make this assumptions when evaluating moves which affect multiple trips.

> > * Multi-trip in combination with non-zero client release times is not yet supported
> 
> Why do we need to restrict multi trip in this way?

It is not trivial to support multi-trip in combination with release times (see discussion of Wouter and Leon https://github.com/PyVRP/PyVRP/issues/366#issuecomment-1904056219). To support client release times with multi-trip is non-trivial as the latest release time should be computed and applied for each trip while currently it is computed once for the whole route. I do not have the use case to support non-zero release times in combination with multi-trip, so I left this out-of-scope for now.

N-Wouda:
Hi @joopvdh, for your use case I can understand that this implementation of multi-trip suffices, but for PyVRP as a package we will want to support a more general feature set without many of these restrictions. We probably also want to spend more time thinking about the interface (reload depots and/or max trips) and gradually merge part-by-part. This also holds for steps like changing the solution representation, the LS solution representation, and LS moves. Is that something you would be willing to undertake?

joopvdh:
Hi @N-Wouda, I understand your desire for a more comprehensive feature set for multi-trip routing in the PyVRP package. Unfortunately, I won’t have the time to work on implementing this more general version in the near future. However, I have updated my multi-trip branch to be based on the latest commit from `main` and resolved all merge conflicts. I think that this branch can serve as a good starting point for the general multi-trip implementation you have in mind. Unfortunately, as mentioned, I don't have the time to contribute to this implementation myself the coming period.

N-Wouda:
@joopvdh thanks for letting me know. I'll see what we can do in the coming few weeks!

Issue #676, opened 2024-11-06
Release 0.11.0


N-Wouda:
Stub for 0.11.0 release notes. See the [milestone](https://github.com/PyVRP/PyVRP/milestone/5) for details. Tentative release date: around February 2025.

# PyVRP 0.11.0

## New features

- Latest start constraint on `VehicleType` (in #692), as `start_late`. Defaults to `tw_late`.
- Route schedule information, available via `route.schedule()` (in #716).
- #750 (only depot time windows, service was removed again in #776)
- #739 / #366 
- 

## Breaking changes

- 
-

## Additional

- Microbenchmarks in CI (in #668).
- Expand documentation (in #675, #694).
- Penalty weights are now floating point numbers, rather than integers (in #711). This resulted in a slight improvement in solution quality.
- There is now one penalty weight for each load dimension, rather than one for all (in #713). This should result in better performance on instances with multiple load dimensions and unscaled data.
- #730
- #738
- #747
- #787 (#558)
- #783 
- #797 (#795)
- 

## Bugs fixed

- #681 (also backported and released as 0.10.1).
- #724
- #742 
-

N-Wouda:
The issues currently assigned to this release/milestone are mostly small and can (likely) be resolved quickly. As for the "big thing" of this release, I think we should focus on performance. We have added a lot of new features since our initial release two years ago, and while we have taken great care to ensure performance remains 'up to par', it has slipped a little bit. This is also evidenced by [our benchmark results](https://pyvrp.org/setup/benchmarks.html).

We can do two things:
- Improve the speed of our code;
- Improve the quality of our heuristic.

Both are fine and in scope. The former, if not too involved, is always a nice win. I have some ideas. For the latter, we could proceed by gathering a subset of benchmark instances where we perform somewhat badly, and then figuring out why that is. We have already done this in a few cases (e.g., #558), but not yet systematically across all instances.

N-Wouda:
I expect that we'll release this in a few weeks.

N-Wouda:
Before release (later this week):
- [ ] Check the tutorial. Do we explain reloading sufficiently clearly?
- [ ] Write release notes.
- [ ] Determine a candidate commit for 0.11 tag, and run benchmarks.

N-Wouda:
Candidate: https://github.com/PyVRP/PyVRP/commit/68d07577540431fe6a6fb26117e7598fcc270226. @leonlan could you run the benchmarks for this commit?

[...]

PR #731, opened 2025-01-21
Do not store intermediate cost matrices when determining average costs

### LINKED ISSUE #730: Solver crash with large distance matrices and many vehicle types
Hi,

I've run into problems when solving large VRPs (in terms of fleet size and number of clients). Even for moderately sized problems (2000 drops), the solver runs out of memory when a sufficiently large (~50 on my machine) number of vehicle types is defined. My use case is a homogeneous fleet (all vehicles have the same `unit_cost`) but follow a slightly staggered shift profile.

The issue seems to be caused by (implicitly) materializing distance and duration matrices when constructing the neighborhood and problem data.

See the accompanying PR for a potential fix.

libklein:
This PR attempts to fix a memory inefficiency in PyVRP. See #730 for details.

- [x] Closes #730.

### DIFF
--- a/pyvrp/PenaltyManager.py
+++ b/pyvrp/PenaltyManager.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from functools import reduce
 from statistics import fmean
 from warnings import warn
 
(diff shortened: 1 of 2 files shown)


N-Wouda:
Hi @libklein, thanks for opening this PR and the accompanying issue! I see I'm allowed to push to your branch so I'll make the CI pass shortly, and then merge it in.

N-Wouda:
This'll be part of 0.11, tracked in #676. I expect we'll release it in a month or so.

Issue #620, opened 2024-06-24
Rethink examples/tutorials in documentation


N-Wouda:
TODO

N-Wouda:
Cross-ref: #273

N-Wouda:
I think a general "Tutorial" section might be useful to have. We could have (relevant) examples listed there, along with explanations of different concepts and "how to do X/Y/Z with PyVRP". How do other packages solve this? I know statsmodels has, for example, a [user guide](https://www.statsmodels.org/stable/user-guide.html) and a [bunch of examples](https://www.statsmodels.org/stable/examples/index.html).

leonlan:
Here are multiple suggestions. The most one that I'm least sure of is how to organize our tutorials: should we organize it by VRP variant or by feature set?

- [ ] Turn https://github.com/PyVRP/PyVRP/issues/441 into a "Supported VRP variants" page
  - Forward references to the user guide examples

- [x] Simplify quickstart
  - Quickstart is a collection of examples that showcase the most common PyVRP use cases.
  - Include guidance to users how to navigate further: forward references to API and user guide.
  - Proposal: only keep CVRP, VRPTW and MDVRPTW. Put the other examples in tutorials (later).

- [ ] User guide consisting of: concepts and tutorials.

- [ ] Expand concepts page with:
  - Locations: Depots and Clients
  - Vehicles
  - Constraints
    - Time and duration constraints
    - Load constraints
    - Selection constraints
  - Objective function
  - Solution: (in)feasibility


- [x] Tutorials: organize by problem variant or by feature?

<details>
<summary>Sketch of outline by option</summary>

  - By problem variant: one notebook per variant (VRPTW, MDVRP, PCVRP, etc.)
    - CVRP
    - VRPTW
    - PCVRP
    - MDVRP
    - HFVRP
    - Simultaneous linehaul/backhaul VRP
    - Site-dependent VRP
    - Clustered VRP
    - etc.
  - By feature:
    - VRPs with time and duration constraints
      - Time windows
      - Vehicle shifts
      - Max duration
      - Multiple time windows
    - VRPs with loading constraints
      - Simultaneous linehaul and backhaul
      - Multiple load dimensions
    - VRPs with optional clients and client groups
      - Exclusive
      - Multiple time windows
    - VRPs with multiple depots and heterogeneous fleet
      - Multiple load dimensions
      - Routing profiles
    - VRPs with different objective functions
      - Prizes, unit distance, unit duration
[...]

PR #738, opened 2025-02-09
Add initial vehicle loads

### LINKED ISSUE #736: Initial vehicle loads

With open routes, vehicles may start or end their routes someplace other than a depot. This means that anything collected on the route is not dropped off before the shift ends, and the vehicle is not empty when the next shift starts. Once we support reloading, we need to be able to take this into account, and let vehicles start their route with an initial load already present.

Support to set the initial vehicle loads, perhaps as an `initial_load: list[int]` argument on the `VehicleType`. 

Right now the only alternative is to adjust the remaining vehicle capacity, but this will not suffice once we support #366. 

#### Additional context

Not really relevant now, but will be once we have reloading support.

N-Wouda:
This PR:

- [x] Closes #736.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/LoadSegment.cpp
+++ b/pyvrp/cpp/LoadSegment.cpp
@@ -17,6 +17,8 @@ LoadSegment::LoadSegment(ProblemData::Client const &client, size_t dimension)
 LoadSegment::LoadSegment(ProblemData::VehicleType const &vehicleType,
                          size_t dimension)
     : delivery_(0),
+      // Initial load is always a pickup quantity: it's already on the vehicle,
+      // and needs to be dropped off at a depot.
       pickup_(vehicleType.initialLoad[dimension]),
       load_(vehicleType.initialLoad[dimension])
 {
--- a/pyvrp/cpp/LoadSegment.cpp
+++ b/pyvrp/cpp/LoadSegment.cpp
@@ -13,3 +13,11 @@ LoadSegment::LoadSegment(ProblemData::Client const &client, size_t dimension)
       load_(std::max<Load>(delivery_, pickup_))
 {
 }
+
+LoadSegment::LoadSegment(ProblemData::VehicleType const &vehicleType,
+                         size_t dimension)
+    : delivery_(0),
+      pickup_(vehicleType.initialLoad[dimension]),
+      load_(vehicleType.initialLoad[dimension])
+{
+}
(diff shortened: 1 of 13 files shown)


N-Wouda:
Not so useful yet but should be soon enough.

pyvrp/cpp/LoadSegment.cpp
```diff
@@ -13,3 +13,13 @@ LoadSegment::LoadSegment(ProblemData::Client const &client, size_t dimension)
       load_(std::max<Load>(delivery_, pickup_))
 {
 }
+
+LoadSegment::LoadSegment(ProblemData::VehicleType const &vehicleType,
+                         size_t dimension)
+    : delivery_(0),
+      // Initial load is always a pickup quantity: it's already on the vehicle,
+      // and needs to be dropped off at a depot.
```
leonlan, line 21: If initial load is always a pickup, wouldn't the name `initial_pickup_load` be a better name for this? It's verbose, but this use case is relatively niche (reoptimization + pickup of goods).

pyvrp/cpp/ProblemData.cpp
leonlan, line 1: Todo: update `Model.add_vehicle_type` as well.

pyvrp/cpp/LoadSegment.cpp
leonlan: The interface is also getting more complicated with this. After #366, we should have a go at #620 and include a notebook on reoptimization and this particular use case.

pyvrp/cpp/ProblemData.h
```diff
@@ -402,6 +406,9 @@ class ProblemData
      * start_late
      *     Latest start of the vehicle type's shift. This is equal to
      *     ``tw_late`` when the latest start is not constrained.
+     * initial_load
+     *     Load quantities that are already present on the vehicle, before
+     *     accouting for client visits.
```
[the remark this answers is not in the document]
leonlan: Or see https://github.com/PyVRP/PyVRP/pull/738#discussion_r1950423418 to rename the comment.

pyvrp/cpp/ProblemData.cpp
N-Wouda: Oh yeah I forgot about that!

pyvrp/cpp/LoadSegment.cpp
N-Wouda: Agree, I don't really like the current docs/tutorial structure. We need to go back and fix that (and make it more prominent on the index page/README)

PR #739, opened 2025-02-09
Multi-trip / reloading

### LINKED ISSUE #366: Multi-trip vehicle routing

N-Wouda:
This is a stacked PR for route reloading and multi-trip. It eventually:

- [x] Closes #366.

See #366, #627, #729, #734 for background details.
Related to this PR:

- #741 
- #746
- #754
- #759 
- #761 
- #762 
- #767
- #770
- #771
- #772
- #773 
- #789
- #790 
- #794 
- #793 
- #796 
- 

NB: Most of the performance loss CodSpeed reports on is due to the time it takes to construct a `pyvrp::Route`: that involves constructing a bunch of `pyvrp::Trip` objects as well, which is a bit more expensive. But since 90% or so of our time is in the local search, I think it's OK. A benchmark should clarify.

TODO:
- [x] Release times (via #773).
- [x] Do not preallocate depot memory based on `maxReloads` (via #771).

<details>

**Notes**:

In particular:

</details>


N-Wouda:
I am taking the week of 24-28 February to work on this full time. That should hopefully suffice.

N-Wouda:
Next up is the local search routine.

N-Wouda:
#767 provides the basic functionality. The next two or so PRs add:
- A larger test case, probably from the instances in the sandbox. I suspect that'll show some bugs so this'll be good to have.
- Release time support.

N-Wouda:
Fixed most of the bugs - running larger examples now works fine. So I'm pretty sure this is functional, and the results so far show we're not that far off from optimal either. What still needs to happen is to add support for release times in a multi-trip context (#771), and then a bunch more tests to cover important missing lines. Oh, and some performance improvements.

N-Wouda:
The wheel tests suggest some platform differences. I'll have a look at those tomorrow.

N-Wouda:
OK, that seems to have done the trick.

N-Wouda:
This is functionally complete but I want to have another look at performance before we merge.

leonlan:
main (46e0df95672ace7ebacd3f85f2bc375fcad797a1):
```
 mean: 333832 (min: 333748, max: 334026)
  gap: 0.67%
iters: 48527 (min: 47932, max: 49301)
```

this branch (f10f2182babf7c29a795beebc61f66f41060431a)
```
 mean: 333914 (min: 333795, max: 334105)
  gap: 0.69%
iters: 43245 (min: 42365, max: 44649)
```

N-Wouda:
@leonlan I think this is ready. You've reviewed the individual PRs, so there's no point in looking through this again (also because it's 5K lines and basically undoable). So, instead: what do you think? Merge? Anything more we should test before we do so?

leonlan:
Let's go :shipit:!

leonlan (approved):

PR #742, opened 2025-02-14
Bugfix: SwapRoutes always evaluates to zero 

joopvdh:
This PR:

- [ ] Closes #xxxx.

**Problem**
There is a bug in the `LocalSearch::intensify` method that prevents `SwapRoutes` moves to be ever applied. The issue arises because `applyRouteOps` is invoked for each route pair `(U, V)` such that the index of route U (`rU`) is always greater than that of route V (`rV`), meaning `rU > rV` consistently holds true. When the `SwapRoutes::evaluate` function is triggered, it calls `SwapTails::evaluate`, which immediately returns a delta cost of 0 if `rU > rV`. As a result, `SwapRoutes` moves are never properly evaluated and thus never applied, as the condition `rU > rV` is always met during the execution of `LocalSearch::intensify`.

**Solution approach**

In the `LocalSearch::intensify` method only consider route pairs `(U, V)` for which holds that the index of route U (`rU`) is always smaller than that of route V (`rV`), meaning that `rU < rV` consistently holds true. This means that the `SwapRoutes` moves will be evaluated in `SwapTails::evaluate`.

Note that `ApplyRouteOps` (and therefore `SwapRoutes`) is only called when there is enough overlap between routes `U` and `V`. Whether a `SwapRoutes` move is an improving move, has no relation with the overlap between the routes, so it is questionable if this overlap restriction is in the right place. This issue is addressed in #744 .

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/tests/search/test_LocalSearch.py
+++ b/tests/search/test_LocalSearch.py
@@ -396,7 +396,8 @@ def test_intensify_can_improve_solution_further(rc208):
 
 def test_intensify_can_swap_routes(ok_small):
     """
-    Tests that the intensify method can improve a solution by swapping routes.
+    Tests that the bug identified in #742 is fixed. The intensify method should
+    be able to improve a solution by swapping routes.
     """
     rng = RandomNumberGenerator(seed=42)
 
@@ -420,11 +421,11 @@ def test_intensify_can_swap_routes(ok_small):
 
     # This solution can be improved by using the intensifying route operators
     # to swap the routes in the solution.
-    intensify_opt = ls.intensify(init_sol, cost_eval, overlap_tolerance=1)
-    intensify_cost = cost_eval.penalised_cost(intensify_opt)
+    intensify_sol = ls.intensify(init_sol, cost_eval, overlap_tolerance=1)
+    intensify_cost = cost_eval.penalised_cost(intensify_sol)
 
     assert_(intensify_cost < init_cost)
-    assert_equal(intensify_opt.excess_load(), [0])
+    assert_equal(intensify_sol.excess_load(), [0])
 
 
 def test_local_search_completes_incomplete_solutions(ok_small_prizes):
--- a/tests/search/test_LocalSearch.py
+++ b/tests/search/test_LocalSearch.py
@@ -17,6 +17,7 @@ from pyvrp.search import (
     Exchange11,
     LocalSearch,
     NeighbourhoodParams,
+    SwapRoutes,
     SwapStar,
     compute_neighbours,
 )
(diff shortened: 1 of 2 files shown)


joopvdh:
@N-Wouda I have updated this PR. This PR ensures that `SwapRoutes` moves are no longer "skipped" in `SwapTails::evaluate` and it ensures that `SwapRoutes` moves are no longer skipped if the route pair has too little overlap (in `LocalSearch::intensify`).

N-Wouda:
Hi @joopvdh, could you split the bug fix (https://github.com/PyVRP/PyVRP/pull/742/commits/09f3297f84fe6be26ef992acf7a8f1a7496a2c03) and functional change (https://github.com/PyVRP/PyVRP/pull/742/commits/05638f841e930d29f1b021b45794c1ab8e368e20) commits into two PRs? I think [splitting](https://graphite.dev/guides/how-to-move-a-commit-to-another-branch-in-git) should be relatively straightforward.

joopvdh:
Hi @N-Wouda, the change is now splitted into 2 PRs: this PR and #744 .

tests/search/test_LocalSearch.py
```diff
@@ -393,6 +394,39 @@ def test_intensify_can_improve_solution_further(rc208):
...
+    route1 = Route(data, [1, 2, 3], 0)  # Excess load: 13 - 5 = 8
+    route2 = Route(data, [4], 1)  # Excess load: 0
+    init_sol = Solution(data, [route1, route2])
+    init_cost = cost_eval.penalised_cost(init_sol)
+
+    assert_equal(init_sol.excess_load(), [8])
+
+    # This solution can be improved by using the intensifying route operators
+    # to swap the routes in the solution.
+    intensify_opt = ls.intensify(init_sol, cost_eval, overlap_tolerance=1)
```
N-Wouda, line 423:
```suggestion
    intensify_sol = ls.intensify(init_sol, cost_eval, overlap_tolerance=1)
```

tests/search/test_LocalSearch.py
```diff
@@ -393,6 +394,39 @@ def test_intensify_can_improve_solution_further(rc208):
         assert_equal(ls.intensify(intensify_opt, cost_eval), intensify_opt)
 
 
+def test_intensify_can_swap_routes(ok_small):
+    """
+    Tests that the intensify method can improve a solution by swapping routes.
```
N-Wouda, line 399: This test relates to a bug (and its fix). Could you link back to the PR in this docstring? Something like "Asserts that the bug identified in #742 is fixed." is sufficient. That's helpful if ever we need to figure out why this test existed, and how it might need to be updated.

N-Wouda:
The CodSpeed stuff seems incidental, or related to the fact that now `SwapRoutes` does meaningful work.

N-Wouda:
Thanks @joopvdh!

PR #744, opened 2025-02-18
Move check overlapping routes to SwapStar

joopvdh:
This PR:

- [ ] Closes #xxxx.

**Problem**
When the issue described in #742 is fixed, then `SwapRoutes` moves will be tried in `LocalSearch::intensify` but only when the routes to be swapped have enough "overlap" (when considering their center's angle to the center of all clients). The evaluation of `SwapRoutes` moves does not depend on the location of the clients in the route, but only on the `VehicleType` properties (fixed cost, distance cost, max distance, etc.). Therefore, SwapRoutes moves should not be skipped based on (too little) overlap between the routes. 

**Solution approach**
I made the `overlapTolerance` an attribute of `SwapStar`, since the overlap tolerance is only relevant to limit the number of `SwapStar` move evaluations. The check if a route pair has sufficient overlap, moved from `LocalSearch::intensify` to `SwapStar::evaluate`. Alternative approach would be to add the `overlapTolerance` as parameter to `LocalSearchOperator::evaluate` method, instead of adding an attribute in `SwapStar` for the overlap tolerance. However, given that the overlap tolerance is only relevant for `SwapStar`, I prefer to keep the overlap logic in `SwapStar` compared to adding the overlap tolerance to every local search operator.

<details>

**Notes**:

In particular:

</details>


tests/search/test_LocalSearch.py
joopvdh, line 1: Note to self: remove overlap tolerance from new unit test in #742 (test_intensify_can_swap_routes) when PR #742 is merged

N-Wouda:
I'll try to have a look at this tomorrow!

tests/search/test_SwapStar.py
```diff
@@ -244,7 +243,53 @@ def make_routes(data):
...
-    swap_star = SwapStar(data)
+    swap_star = SwapStar(data, 1)
     route1, route2 = make_routes(data)
     assert_equal(swap_star.evaluate(route1, route2, cost_eval), -13_243)
     assert_equal(10 * (6_220 - 5_000) + 1_043, 13_243)
+
+
+def test_swap_star_overlap_tolerance(ok_small):
+    """
+    Tests that swap star respects the overlap tolerance argument when
```
N-Wouda, line 254:
```suggestion
    Tests that SwapStar respects the overlap tolerance argument when
```

tests/search/test_SwapStar.py
```diff
@@ -244,7 +243,53 @@ def make_routes(data):
...
+    route1.update()
+
+    route2 = Route(data, idx=1, vehicle_type=1)
+    route2.append(nodes[3])
+    route2.update()
+
+    cost_eval = CostEvaluator([1_000], 0, 0)
+
+    # Overlap tolerance is zero, so no routes should have overlap and thus
+    # should the move evaluate to 0.
```
N-Wouda, line 278:
```suggestion
    # Overlap tolerance is zero, so no routes should have overlap and thus
    # the move should evaluate to 0.
```

tests/search/test_SwapStar.py
```diff
@@ -37,15 +38,13 @@ def test_swap_star_identifies_additional_moves_over_regular_swap(rc208):
     ls = LocalSearch(rc208, rng, compute_neighbours(rc208, nb_params))
 
     ls.add_node_operator(Exchange11(rc208))
-    ls.add_route_operator(SwapStar(rc208))
+    ls.add_route_operator(SwapStar(rc208, 1))
```
N-Wouda, line 41:
```suggestion
    ls.add_route_operator(SwapStar(rc208, overlap_tolerance=1))
```
(here, and other places in this file)

N-Wouda (commented):
Few small comments, nothing big. I think we can do this to simplify `LocalSearch::intensify`.

N-Wouda:
The MDVRP instance has way more vehicles and vehicle types than the others, so it makes some sense that CodSpeed shows a performance regression. This PR makes `SwapRoutes` do far more work, particularly on that instance.

joopvdh:
Hi @N-Wouda, I have rebased my changes on the latest commit of main and I have processed the review comments! The ReadTheDocs build failed, but I guess that is related to what you found in #745 ?

N-Wouda:
Yeah RTD seems to just be broken due to a Sphinx incompatibility; not something due to this PR.

PR #746, opened 2025-02-19
Add Trip class

N-Wouda:
This PR:

- [x] Is part of #739.

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/Route.h
N-Wouda, line 1: The `Trip` class should probably be in its own header. I expect it'll grow quite a bit from here, with additional member data and accessor methods.

N-Wouda:
I've moved most of the Route eval logic down to the Trip class. Tomorrow I'll integrate that into Route proper, and add the access methods needed for this to work nicely.

tests/test_Trip.py
```diff
@@ -0,0 +1,157 @@
...
+    TODO
+    """
+    pass
+
+
+def test_after():
+    """
+    TODO
+    """
+    pass
```
N-Wouda, line 157: These are for tomorrow. Once this works I think I can integrate `Trip` into the `Route` object, and after that we should be done on the user-facing side.

pyvrp/cpp/Route.h
```diff
@@ -72,7 +72,6 @@ class Route
     Duration timeWarp_ = 0;         // Total time warp on this route
     Duration travel_ = 0;           // Total *travel* duration on this route
     Duration service_ = 0;          // Total *service* duration on this route
-    Duration wait_ = 0;             // Total *waiting* duration on this route
```
N-Wouda, line 75: This was actually never used anywhere 😅

pyvrp/cpp/Route.cpp
```diff
@@ -95,8 +95,8 @@ Route::Route(ProblemData const &data, Visits visits, size_t const vehicleType)
             loadSegments[dim].load() - vehType.capacity[dim], 0));
     }
 
-    DurationSegment const depotEnd(vehType, vehType.twLate);
-    DurationSegment const vehEnd(data.location(endDepot_), 0);
+    DurationSegment const depotEnd(data.location(endDepot_), 0);
+    DurationSegment const vehEnd(vehType, vehType.twLate);
```
N-Wouda, line 99: This is wrong on current main

leonlan (approved):
LGTM!

PR #748, opened 2025-02-26
Route is determined from first Proposal segment

### LINKED ISSUE #747: `SegmentBefore` cache validity: should proposal correspond to same route?
A quick check through the codebase shows that currently it always holds that the `SegmentBefore` corresponds to the same `route` (and thus `vehicleType`) as the `proposal`, but this is not guaranteed/enforced/checked anywhere.

My suggestion is to enforce this property, since:
* the route should always start with the depot corresponding to the vehicle type, which is the `SegmentBefore` of size 1.
* if it does not hold, I think the `SegmentBefore::duration` caching (see below) is incorrect if the proposal has a `VehicleType` with a different `twEarly` or `startLate` than the `SegmentBefore`, which is not checked by the current check on `profile`.

Indeed, I think the below code currently always uses the cache as the profile check is always true (will test soon), so could be simplified. The same holds for distance b.t.w. but there it is not a problem.

https://github.com/PyVRP/PyVRP/blob/4a52bb88e9232f813d4c9802a78d97b29fd8de84/pyvrp/cpp/search/Route.h#L526-L533

N-Wouda:
This PR:

- [x] Closes #747.

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/search/Route.h
```diff
@@ -472,115 +477,112 @@ bool Route::Node::isDepot() const
...
-    auto const between = SegmentBetween(route, start, route.size() + 1);
+    auto const between = SegmentBetween(route_, start, route_.size() + 1);
     return between.distance(profile);
 }
 
 DurationSegment Route::SegmentAfter::duration(size_t profile) const
 {
-    if (profile == route.profile())
-        return route.durAfter[start];
+    if (profile == route_.profile())
```
wouterkool, line 511: Thinking about it, I think the `SegmentAfter` cache is also only valid if the `twLate` attribute is the same?

pyvrp/cpp/search/Route.h
```diff
@@ -637,9 +639,9 @@ Route::Node *Route::operator[](size_t idx)
 }
 
 template <typename... Segments>
-Route::Proposal<Segments...> Route::proposal(Segments &&...segments) const
+Route::Proposal<Segments...> Route::proposal(Segments &&...segments)
```
wouterkool, line 643: Why not just directly call the constructor but use a static method? Is that because of the templating?

pyvrp/cpp/search/Route.h
wouterkool: Maybe even `twEarly` as well as in theory a client can have a time window before the `VehicleType`'s `twEarly`, so the caches could be different based on `twEarly` but I don't think that would affect the results since ultimately the route start should be after `twEarly` and therefore arrival at the final depot also.

Given that this problem already existed I don't think it needs to be fixed now but maybe we can open a new issue? Or fix it now but probably we also want a unit test then.

wouterkool (changes requested):
LGTM except I think the `SegmentAfter::duration()` cache can also be invalid ; should we fix that as well now?

pyvrp/cpp/search/Route.h
N-Wouda: Probably. I think the only hard requirement is that a proposal's first and last segment are based on the same route.

wouterkool (approved):
LGTM now! I did not realize it also holds that the end segment must correspond to the route of the proposal, then indeed the assert is sufficient.

PR #757, opened 2025-03-12
Create Solution object in `read_solution`

leonlan:
Small QOL improvement to `read_solution` that came up when addressing https://github.com/PyVRP/Bench/pull/9#discussion_r1989469850.

<details>

**Notes**:

In particular:

</details>


pyvrp/read.py
```diff
@@ -83,24 +90,38 @@ def read(
     return builder.data()
 
 
-def read_solution(where: str | pathlib.Path) -> _Routes:
+def read_solution(where: str | pathlib.Path, data: ProblemData) -> Solution:
     """
     Reads a solution in ``VRPLIB`` format from the give file location, and
-    returns the routes contained in it.
+    returns the correspoding Solution object.
```
N-Wouda, line 96: Thinking out loud: should we make `data` the first argument? That'd be consistent with almost all other code points that take a `ProblemData` instance. But it also makes sense that `where` is the first argument to a `read()` function, so I don't really know. I guess we can keep it as is.

N-Wouda:
LGTM!

pyvrp/read.py
leonlan: I'll keep it as is!

PR #758, opened 2025-03-13
Use floats in computing proximity

leonlan:
This PR reverts the change in #737.

We have `client.tw_late = np.iinfo(np.int64).max` as default. Consequently, using `np.int64` arrays will result in overflows here:

```python
    min_wait = early[None, :] - min_duration - service[:, None] - late[:, None]
    min_tw = early[:, None] + service[:, None] + min_duration - late[None, :]
```

when `service > 0`. This results in "bad" neighbourhoods.

<details>

**Notes**:

In particular:

</details>


N-Wouda (approved):
LGTM if the tests pass 🤞

PR #761, opened 2025-03-18
Add reloading evaluation to search Route, update operators

N-Wouda:
This PR:

- [x] Is part of #739.

<details>

**Notes**:

In particular:

</details>


N-Wouda:
I'll try to finish load/export of solutions today, and then update the operators (where needed) tomorrow.

pyvrp/cpp/search/SwapTails.cpp
N-Wouda, line 1: This is not yet correct.

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -410,18 +424,43 @@ Solution LocalSearch::exportSolution() const
...
+
     for (auto const &route : routes)
     {
         if (route.empty())
             continue;
 
-        std::vector<size_t> visits;
-        visits.reserve(route.size());
+        trips.clear();
+        trips.reserve(route.numTrips());
```
leonlan, line 436: Question: why clear here instead of redefining the variable, i.e., 

```suggestion
        std::vector<Trip> trips;
        trips.reserve(route.numTrips());
```
? Idem for `visits`.

leonlan (approved):

pyvrp/cpp/search/LocalSearch.cpp
N-Wouda: Clear does not reset the vector's capacity, but redefining does. By reusing the previous trips vector many (most?) of the reserve() calls don't necessarily do anything because a previous iteration is may have already allocated sufficient memory (which we can reuse, we just need to clear it first).

N-Wouda: Consider:
```cpp
#include <iostream>
#include <vector>

int main()
{
    std::vector<int> items;
    std::cout << items.capacity() << '\n';

    items.reserve(10);
    std::cout << items.capacity() << '\n';

    items.clear();
    std::cout << items.capacity() << '\n';
}
```
with example output:
```
0
10
10
```

PR #763, opened 2025-03-21
Generalize default service times for multi-depot

leonlan:
This PR fixes an edge case in `read.py` where data sections that are expected to 2D are actually interpreted as 1D by `vrplib` if the data section has only one element per row. Encountered this when reading MTVRPTW instances with only one reload depot.

Note to self: there are basically two casts:
- Scalar -> 1D array (specification -> data section)
- 1D data section -> 2D data section (edge case, when the data section is expected to be 2D)


pyvrp/read.py
```diff
@@ -212,10 +212,14 @@ def reload_depots(self) -> list[tuple[int, ...]]:
...
-        return [
-            tuple(idx - 1 for idx in group)
-            for group in self.instance["vehicles_reload_depot"]
-        ]
+        reload_depots = self.instance["vehicles_reload_depot"]
+
+        if isinstance(reload_depots[0], Number):
+            # Some instances describe only one reload depot per vehicle, so
+            # we first cast it to a 2D array.
+            reload_depots = [[item] for item in reload_depots]
```
N-Wouda, line 220: Perhaps `np.atleast_2d`?

tests/test_read.py
```diff
@@ -539,12 +539,12 @@ def test_read_solution_multiple_vehicle_types(ok_small_multi_depot):
...
 
 
-def test_multi_trip_instance():
+def test_multi_trip_instance(ok_small_multiple_trips):
     """
     Tests that a small multi-trip instance with vehicle reload options is
     parsed correctly.
     """
-    data = read("data/OkSmallMultipleTrips.txt")
+    data = ok_small_multiple_trips
```
N-Wouda, line 547: I didn't use a fixture here, because the fixture is essentially what we're testing (does `read()` work?).

pyvrp/read.py
```diff
@@ -189,7 +189,7 @@ def service_times(self) -> np.ndarray:
             # Some instances describe a uniform service time as a single value
             # that applies to all clients.
             service_times = np.full(self.num_locations, service_times)
-            service_times[0] = 0
+            service_times[: self.num_depots] = 0
```
N-Wouda, line 192: 👍

leonlan:
@N-Wouda Can this be merged?

N-Wouda (approved):

N-Wouda:
Yeah LGTM!

Issue #503, opened 2024-03-01
Benchmark overhaul


leonlan:
This issue discusses the benchmark overhaul.

Benchmarks should reflect out-of-the-box experience with minimal changes.

- Academic stopping criterion vs. 1 min time limit vs. 1 hour time limit.	
	- Only run academic benchmark when implementing new feature.
- Single configuration file and build.

leonlan:
Data that I used for the LI post https://www.linkedin.com/feed/update/urn:li:activity:7418972756802899969/?actorCompanyId=108519916

```
        v0.12	v0.13  
200		0.08%	0.17% 
400		0.53%	0.63%
600		1.56%	1.01%  
800		3.23%	1.48%  
1000	5.08%	1.91% 
```

leonlan:
New structure for benchmarks:
- 1-minute benchmarks (comparison with other solvers)
	- CVRP
	- VRPTW
- Academic benchmarks (what we did before)

<img width="605" height="350" alt="Image" src="https://github.com/user-attachments/assets/8b7cd5aa-36fe-450e-b88e-497e09e35714" />


<img width="578" height="371" alt="Image" src="https://github.com/user-attachments/assets/051a4655-af0a-41a3-82cd-b281cf573ca2" />

leonlan:
[Hexaly, Gurobi, OR-Tools on the Capacitated Vehicle Routing Problem with Time Windows (CVRPTW).html](https://github.com/user-attachments/files/25051923/Hexaly.Gurobi.OR-Tools.on.the.Capacitated.Vehicle.Routing.Problem.with.Time.Windows.CVRPTW.html)
[Hexaly, Gurobi, OR-Tools, OptaPlanner, jsprit on the Capacitated Vehicle Routing Problem (CVRP).html](https://github.com/user-attachments/files/25051924/Hexaly.Gurobi.OR-Tools.OptaPlanner.jsprit.on.the.Capacitated.Vehicle.Routing.Problem.CVRP.html)

In case these pages get updated

PR #764, opened 2025-03-25
Replace CLI with Bench

leonlan:
This PR replaces the `pyvrp` CLI with `bench solve`. 
- The CLI is mostly an internal tool used for quick benchmarking. We remove two install dependencies (`tqdm` and `tomli`) by moving this functionality to `bench`.
- Keeping `vrplib` is handy for `read()`, which many of our tests rely on. We _could_ also make this an extra, but because it's a very lightweight and we're maintaining it ourselves, it may be better to keep it as-is.
- I also removed the `instances`, but I'm not sure if that's really necessary.

Tasks:
- [x] Part of #503.

<details>

**Notes**:

In particular:

</details>


pyproject.toml
```diff
@@ -84,6 +82,7 @@ tabulate = "^0.9.0"
 
 
 [tool.poetry.group.dev.dependencies]
+bench = {git = "ssh://git@github.com/PyVRP/Bench.git"}
```
leonlan, line 85: TODO

pyproject.toml
```diff
@@ -58,8 +58,6 @@ numpy = [
 ]
 matplotlib = ">=2.2.0"
 vrplib = "^1.4.0"
```
leonlan, line 60: Opting to keep `vrplib` as a dependency in `pyvrp`. I remember discussing having `vrplib` as extra, but it's a low-weight dependency and quite useful to have for quick benchmarking.

.gitmodules
leonlan, line 1: Do we want to remove this? I think it's pretty useful to have around for benchmarking.

N-Wouda (commented):
Few comments, but see also the questions on Slack :)

instances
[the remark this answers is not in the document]
N-Wouda: How would we run without the instances? I know `bench` has them in its repo, but I don't think those are also installed when we bring it in using `uv`, right?

PR #771, opened 2025-04-10
Fix pickup and delivery bug, remove depot memory preallocation

N-Wouda:
This PR:

- [x] Is part of #739.

<details>

**Notes**:

In particular:

</details>


N-Wouda:
I'll try to finish this tonight!

N-Wouda:
OK, I've spent some time reading the Cattaruzza paper on the train just yet. It's a bit more complicated than I had hoped, so I don't think this'll be completed this week already. No matter tho, because the main functionality is already there, so we can continue prototyping that without immediate release time support. It just needs to be there before we cut a new release to PyPI.

tests/test_Model.py
```diff
@@ -1050,3 +1050,60 @@ def test_model_solves_multi_trip_instance():
     assert_({2} in visit_sets)
     assert_({3, 4} in visit_sets)
     assert_({5, 6} in visit_sets)
+
+
+def test_model_simultaneous_pickup_delivery_with_reloading():
```
N-Wouda, line 1055: This test case is broken (probably a bug with simultaneous PD) but I don't have the time to investigate right now, so I'm keeping this around for later.

PR #773, opened 2025-04-14
Multi-trip release times per trip

N-Wouda:
This PR:

- [x] Is part of #739.

TODO:
- [x] `pyvrp::Route` general statistics;
- [x] `plotting.plot_route_schedule`;
- [x] `pyvrp::Route::schedule` information;
- [x] `pyvrp::search::Route`;
- [x] Move evaluations;

<details>

**Notes**:

In particular:

</details>


N-Wouda:
The Cattaruzza paper discusses this a bit. Release times need to be tracked per trip, and then start times can be calculated from each trip segment (in `O(num_trips)`). That's a little tricky because we then cannot easily evaluate e.g. `SegmentAfter::duration`, because that can include multiple trips, for which we only know the start times once the entire proposal is available.

I'll think about how to fix this.

N-Wouda:
I think this now works for `pyvrp::Route`. I'll need to think about how to update the relevant bits for `pyvrp::search::Route` and the local search moves.

N-Wouda:
`finalise()` works pretty well, but I've noticed I need to differentiate between left-to-right finalise (forwards, along the route), and right-to-left (backwards). For `LoadSegment`, those two concepts coincide - excess load does not care about forwards or backwards evaluation. But time/duration does, and it matters for e.g. the earliest and latest start on the `DurationSegment`.

Which 'side' to finalise - left or right - coincides with whether we're appending segments to the finalised segment's front or back. Since front/back is the relevant C++ terminology as well, I'll probably call these functions "finalise front" and "finalise back".

pyvrp/cpp/DurationSegment.cpp
```diff
@@ -1,10 +1,46 @@
 #include "DurationSegment.h"
 
-#include <cassert>
+#include <fstream>
 
 using pyvrp::Duration;
 using pyvrp::DurationSegment;
 
+DurationSegment DurationSegment::finaliseBack() const
```
N-Wouda, line 8: I think this now works, but need to add way more tests to be sure.

pyvrp/cpp/search/Route.cpp
```diff
@@ -279,23 +279,33 @@ void Route::update()
...
-                                     durBefore[idx - 1],
-                                     durAt[idx]);
+    {
+        auto const prev = idx - 1;
+        auto const before = nodes[prev]->isReloadDepot()
+                                ? durBefore[prev].finaliseBack()
+                                : durBefore[prev];
+
+        auto const edgeDur = durations(visits[prev], visits[idx]);
+        durBefore[idx] = DurationSegment::merge(edgeDur, before, durAt[idx]);
```
N-Wouda, line 294: This does not yet work, because the depot's service duration is counted as part of the depot return at the end of the trip. I don't have a clean fix for this:
- We could duplicate all reload depot nodes into start/end nodes, and have the end nodes be free of service. This is similar to how we deal with the route's start and end depots. I tried to implement this, but it gets tedious quick. Lots of bookkeeping that adds complexity, and is error prone.
- We can hack something together at each point where this matters. Not pretty and also fairly error prone.
- We can remove depot service duration. It's unreleased (added in #750, which is not yet on PyPI), and the same effect can be obtained by modifying the duration matrix. Depot time windows already provide everything needed to model visit constraints, so there is no *hard* need for service duration.

I'm leaning towards removing the `Depot::serviceDuration` field.

N-Wouda:
I think this is now completely functional. All it still needs are some higher-level tests involving `Model` and an instance with multi-trip and release times.

N-Wouda:
Ok there's something not quite right yet in the proposal. I'll see if I can figure that out once I have the new test instance/fixture completed - see https://github.com/PyVRP/Sandbox/pull/14/.

tests/test_Model.py
```diff
@@ -1050,3 +1050,16 @@ def test_model_solves_multi_trip_instance():
...
+    """
+    TODO
+    """
+    m = Model.from_data(mtvrptw_release_times)
+    res = m.solve(stop=MaxIterations(50))
+    print(res)
+
+    # OPT is 1068.7 - see table EC.9, p. ec25 of the online supplement
+    # (here: https://doi.org/10.1287/trsc.2022.1161).
+    assert False
```
N-Wouda, line 1065: This triggers an assert (cost after move != cost before + delta). I'll see if I can fix that tomorrow.

N-Wouda: I have distilled a test case. The problem's in `TripRelocate`, which makes sense.

N-Wouda:
@leonlan I think this is now complete. There's a lot of stuff here, not all of which relates directly to release times (a lot has to do with improving debugging too, and there's some general improvements to multi-trip as well).

The main thing that's seen a significant expansion is the `DurationSegment`. Could you have a look at that in particular?

N-Wouda:
Some functions are a bit slower, but on the whole I think we're going to be OK:

image

docs/source/dev/supported_vrplib_fields.rst
leonlan, line 1: My internet is loading too slow so I'm adding a question here instead but it should be at `Route.cpp`: why merge segments back-to-front instead of front-to-back? Is it because `finaliseFront` is faster than `finaliseBack`?

pyvrp/cpp/Route.cpp
```diff
@@ -221,49 +215,68 @@ Route::Route(ProblemData const &data, Trips trips, size_t vehType)
...
 
             delivery_[dim] += ls.delivery();
             pickup_[dim] += ls.pickup();
             excessLoad_[dim] += ls.excessLoad(vehData.capacity[dim]);
         }
+    }
 
-        size_t prevClient = trip.startDepot();
-        for (auto const client : trip)
+    // Duration statistics. We iterate in reverse, that is, from the last to
```
leonlan, line 245: Question: why merge segments back-to-front instead of front-to-back? Is it because `finaliseFront` is faster than `finaliseBack`?

N-Wouda: Yes, and it's a bit easier at the start  - there's some statistics like `slack` and startTime that we track about the first trip/start of the route, that now no longer need a special check, we can just get them from the complete duration segment.

tests/conftest.py
```diff
@@ -108,6 +108,15 @@ def gtsp():
     return read("data/50pr439.gtsp", round_func="round")
 
 
+@pytest.fixture(scope="session")
+def mtvrptw_release_times():
+    """
+    Fixture that returns a multi-trip instance with time windows and release
+    times, of 70 clients.
+    """
+    return read("data/C201R0.25.vrp", round_func="dimacs")
```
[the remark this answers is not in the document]
N-Wouda: I think we can get away with this for tests, but we should make sure for benchmarking what's actually needed here.

pyvrp/cpp/DurationSegment.h
```diff
@@ -157,35 +214,121 @@ DurationSegment::merge([[maybe_unused]] Duration const edgeDuration,
             first.timeWarp_ + second.timeWarp_ + diffTw,
             std::max(second.twEarly_ - atSecond, first.twEarly_) - diffWait,
             std::min(secondLate, first.twLate_) + diffTw,
```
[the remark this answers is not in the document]
N-Wouda: Agree, thanks

PR #663, opened 2024-10-23
Support Python 3.13, drop 3.9

N-Wouda:
This PR:

- [x] Closes #662.
- [ ] Adds and passes relevant tests.
- [ ] Has well-formatted code and documentation.

<details>

**Notes**:

Please read our [contributing guidelines](https://pyvrp.org/dev/contributing.html) first.
In particular:

- You must add tests when making code changes.
  This keeps the code coverage level up, and helps ensure the changes work as intended.
- When fixing a bug, you must add a test that would produce the bug in the master branch, and then show that it is fixed with the new code. 
- New code additions must be well formatted. Changes should pass the pre-commit workflow, which you can set up locally using [pre-commit](https://pre-commit.com/#intro). 
- Docstring additions must render correctly, including escapes and LaTeX.
- Finally, it is essential that all contributions in this PR are license-compatible with PyVRP's MIT license.
  Please check that this PR can be included into PyVRP under the MIT license.

</details>

PR #775, opened 2025-04-25
chore(deps): unrestrict Python upper constraint

alessio-locatelli:
Libraries must not restrict the upper bound unless there is a known incompatibility. See the related PR:
https://github.com/PyVRP/PyVRP/pull/663#discussion_r2059892131

---

Running `poetry lock` did not change the lock. 

Also, I believe that adding 3.14 to the CI workflow is a bit too soon and out of scope of this PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ classifiers = [
 
 
 [tool.poetry.dependencies]
-python = ">=3.10,<4"  # NOTE: Cannot write without the upper bound because 'docblock' defines it.
+python = ">=3.10,<4.0"
 numpy = [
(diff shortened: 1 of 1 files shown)


N-Wouda:
LGTM, thanks @alessio-locatelli. Let me see if I can update `docblock`.

N-Wouda:
OK, the issue with that is that poetry inserts it automatically. We need to switch to `project.requires-python` in our specification, which requires Poetry 2, and a whole host of other changes. Since we're considering switching to `uv` instead (#659), I do not want to go through the trouble of upgrading. I can live with `python <4`.

N-Wouda:
Thanks @alessio-locatelli!

Issue #623, opened 2024-06-26
Switch back to doubles


N-Wouda:
#### Is your feature request related to a problem? Please describe

Modelling fractional costs (e.g., 21ct/km, 19.95 EUR/h) gets out of hand quick with all integer data. We have already seen that before, and this was one of the main motivators for 64-bit integers in #468. However, in #550 we need to do weird things to account for this again. Specifying fractional costs is just fundamentally awkward in integers.

Additionally, integers overflow. This has already caused at least one bug (e.g., #590) and makes the code more brittle because we need to work around the issue (e.g., #599, #586). Further, the lack of builtin support for infinity means we need to use a large `MAX_VALUE` constant instead to avoid the overflow problem - with limited success (e.g., fails in #617, #550).

Floats have their own problems. For example, equality checks between floats needs some tolerance that is [a bit iffy](https://stackoverflow.com/q/17333/4316405).

#### Describe the solution you'd like

Switch costs, distances, durations etc. back over to float or double, rather than the integral values we impose now. I want to revert #518, re-instate #47, and probably remove integer support instead.

#### Describe alternatives you have considered

This will remain awkward in any other way, so there aren't really any. 64-bit integers give us a lot of flexibility already, but without hardware support for infinity (or NaN) and always the risk of overflows it does not seem to be worth it.

#### Additional context

Most of the double vs int discussions previously focused on performance (ints were a bit faster). Most of that was likely due to us using 32-bit integers, which are half the size of a double. We now use 64-bit integers since we need the extra space for modelling (see #468), so that should essentially remove the performance issue.

N-Wouda:
A question is whether we need `double` (=float64) or `float` (=float32). 32 bits might be enough, but I do not know that for sure. We could experiment to find out?

leonlan:
I'm in favor of having floats because it improves user experience by not having to scale all the time. I notice this in the benchmarks, especially since we now have unit distance costs (#550). This requires scaling twice and often results in overflows.

What are the other downsides of switching back to floats? I recall from somewhere that we get about 25% less iterations. That's a significant number of iterations, but we can probably get that performance back from somewhere else.

N-Wouda:
> What are the other downsides of switching back to floats? I recall from somewhere that we get about 25% less iterations. That's a significant number of iterations, but we can probably get that performance back from somewhere else.

[...]

Issue #558, opened 2024-03-30
Improve performance on HFVRP instances


leonlan:
Benchmark runs on HFVRP instances show a lot of variation in performance: some instances have gaps between 2-7%. @wouterkool also told me once that he finds very high gaps on HFVRP-type instances.

## Results
- See [gaps.pdf](https://github.com/PyVRP/PyVRP/files/14812517/gaps.pdf) for boxplots of gaps per instance.
- See table below for the top 10 instances with highest coefficient of variation on the gap, along with the gap means.

| name       |   ('gap', 'cov') |   ('gap', 'mean') |
|:-----------|-----------------:|------------------:|
| X134-FSMD  |           137.44 |              1.56 |
| X685-FSMD  |            56.71 |              3.04 |
| X233-FSMD  |            52.36 |              2.58 |
| X294-HD    |            36.06 |              1.06 |
| X204-FSMD  |            33    |              5.36 |
| X336-FSMF  |            30.83 |              1.6  |
| X101-FSMFD |            29.1  |              2.28 |
| X186-FSMD  |            28.48 |              2.49 |
| X819-FSMD  |            28.47 |              3.81 |
| X327-FSMFD |            20.1  |              1.1  |

- Benchmark runs with and without SwapRoutes show no difference.

## Discussion
- There's a lot of variation.
- The gaps are very high (and low) on some instances; I have checked that the costs are computed correctly.
- There's a 0.50 correlation between cost gaps and route gaps, meaning having more routes correlates with higher costs.

## Action points

- [ ] **Improve route minimization**: Most of the high gaps and high variance is observed on "F" type instances, that is, instances with unlimited fleet.
- [x] **Make SwapRoutes actually do something**: I don't know if this will help, but it surprises me that it doesn't help to improve performance. This was also identified earlier on MDVRPTW instances.

N-Wouda:
> **Make SwapRoutes actually do something**: I don't know if this will help, but it surprises me that it doesn't help to improve performance. This was also identified earlier on MDVRPTW instances.

Or remove it?

leonlan:
SwapRoutes as it stands now could be removed. But I don't know if it's the "intensification" framework that is limiting its effectiveness or if SwapRoutes just, in general, doesn't work. I should test that first.

Another solution from the literature is to do something like a local search move + reassignment. This excerpt is from [Penna et al. (2019)](https://link.springer.com/article/10.1007/s10479-017-2642-9):

[...]

PR #777, opened 2025-05-05
Read HFVRP instances

### LINKED ISSUE #543: Add support for mixed distance and duration objectives

leonlan:
This PR:

- [x] Closes #543.

<details>

**Notes**:

In particular:

</details>


leonlan:
@N-Wouda Could you check this? I finally got to work on this again. A quick recollection of what put this on-hold:

- Support for HFVRP was implemented in v0.9.0.
- Performance was OK (1% compared to BKS) although performance could be further improved (see #558).
- We attempted to add support for VRPLIB instances, but it's was hard to incorporate unit distance costs (floats) because there is a multiplicative scaling effect when we round & scale distances _and_ unit distance cost. 
- We then opted to support doubles again (#623), but this was also put on hold.
- I finally had the time to revisit this again and thought of a solution to the double scaling problem. Unit distance costs can be nicely incorporated in routing profiles as long as we do not round/scale them. I think this solution is OK for benchmarking purposes since unit distances are niche.

leonlan:
As discussed, I reverted the changes in which unit distance cost where incorporated into the routing profile. I modified the instances to multiply unit distance and fixed cost by 100, while keeping unit distance cost unrounded/unscaled. The instances in Sandbox are updated accordingly (https://github.com/PyVRP/Sandbox/pull/6).

@N-Wouda could you take a look at this?

pyvrp/read.py
```diff
@@ -270,6 +270,20 @@ def max_durations(self) -> np.ndarray:
...
         return self.round_func(max_durations)
 
+    def fixed_costs(self) -> np.ndarray:
+        if "vehicles_fixed_cost" not in self.instance:
+            return np.zeros(self.num_vehicles, dtype=np.int64)
+
+        return self.round_func(self.instance["vehicles_fixed_cost"])
+
+    def unit_distance_costs(self) -> np.ndarray:
+        if "vehicles_unit_distance_cost" not in self.instance:
```
N-Wouda, line 280: I think we should also add these new fields to the docs, right?

N-Wouda (commented):
LGTM, but the new fields are missing from the docs. If we can add those then I think this is good to go!

N-Wouda:
Sorry, there were still a few old review comments from a few commits ago that I hadn't posted. Those aren't relevant anymore, so I deleted them.

PR #781, opened 2025-05-08
[Draft] Create real-world example

alessio-locatelli:
Related to https://github.com/PyVRP/PyVRP/issues/620

:warning: You can read the `README.md` for the task description, but  the code is not ready.

:information_source:  Sorry, I triggered CI by mistake. I always skip CI until ready.

#### TO-DO

- [ ] Create the model. How to define a single traveler, but different transport between the nodes?
- [ ] Solve.
- [ ] Map the solver results back to the real transport.

### DIFF
--- a/examples/real_world_travel_example/README.md
+++ b/examples/real_world_travel_example/README.md
@@ -1,0 +1,16 @@
+# Summary
+
+In this tutorial, we formulate and solve a Prize-Collecting Vehicle Routing Problem (VRP) for a traveler who has between 2 and 7 days to visit popular European cities and must return to a depot. We use PyVRP to model the problem with integer costs (durations in minutes, distances in meters, prices in EUR), two parallel edges (fast‐expensive vs. slow‐cheap trains) between each city pair, and integer prizes for visiting each city. We walk through (1) data setup, (2) PyVRP model creation, (3) solving with PyVRP’s heuristic solver, and (4) mapping the solution back to real‐world city names and chosen trains.
+
+# Problem Description
+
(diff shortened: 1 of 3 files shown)


leonlan:
Hi @alessio-locatelli, I’m not sure why you’d want to use PyVRP for your problem. Although it resembles a routing problem, the multi-mode travelling options are not common to pure routing problems. It looks like your problem can be easily modeled as MILP or CP and solved   using HiGHS or CP-SAT, respectively.

alessio-locatelli:
> Hi @alessio-locatelli, I’m not sure why you’d want to use PyVRP for your problem. Although it resembles a routing problem, the multiple, parallel edge options are not common in pure routing problem unless its different routing profiles. It does look like your problem can be easily modeled as MILP or CP and solved using the HiGHS or CP-SAT solver, respectively.

- While I have years of commercial experience, I am still new to Linear Programming. I am aware that the chosen approach may be misguided.
- I am not sure if that is hard to implement or not, but given a single profile, PyVRP could expose the chosen edges in the solution, which would allow mapping the edges from the solution to the specific edges between two nodes.
- Thank you, I will focus more on MiniZinc and OR-Tools.

N-Wouda:
> * I am not sure if that is hard to implement or not in the PyVRP solver, but given a single profile and parallel edges, PyVRP could expose the chosen edges in the solution, which would allow mapping the edges from the solution to the specific edges between two nodes.

It took me a minute to parse this, but I think I understand the source of confusion here. PyVRP does not support multigraphs. Reading "a single profile and parallel edges", I think you mean calling `m.add_edge()` repeatedly with the same `frm`, `to`, and `profile` arguments? If you look at how [edges are transformed into matrices](https://github.com/PyVRP/PyVRP/blob/f25378cd8f86deac40371d8a5413cc52a81cb9b2/pyvrp/Model.py#L401), you'll see that effectively only the last constructed edge gets used in that case.

I have added a note about this to the documentation.

PR #791, opened 2025-05-13
Add delta cost shortcuts based on proposal size and duration

### LINKED ISSUE #49: Pre-checks with nonzero timewarps
Pre-checks should be improved to work for nonzero timewarp routes as well. Assuming that all time warp gets resolved, there should be an improvement, otherwise we do not need to fully check the move as it can never be improving. This case is currently ignored.

N-Wouda:
This PR:

- [x] Closes #49 (see there for details).

TODO:
- [x] Segments need to track their size
- [x] We then need to determine if the proposal consists of segments all from the same route, and if the proposed size is smaller than that of the current route
- [x] If that's the case it's a pure remove. We can then shortcut time or distance calculations if the objective allows it.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/CostEvaluator.h
+++ b/pyvrp/cpp/CostEvaluator.h
@@ -279,16 +279,14 @@ bool CostEvaluator::deltaCost(Cost &out, T<Args...> const &proposal) const
                 proposal.loadSegment(dim).load(), capacity[dim], dim);
     }
 
-    auto const hasDurationCost = route->unitDurationCost() != 0;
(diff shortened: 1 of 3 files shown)


N-Wouda:
I'm just experimenting a bit with some ideas. I don't know if they're all valid, but as long as performance doesn't improve that question is somewhat moot.

N-Wouda:
Yeah this is going nowhere. Nice attempt, but there's not much here it seems.

PR #807, opened 2025-05-18
Remove problem type compile flag

### LINKED ISSUE #502: Get rid of problem type compile flag
We currently have a `--problem` compile flag that allows us to compile with TW support (`vrptw`) or without TW support (`cvrp`). The reason why we have this is because of performance: see https://github.com/PyVRP/PyVRP/issues/48. 

I recall that we get about 25% more iterations on the CVRP benchmark when we compile without TW support, which is about 0.10% gap or so difference.

We should get rid of this because:
- Compile flags are not user-friendly; users have to build PyVRP manually to get the desired version of PyVRP. In the same vein, see #491.
- Since most users install PyVRP directly from PyPI and use a default build, our benchmarks are somewhat unfair because we show them the results from build they are not using. We should therefore also evaluate PyVRP on the default build. Related: #503.

N-Wouda:
This PR:

- [x] Closes #502.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/buildtools/build_extensions.py
+++ b/buildtools/build_extensions.py
@@ -22,12 +22,6 @@ def parse_args():
         help="The type of build to provide. Defaults to release mode.",
     )
     parser.add_argument(
-        "--problem",
-        default="vrptw",
-        choices=["cvrp", "vrptw"],
-        help="Which type of solver to compile. Defaults to 'vrptw'.",
-    )
-    parser.add_argument(
         "--clean",
         action="store_true",
         help="Clean build and installation directories before building.",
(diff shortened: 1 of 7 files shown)


N-Wouda:
Removing for now. #555 may reintroduce some special-casing if desired. I can imagine skipping `Route::Proposal::duration` can be very beneficial.

N-Wouda:
I'll continue with this later.

leonlan (approved):

leonlan:
> @leonlan I think this is OK as-is. We might want to add a note to the benchmark results table for 0.12 about this, if need be. But I think ILS will not result in worse gaps anyway, and then it's kind of moot.

I don't think it's needed, also in view of #503.

PR #778, opened 2025-05-05
Iterated local search

leonlan:
This PR:

- [x] Closes #533.
- [ ] Adds and passes relevant tests.
- [ ] Has well-formatted code and documentation.

<details>

**Notes**:

Please read our [contributing guidelines](https://pyvrp.org/dev/contributing.html) first.
In particular:

- You must add tests when making code changes.
  This keeps the code coverage level up, and helps ensure the changes work as intended.
- When fixing a bug, you must add a test that would produce the bug in the master branch, and then show that it is fixed with the new code. 
- New code additions must be well formatted. Changes should pass the pre-commit workflow, which you can set up locally using [pre-commit](https://pre-commit.com/#intro). 
- Docstring additions must render correctly, including escapes and LaTeX.
- Finally, it is essential that all contributions in this PR are license-compatible with PyVRP's MIT license.
  Please check that this PR can be included into PyVRP under the MIT license.

</details>

PR #810, opened 2025-05-22
Acceptance criteria

leonlan:
This PR:

- [x] Is part of #778.

<details>

**Notes**:

In particular:

</details>


pyvrp/accept/MovingAverageThreshold.py
leonlan, line 1: Some background:
- I took MAT from [ALNS](https://github.com/N-Wouda/ALNS/tree/master).
- I rewrote the history into a numpy array. The original `deque` implementation took up to 20% runtime, now it's at most 1%.
- The most important additions are the `max_runtime` and `max_iterations` parameters. We've never bothered too much with this in ALNS (only `autofit()` method), but these convergence effects are significant for performance. I use only `max_runtime` in my ILS testing, but `max_iterations` should be useful for tests where we use iteration-based stopping criteria.

pyvrp/accept/MovingAverageThreshold.py
```diff
@@ -0,0 +1,127 @@
...
+           routing problem, *European Journal of Operational Research* 294 (3):
+           1108 - 1119.
+    """
+
+    def __init__(
+        self,
+        eta: float,
+        gamma: int,
+        max_runtime: float = float("inf"),
+        max_iterations: float = float("inf"),
```
N-Wouda, line 68: This is inherently integral; why a float?

pyvrp/accept/MovingAverageThreshold.py
```diff
@@ -0,0 +1,127 @@
...
+    recently observed solution. The recently observed solutions are stored in
+    a ``history`` attributed of size at most :math:`\\gamma`.
+
+    The algorithm incorporates a decreasing factor :math:`\\tilde{\\eta}` that
+    converges to zero as the search approaches its maximum time limit or
+    iterations. It is calculated as:
+
+    .. math::
+        \\tilde{\\eta} = \\eta \\times \\min\\left( 1 - \\frac{t}{T},
+            1 - \\frac{i}{I} \\right)
```
N-Wouda, line 31: Where does this happen? I don't see it in the code.

pyvrp/accept/MovingAverageThreshold.py
```diff
@@ -0,0 +1,127 @@
...
+
+        recent_best = history.min()
+        recent_avg = history.mean()
+
+        runtime = min(perf_counter() - self._start_time, self.max_runtime)
+        pct_runtime = runtime / self.max_runtime if self.max_runtime else 1
+        pct_iters = (
+            self._iters / self.max_iterations if self.max_iterations else 1
+        )
+        factor = self.eta * min(1 - pct_runtime, 1 - pct_iters)
```
N-Wouda, line 123: Should this update $\eta$?

pyvrp/accept/MovingAverageThreshold.py
```diff
@@ -0,0 +1,127 @@
...
+
+        self._iters += 1
+
+        return candidate <= recent_best + factor * (recent_avg - recent_best)
```
N-Wouda, line 127: OK, so, for me:
- We check if `candidate` is better than a weighted average of the best solution and the average candidate solutions in recent history. We weigh the best as $1 - \eta$, and the average as $\eta$.
- $\eta$ is based on an initial value $\eta_0$, multiplied by the nearest of time and iteration convergence. So $\eta \rightarrow 0$. Basically, the closer we are to stopping, the smaller the weight on the average cost. So we converge towards pure improvement.

N-Wouda (commented):
I have questions 😆

pyvrp/accept/MovingAverageThreshold.py
leonlan: Yea that's right, I wanted to make use of `float('inf')` like the CLI, so I type-hinted it with `float`. I think rewriting with None's should also do the trick.

pyvrp/accept/MovingAverageThreshold.py
```diff
@@ -0,0 +1,151 @@
...
+    where :math:`s^*` is the best solution observed in the last :math:`N`
+    iterations, :math:`f(\cdot)` is the objective function,
+    :math:`N \in \mathbb{N}` is the history length parameter, and each
+    :math:`s^j` is a recently observed solution.
+
+    The dynamic weight :math:`w` converges to zero as the search approaches
+    its maximum runtime or iterations. It is calculated as:
+
+    .. math::
+        w = w_0 \times \min\left(1 - \frac{t}{T}, 1 - \frac{i}{I} \right)
```
N-Wouda, line 26:
```suggestion
    .. math::

       w = w_0 \times \min\left(1 - \frac{t}{T}, 1 - \frac{i}{I} \right)
```

pyvrp/accept/MovingAverageThreshold.py
```diff
@@ -0,0 +1,151 @@
...
+
+    where :math:`w_0 \ge 0` is the initial weight parameter, :math:`T \ge 0`
+    and :math:`I \ge 0` are the maximum runtime and iterations parameters,
+    and :math:`t` and :math:`i` are the elapsed runtime and number of
+    iterations. The dynamic weight uses whichever limit (runtime or iterations)
+    is most restrictive.
+
+    Note: The parameters ``weight`` and ``history_length`` correspond to
+    :math:`\eta` and :math:`\gamma` respectively in [1]_.
```
N-Wouda, line 35:
```suggestion
    .. note::
    
       The parameters ``weight`` and ``history_length`` correspond to
       :math:`\eta` and :math:`\gamma` respectively in [1]_.
```
maybe?

pyvrp/accept/MovingAverageThreshold.py
```diff
@@ -0,0 +1,151 @@
...
+    @property
+    def max_runtime(self) -> float | None:
+        return self._max_runtime
+
+    @property
+    def max_iterations(self) -> float | None:
+        return self._max_iterations
+
+    @property
+    def _runtime_budget(self) -> float:
```
N-Wouda, line 108: There's no fixed definition for what is and what is not a property, but my heuristic is:
- It should be constant-ish. If I repeatedly access a property without changing anything else, I should get the same result.
- Accessing a property should not do meaningful work. It's basically equivalent to looking up some data on an object.

This property somewhat violates both those 'rules'. Perhaps it should just be a regular method. For good measure, we should then probably also update `_iteration_budget` to match.

pyvrp/accept/MovingAverageThreshold.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        if self.max_iterations is None:
+            return 1
+
+        if self.max_iterations == 0:
+            return 0
+
+        return 1 - self._iters / self.max_iterations
+
+    def __call__(self, best: float, current: float, candidate: float) -> bool:
```
N-Wouda, line 136: This gets called quite a bit, right? For performance, it's then probably best to go through the variables directly, where possible, rather than their property members. So `self._history_length` rather than `self.history_length`, and `self._weight`.

pyvrp/accept/MovingAverageThreshold.py
```diff
@@ -0,0 +1,151 @@
...
+
+       f(s^*) + w \times \left(\sum_{j = 1}^N \frac{f(s^j)}{N} - f(s^*) \right)
+
+    where :math:`s^*` is the best solution observed in the last :math:`N`
+    iterations, :math:`f(\cdot)` is the objective function,
+    :math:`N \in \mathbb{N}` is the history length parameter, and each
+    :math:`s^j` is a recently observed solution.
+
+    The dynamic weight :math:`w` converges to zero as the search approaches
+    its maximum runtime or iterations. It is calculated as:
```
N-Wouda, line 23:
```suggestion
    The dynamic weight :math:`w` converges to zero as the search approaches
    its maximum runtime or number of iterations. It is calculated as:
```

pyvrp/accept/MovingAverageThreshold.py
```diff
@@ -0,0 +1,151 @@
...
+        Maximum number of iterations. As the search approaches this limit,
+        :math:`w \to 0`. Must be non-negative. Default is ``None``, meaning
+        that :math:`w` stays equal to :math:`w_0`.
+
+    References
+    ----------
+    .. [1] Máximo, V.R. and M.C.V. Nascimento. 2021. A hybrid adaptive iterated
+           local search with diversification control to the capacitated vehicle
+           routing problem, *European Journal of Operational Research* 294 (3):
+           1108 - 1119.
```
N-Wouda, line 60: Perhaps include DOI?

PR #811, opened 2025-05-22
Introduce from_json method

braktar:
Hi, the project is great !

Here is a first contribution to ease the integration of PyVRP. The idea to define a ProblemData directly from a parsed JSON.
Tell me if this a relevant addition. Moreover if you could point out the elements that could be improved (architecture, doc, etc.) I'd be very grateful.

This PR:

Best,

Gwénaël

### DIFF
--- a/pyvrp/read.py
+++ b/pyvrp/read.py
@@ -552,3 +552,40 @@ class _ProblemDataBuilder:
                 allowed_clients2profile_idx[clients] = next(profile_idx)
 
         return allowed_clients2profile_idx
+
+
+def _problem_data_from_json(cls, data: dict) -> ProblemData:
+    """
+    Creates a :class:`~pyvrp._pyvrp.ProblemData` instance from a JSON-like
+    dictionary.
+
+    Parameters
+    ----------
+    data : dict
+        Dictionary containing all fields required to instantiate a ProblemData
+        object. The structure must match the ProblemData constructor signature
+        (clients, depots, vehicle_types, distance_matrices, duration_matrices,
+        groups).
+
+    Returns
+    -------
+    ProblemData
+        Data instance constructed from the provided dictionary.
+    """
+    clients = [Client(**client) for client in data["clients"]]
+    depots = [Depot(**depot) for depot in data["depots"]]
+    vehicle_types = [VehicleType(**vt) for vt in data["vehicle_types"]]
+    distance_matrices = [np.array(mat) for mat in data["distance_matrices"]]
+    duration_matrices = [np.array(mat) for mat in data["duration_matrices"]]
+    groups = [ClientGroup(**group) for group in data.get("groups", [])]
+    return ProblemData(
+        clients=clients,
+        depots=depots,
+        vehicle_types=vehicle_types,
+        distance_matrices=distance_matrices,
+        duration_matrices=duration_matrices,
+        groups=groups,
+    )
+
+
+setattr(ProblemData, "from_json", classmethod(_problem_data_from_json))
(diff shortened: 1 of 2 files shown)


N-Wouda:
Hi @braktar! Thanks for this PR, I'll have a look later today!

pyvrp/read.py
```diff
@@ -552,3 +552,40 @@ def _allowed2profile(self) -> dict[tuple[int, ...], int]:
...
+        clients=clients,
+        depots=depots,
+        vehicle_types=vehicle_types,
+        distance_matrices=distance_matrices,
+        duration_matrices=duration_matrices,
+        groups=groups,
+    )
+
+
+setattr(ProblemData, "from_json", classmethod(_problem_data_from_json))
```
N-Wouda, line 591: I struggle with this monkey patch. I'd rather do one of the following:
- Add the method to the `ProblemData` class, on the C++ side. This'd probably require us to bring in e.g. https://github.com/nlohmann/json to parse a JSON-formatted string.
- Make `from_json` a standalone function. But then I do not immediately know where it needs to go in the PyVRP codebase (should it?).

pyvrp/read.py
```diff
@@ -552,3 +552,40 @@ def _allowed2profile(self) -> dict[tuple[int, ...], int]:
                 allowed_clients2profile_idx[clients] = next(profile_idx)
 
         return allowed_clients2profile_idx
+
+
+def _problem_data_from_json(cls, data: dict) -> ProblemData:
```
N-Wouda, line 557: This doesn't do what it says: it parses a dict, not a JSON string. Presumably the dict has previously been parsed from a JSON-like string?

A `from_json`-like interface should probably have the following signature:
```python
def from_json(cls, json_str: str) -> ProblemData
```
with intended usage
```python
with open("data.json") as fh:
    data = ProblemData.from_json(fh.read())
```

tests/test_ProblemData.py
```diff
@@ -1220,3 +1220,62 @@ def test_vehicle_max_trips_is_one_if_no_reload_depots(ok_small):
...
+        "duration_matrices": [[[0, 2], [2, 0]]],
+        "groups": [],
+    }
+    data = ProblemData.from_json(json_data)
+    assert isinstance(data, ProblemData)
+    assert data.num_clients == 1
+    assert data.num_depots == 1
+    assert data.num_vehicle_types == 1
+    assert data.num_profiles == 1
+    assert data.num_groups == 0
```
N-Wouda, line 1281: We use `numpy.testing` for our assertions, not 'naked' asserts. Could you update these to use `assert_equal` and/or `assert_`?

pyvrp/read.py
```diff
@@ -552,3 +552,40 @@ def _allowed2profile(self) -> dict[tuple[int, ...], int]:
...
+    """
+    Creates a :class:`~pyvrp._pyvrp.ProblemData` instance from a JSON-like
+    dictionary.
+
+    Parameters
+    ----------
+    data : dict
```
N-Wouda, line 564:
```suggestion
    data
```
(argument types are mostly unnecessary in the docstrings because we already type them in the function signature)

N-Wouda (commented):
Hi @braktar, thanks for having a go at adding support for JSON instance files. I see the utility of how parsing data from multiple different formats can be helpful, besides just the basic pickle support we have now. A `from_json` method in that sense could be very helpful. I am not sure, however, if we should include these directly in PyVRP, or on the `ProblemData` class.

A few things I think we need to hash out are the following:
- Do we want to have a schema for the JSON files? Or can we rely on the user/validation inside the `ProblemData` constructor?
- Where should we put these sorts of I/O interop methods? Should we have them just for the instance data, or also for solutions? Is this a concern for PyVRP, or should such things exist user-side?

braktar:
Ok, I understand your point. You have the big picture, so I will rely on your opinion.

Before closing the pull request, what I can do is revert the latest changes to bring back the monkey patch, so it will remain available for any other users who might need it.

PR #812, opened 2025-05-23
Add failing test case

N-Wouda:
This PR fixes a bug where iterating over a route with an empty last trip triggered an assert. As such, it:

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/tests/test_Route.py
+++ b/tests/test_Route.py
@@ -852,3 +852,14 @@ def test_multi_trip_release_time_routes(mtvrptw_release_times, trips):
 
     for trip1, trip2 in pairwise(trips):
         assert_(trip1.release_time() <= trip2.release_time())
+
+
+def test_bug_iterating_with_empty_last_trip(ok_small_multiple_trips):
+    """
+    TODO
+    """
+    trip1 = Trip(ok_small_multiple_trips, [1, 2], 0)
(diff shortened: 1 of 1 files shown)


leonlan (approved):

pyvrp/cpp/Route.cpp
```diff
@@ -69,7 +69,7 @@ Route::Iterator &Route::Iterator::operator++()
     // we continue to the next until we either exhaust all trips, or we find a
     // non-empty trip.
     ++trip_;
-    while (trip_ + 1 < trips.size() && trips[trip_].empty())
+    while (trip_ < trips.size() && trips[trip_].empty())
```
N-Wouda, line 72: So we don't forget:

`Route::Iterator` provides a flattened view of the clients in a route. A route is composed of one or more trips, so we need to iterate through both the list of trips, and each trip itself. We do this by tracking a `trip_` index, and an `idx_` into `trips[trip_]`.

We should skip empty trips, since indexing those is a big no-no. We do this by checking if the `trips[trip_]` is empty, and then incrementing `trip_` if it is. We should also increment in case we're at the last trip, since that gets us to the iterator returned by `route.end()` and signals end of iteration.

Previously we did not do this - we didn't increment past the last trip. This resulted in the indexing of the last trip, which blows up when that trip is empty. The change here fixes that bug.

PR #813, opened 2025-05-23
Fix bug evaluating moves involving initial vehicle load

N-Wouda:
This PR fixes a bug where moves involving vehicle types with initial vehicle load sometimes determined a no-op move to be improving, because they incorrectly decided excess load would 'disappear' as a consequence of the move. Finally, this PR:

- [x] Adds and passes relevant tests.
- [x] Has well-formatted code and documentation.

Thanks @codyfletcher for bringing this to my attention!

<details>

**Notes**:

Please read our [contributing guidelines](https://pyvrp.org/dev/contributing.html) first.
In particular:

- You must add tests when making code changes.
  This keeps the code coverage level up, and helps ensure the changes work as intended.
- When fixing a bug, you must add a test that would produce the bug in the master branch, and then show that it is fixed with the new code. 
- New code additions must be well formatted. Changes should pass the pre-commit workflow, which you can set up locally using [pre-commit](https://pre-commit.com/#intro). 
- Docstring additions must render correctly, including escapes and LaTeX.
- Finally, it is essential that all contributions in this PR are license-compatible with PyVRP's MIT license.
  Please check that this PR can be included into PyVRP under the MIT license.

</details>

PR #818, opened 2025-05-26
Evaluate inserting a reload depot just after start, or just before end

N-Wouda:
Based on feedback I got, adding a reload depot just after the start depot, or just before the end depot can be very beneficial for modelling certain routing problems (see also #812, #813 for related bug fixes). This PR modifies `TripRelocate` to explicitly evaluate those moves, too.

<details>

**Notes**:

In particular:

</details>


leonlan (approved):
LGTM!

PR #756, opened 2025-03-11
Iterated local search

leonlan:
Hacky prototype implementation for ILS.

This PR:

- [x] Closes #533.
- [ ] Adds and passes relevant tests.
- [ ] Has well-formatted code and documentation.

<details>

**Notes**:

Please read our [contributing guidelines](https://pyvrp.org/dev/contributing.html) first.
In particular:

- You must add tests when making code changes.
  This keeps the code coverage level up, and helps ensure the changes work as intended.
- When fixing a bug, you must add a test that would produce the bug in the master branch, and then show that it is fixed with the new code. 
- New code additions must be well formatted. Changes should pass the pre-commit workflow, which you can set up locally using [pre-commit](https://pre-commit.com/#intro). 
- Docstring additions must render correctly, including escapes and LaTeX.
- Finally, it is essential that all contributions in this PR are license-compatible with PyVRP's MIT license.
  Please check that this PR can be included into PyVRP under the MIT license.

</details>

PR #827, opened 2025-05-30
Perturbation methods in local search

leonlan:
This PR adds a `perturb()` method to LocalSearch. It replaces #826 (perturbation methods) because there was a lot of duplication between DestroyRepair and LocalSearch, and there was overhead because of double load/export solution. 

Instead, this PR directly adds the perturbation operators to local search with an `add_perturbation_operator` method. The perturbation operators can do anything with the nodes and routes (e.g., only destroy, destroy + repair, only vehicle type changes), it's up to the implementation.

This should be sufficient to have a basic implementation of ILS, where we possibly want to call `perturb()` inside `LocalSearch::operator()` before proceeding with search. We may add a few more perturbation operators, including a "swap routes" perturbation operator, which changes a route's vehicle type.

- [x] Is part of #778.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/search/LocalSearch.cpp
+++ b/pyvrp/cpp/search/LocalSearch.cpp
@@ -17,6 +17,7 @@ Solution LocalSearch::operator()(Solution const &solution,
                                  CostEvaluator const &costEvaluator)
 {
     loadSolution(solution);
+    perturb(costEvaluator);
 
     while (true)
     {
--- a/tests/search/test_LocalSearch.py
+++ b/tests/search/test_LocalSearch.py
@@ -49,6 +49,26 @@ def test_local_search_returns_same_solution_with_empty_neighbourhood(ok_small):
     assert_equal(ls.search(sol, cost_evaluator), sol)
 
 
+def test_local_search_call_perturbs_solution(ok_small):
+    """
+    Tests that calling local search also perturbs a solution.
+    """
+    rng = RandomNumberGenerator(seed=42)
+    neighbours = compute_neighbours(ok_small)
+    ls = LocalSearch(ok_small, rng, neighbours)
+    ls.add_perturbation_operator(NeighbourRemoval(ok_small, 2))
+
+    sol = Solution.make_random(ok_small, rng)
+    cost_eval = CostEvaluator([1], 1, 0)
+
+    # ``__call__()`` should perturb the solution even though no node and route
+    # operators are added. Because of the neighbourhood removal operator, the
+    # resulting solution should have less clients than the original one.
+    perturbed = ls(sol, cost_eval)
+    assert_(perturbed != sol)
+    assert_(perturbed.num_clients() < sol.num_clients())
+
+
 @pytest.mark.parametrize("size", [1, 2, 3, 4, 6, 7])  # num_clients + 1 == 5
 def test_raises_when_neighbourhood_dimensions_do_not_match(ok_small, size):
     """


leonlan:
Replaces #826 because there's a lot of duplication between DestroyRepair and LocalSearch, and there's overhead because of double load/export solution. Destroy operators can be added to local search with a `destroy()` step and the greedy insert is practically the same -- two minor differences are:
- `perturb::GreedyInsert` inserts in empty routes, and
- `perturb::GreedyInsert` inserts optional clients with some probability.

leonlan:
Strange RTD bug that I cannot reproduce locally: https://app.readthedocs.org/api/v2/build/28351102.txt

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -48,6 +49,14 @@ Solution LocalSearch::intensify(Solution const &solution,
     return exportSolution();
 }
 
+Solution LocalSearch::perturb(Solution const &solution,
+                              CostEvaluator const &costEvaluator)
+{
+    loadSolution(solution);
+    perturb(costEvaluator);
+    return exportSolution();
+}
```
N-Wouda, line 59: Is this mostly for testing?

pyvrp/cpp/search/LocalSearch.h
```diff
@@ -176,9 +192,16 @@ class LocalSearch
...
+     * Performs a perturbation step around the given solution, and returns a
+     * new, modified solution.
+     */
+    Solution perturb(Solution const &solution,
+                     CostEvaluator const &costEvaluator);
+
     /**
      * Shuffles the order in which the node and route pairs are evaluated, and
-     * the order in which node and route operators are applied.
+     * the order in which all operators are applied.
```
N-Wouda:
```suggestion
     * the order in which operators are applied.
```
all to me adds undue stress on "we're going to apply them all", when really we're just changing ordering.

pyvrp/cpp/search/NeighbourRemoval.h
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Neighbour removal perturbation operator. This operator removes the closest
+ * neighbours around a randomly selected client. The removed clients are *not*
+ * reinserted back into the solution - this is handled by the local search's
+ * ``search()`` method.
+ */
+class NeighbourRemoval : public PerturbationOperator
+{
+    ProblemData const &data;
+    size_t const numPerturb;
```
N-Wouda:
```suggestion
    ProblemData const &data_;
    size_t const numPerturb_;
```

tests/search/test_LocalSearch.py
```diff
@@ -647,3 +648,23 @@ def test_node_and_route_operators_property(ok_small):
...
+    rng = RandomNumberGenerator(seed=42)
+    ls = LocalSearch(ok_small, rng, compute_neighbours(ok_small))
+
+    # The local search has not yet been equipped with perturbation operators,
+    # so it should start empty.
+    assert_equal(len(ls.perturbation_operators), 0)
+
+    # Now we add a perturbation operator. The local search does not take
+    # ownership, so its only perturbation operator should be the exact same
+    # object as the one we just created.
```
N-Wouda, line 687: OK, but we're not checking that yet.

N-Wouda: Maybe `assert_(ls.perturbation_operators[0] is perturb_op)`?

tests/search/test_NeighbourRemoval.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    rng = RandomNumberGenerator(seed=42)
+    neighbours = compute_neighbours(ok_small)
+    cost_eval = CostEvaluator([1], 1, 0)
+
+    ls = LocalSearch(ok_small, rng, neighbours)
+    ls.add_perturbation_operator(NeighbourRemoval(ok_small, 4))
+
+    sol = Solution(ok_small, [])
+    assert_equal(ls.perturb(sol, cost_eval), sol)
```
N-Wouda:
```suggestion
    rng = RandomNumberGenerator(seed=42)
    neighbours = compute_neighbours(ok_small)
    ls = LocalSearch(ok_small, rng, neighbours)
    ls.add_perturbation_operator(NeighbourRemoval(ok_small, 4))

    sol = Solution(ok_small, [])
    cost_eval = CostEvaluator([1], 1, 0)
    assert_equal(ls.perturb(sol, cost_eval), sol)
```
Here, and for other tests in this file: define things closer to where they're used.

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -146,15 +155,23 @@ void LocalSearch::intensify(CostEvaluator const &costEvaluator)
     }
 }
 
+void LocalSearch::perturb(CostEvaluator const &costEvaluator)
+{
+    if (perturbOps.empty())
+        return;
+
+    (*perturbOps[0])(nodes, routes, costEvaluator, neighbours_, orderNodes);
+}
```
N-Wouda, line 165: So we apply only a single perturbation operator?

pyvrp/cpp/search/NeighbourRemoval.h
```diff
@@ -0,0 +1,44 @@
...
+     * num_perturb
+     *     Maximum number of clients to remove from the solution.
+     */
+    NeighbourRemoval(ProblemData const &data, size_t const numPerturb);
+
+    void operator()(std::vector<search::Route::Node> &nodes,
+                    std::vector<search::Route> &routes,
+                    CostEvaluator const &costEvaluator,
+                    std::vector<std::vector<size_t>> const &neighbours,
+                    std::vector<size_t> const &orderNodes) override;
```
N-Wouda, line 39: I don't really know about this interface but I think we can go ahead with it for now, and revisit that later. I'd rather first get something reasonable going on `feat/ils` than attempt to figure this out in detail without having a fully working implementation to refactor.

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -10,6 +10,7 @@
 using pyvrp::Solution;
 using pyvrp::search::LocalSearch;
 using pyvrp::search::NodeOperator;
+using pyvrp::search::PerturbationOperator;
 using pyvrp::search::RouteOperator;
 
 Solution LocalSearch::operator()(Solution const &solution,
```
N-Wouda, line 16: Should we not perturb here?

N-Wouda: I assumed something like:
```
load
perturb

while not done:
   improve

return export
```
as our main loop

pyvrp/cpp/search/NeighbourRemoval.h
N-Wouda: (I've started doing this with new member data, to clarify that it's not a local variable/argument, but rather part of the object state)

pyvrp/cpp/search/LocalSearch.cpp
leonlan: Yes, the intended use of perturbation is to only apply a single operator. For context, there will be multiple perturbation operators, if necessary, but I don't want to already start implementing those because they may turn out to be redundant.

Other perturbation operators are:
- String removals: Remove consecutive clients of the same route.
- Swap routes perturbation: Change the vehicle types of some routes (e.g., up to `numPerturb` clients are involved).

I could think of a few more, but this is what I did in #756, and will be the first ones that I want to try out when I get to improving ILS.

pyvrp/cpp/search/LocalSearch.cpp
leonlan: Yea, that's right. I refrained from adding it here because it impacts how the entire `solve` algorithm works, if we were to add perturbation operators to the default solver configuration, which doesn't make sense if we still use GA.

The solution is to just add it here now and not change `solve()` until the next PR.

pyvrp/cpp/search/LocalSearch.cpp
leonlan: Yes, to isolate the perturbation step. The solve such just use `LS::operator()`, similar how we use that instead of `search` and `intensify`.

PR #838, opened 2025-06-03
Replace GeneticAlgorithm with IteratedLocalSearch

leonlan:
This PR replaces `GeneticAlgorithm` with `IteratedLocalSearch` and modifies related components to work with ILS. Other GA components will be removed in another PR to minimize changes in this PR.

- [x] Part of #778.

<details>

**Notes**:

In particular:

</details>


tests/test_Model.py
```diff
@@ -533,11 +534,11 @@ def test_model_solves_line_instance_with_multiple_depots():
     depot1 = m.add_depot(x=0, y=0)  # location 0
     depot2 = m.add_depot(x=5, y=0)  # location 1
 
-    m.add_vehicle_type(1, start_depot=depot1, end_depot=depot1)
-    m.add_vehicle_type(1, start_depot=depot2, end_depot=depot2)
+    m.add_vehicle_type(1, capacity=2, start_depot=depot1, end_depot=depot1)
```
leonlan, line 537: There are multiple optimal solutions here, e.g., a solution with routes `[2, 3, 4]` and `[5]`. Setting capacity to two forces the solution to take on routes `[2, 3]` and `[4, 5]`.

tests/test_Model.py
```diff
@@ -582,34 +584,32 @@ def test_client_depot_and_vehicle_type_name_fields():
...
         # And no delivery or pickup amounts should of course also be OK!
-        ([0, 0, 0, 0], [0, 0, 0, 0], 0),
+        ([0, 0, 0, 0], [0, 0, 0, 0]),
     ],
 )
 def test_model_solves_instances_with_pickups_and_deliveries(
-    pickups: list[int],
-    deliveries: list[int],
-    expected_excess_load: int,
+    pickups: list[int], deliveries: list[int]
```
leonlan, line 604: This test relied on `solve()` to return its best infeasible solution. ILS does not track infeasible best solutions, so I changed the test to be feasible at all times.

pyvrp/plotting/plot_objectives.py
leonlan, line 1: Not plotting candidate because it's a lot of noise:
image

pyvrp/plotting/plot_objectives.py
```diff
@@ -38,17 +38,12 @@ def _plot(x, y, *args, **kwargs):
         ax.plot(x[num_to_skip:], y[num_to_skip:], *args, **kwargs)
 
     x = 1 + np.arange(result.num_iterations)
-    y = [d.best_cost for d in result.stats.infeas_stats]
-    _plot(x, y, label="Infeas. best", c="tab:red")
 
-    y = [d.avg_cost for d in result.stats.infeas_stats]
-    _plot(x, y, label="Infeas. avg.", c="tab:red", alpha=0.3, ls="--")
+    y = [d.current_cost for d in result.stats.data]
```
leonlan: Not plotting candidate because it adds a lot of visual noise:
image

pyvrp/Statistics.py
```diff
@@ -164,14 +119,15 @@ def from_csv(cls, where: Path | str, delimiter: str = ",", **kwargs):
...
             datum = {}
 
             for name, value in row.items():
-                if (field_name := name[len(prefix) :]) in field2type:
-                    # If the prefixless name is a field name, cast the row's
-                    # value to the appropriate type and add the data.
-                    datum[field_name] = field2type[field_name](value)
+                if name in field2type:
+                    if field2type[name] is bool:
+                        datum[name] = value.lower() == "true"
```
leonlan: This does not seem to be a robust way to handle booleans, but I wasn't quite sure what else to do. Store as 0/1s instead?

pyvrp/solve.py
```diff
@@ -173,16 +209,45 @@ def solve(
...
-    crossover = srex if data.num_vehicles > 1 else ox
+    # Infer MBAT convergence parameters from stopping criteria, if not set.
+    mbat = copy(params.mbat)
+    if mbat.max_runtime is None:
+        mbat.max_runtime = _stop2runtime(stop)
 
-    gen_args = (data, pm, rng, pop, ls, crossover, init, params.genetic)
-    algo = GeneticAlgorithm(*gen_args)  # type: ignore
+    if mbat.max_iterations is None:
+        mbat.max_iterations = _stop2iterations(stop)
```
leonlan: This is the thing about ILS that bugs me the most: MBAT is coupled with stopping criteria. The way I implemented here is clearly suboptimal, but we might as well just get a real prototype working and refactor it then. At some point, I'd like to change MBAT to not rely on runtime at all. This will require some experimentation that we can do after we have a basic version working.

pyvrp/IteratedLocalSearch.py
```diff
@@ -1 +1,256 @@
...
+        params: IteratedLocalSearchParams = IteratedLocalSearchParams(),
+    ):
+        self._data = data
+        self._pm = penalty_manager
+        self._rng = rng
+        self._search = search_method
+        self._init = initial_solution
+        self._params = params
+
+        self._history: deque[int] = deque(maxlen=self._params.history_length)
```
leonlan, line 96: Reverted to a double-ended queue for now because it makes things much cleaner. In a future PR, we should probably write our own custom `SolutionHistory`/`CostHistory` class in C++.

pyvrp/stop/StoppingCriterion.py
```diff
@@ -6,6 +6,19 @@ class StoppingCriterion(Protocol):  # pragma: no cover
     Protocol that stopping criteria must implement.
     """
 
+    def fraction_remaining(self) -> float | None:
```
leonlan, line 9: Maybe rename to `progress()`, where 0 means no progress and 1 means that the stopping criterion has been met?

docs/source/api/pyvrp.rst
```diff
@@ -6,9 +6,9 @@ PyVRP
 =====
 
 The top-level :mod:`pyvrp` module exposes several core classes needed to run the VRP solver.
-These include the core :class:`~pyvrp.GeneticAlgorithm.GeneticAlgorithm`, and the :class:`~pyvrp.Population.Population` that manages a :class:`~pyvrp._pyvrp.Solution` pool.
+These include the core :class:`~pyvrp.IteratedLocalSearch.IteratedLocalSearch`.
 Most classes take parameter objects that allow for advanced configuration - but sensible defaults are also provided.
```
N-Wouda: Maybe also say something about the Solution and/or Result classes? Users typically interact with those, too.

examples/using_pyvrp_components.ipynb
N-Wouda, line 1: Is this intended? (all cells are commented out)

pyvrp/IteratedLocalSearch.py
N-Wouda: Why - is this slow?

pyvrp/IteratedLocalSearch.py
```diff
@@ -1 +1,256 @@
...
+                candidate = self._search(
+                    candidate, self._booster_cost_evaluator
+                )
+                self._pm.register(candidate)
+
+            stats.collect(
+                *self._stats(current),
+                *self._stats(candidate),
+                *self._stats(best),
+            )
```
N-Wouda: Odd interface. I'll have a look at `Statistics` first to think of what to do here.

pyvrp/IteratedLocalSearch.py
```diff
@@ -1 +1,256 @@
...
+            print_progress.iteration(stats)
+
+            if not candidate.is_feasible():
+                continue  # don't accept infeasible candidates
```
N-Wouda: Those also do not enter the history?

N-Wouda: Or need to be counted for `iters_no_improvement`?

N-Wouda: I assume this'd be a task for the acceptance criterion, but that doesn't even take the candidate solution at all.

pyvrp/ProgressPrinter.py
```diff
@@ -58,9 +59,9 @@ def __init__(self, should_print: bool, display_interval: float):
 
     def iteration(self, stats: Statistics):
         """
-        Outputs relevant information every few hundred iterations. The output
-        contains information about the feasible and infeasible populations,
-        whether a new best solution has been found, and the search duration.
+        Outputs relevant information every five hundred iterations. The output
```
N-Wouda: @@It's now time-based. Maybe something like "every few seconds"?

pyvrp/ProgressPrinter.py
N-Wouda, line 1: How does the output look in a run? (incl. restart?)

pyvrp/Statistics.py
```diff
@@ -1,50 +1,26 @@
...
-
-_FEAS_CSV_PREFIX = "feas_"
-_INFEAS_CSV_PREFIX = "infeas_"
-
 
 @dataclass
 class _Datum:
     """
-    Single subpopulation data point.
+    Single ILS iteration data point.
```
N-Wouda:
```suggestion
    Single iteration data point.
```

pyvrp/Statistics.py
```diff
@@ -1,50 +1,26 @@
...
+    best_cost: int
+    best_feas: bool
 
 
 class Statistics:
     """
-    The Statistics object tracks various (population-level) statistics of
-    genetic algorithm runs. This can be helpful in analysing the algorithm's
-    performance.
+    Statistics about the iterated local search (ILS) progress.
```
N-Wouda:
```suggestion
    Statistics about the search progress.
```
(might just want to keep this general in case we ever change algorithms again)

pyvrp/Statistics.py
```diff
@@ -73,25 +43,38 @@ def __eq__(self, other: object) -> bool:
...
-        self, population: Population, cost_evaluator: CostEvaluator
+    def collect(
+        self,
+        current_cost: int,
+        current_feas: bool,
+        candidate_cost: int,
+        candidate_feas: bool,
+        best_cost: int,
+        best_feas: bool,
     ):
```
N-Wouda: Probably better to just pass the objects, and a cost_evaluator. That way we can collect additional statistics later if desired, without having to the interface.

pyvrp/Statistics.py
N-Wouda: Yeah I think so.

pyvrp/Statistics.py
```diff
@@ -211,30 +166,19 @@ def to_csv(
...
-
             writer.writeheader()
 
-            for idx in range(self.num_iterations):
+            for idx, datum in enumerate(data):
                 row = dict(runtime=self.runtimes[idx])
-                row.update(feas_data[idx])
-                row.update(infeas_data[idx])
-
+                row.update(datum)
```
N-Wouda: Enumerate runtime with zip, and then do `row = dict(runtime=runtime, **datum)`?

pyvrp/plotting/plot_objectives.py
N-Wouda: Make stats iterable?

N-Wouda: Maybe make candidate a bit opaque? I think there's value in seeing that, too, but I agree this adds too much noise.

pyvrp/plotting/plot_objectives.py
```diff
@@ -11,7 +11,7 @@ def plot_objectives(
     ylim_adjust: tuple[float, float] = (0.95, 1.15),
 ):
     """
-    Plots each subpopulation's objective values.
+    Plots each iteration's current and best objective values.
```
N-Wouda:
```suggestion
    Plots each iteration's objective values.
```
I think we can be (intentionally) vague about _which_ objectives we plot.

pyvrp/solve.py
```diff
@@ -183,16 +202,14 @@ def solve(
...
-    ]
+    for perturb_op in params.perturbation_ops:
+        ls.add_perturbation_operator(
+            perturb_op(data, params.num_perturbations)
+        )
 
-    # We use SREX when the instance is a proper VRP; else OX for TSP.
-    crossover = srex if data.num_vehicles > 1 else ox
+    pm = PenaltyManager.init_from(data, params.penalty)
+    init = Solution(data, [])  # type: ignore
```
N-Wouda, line 211: I think this is OK for now but the next step is probably that we make an argument somehow. (#855)

pyvrp/stop/MultipleCriteria.py
```diff
@@ -12,5 +12,13 @@ def __init__(self, criteria: list[StoppingCriterion]) -> None:
 
         self.criteria = criteria
 
+    def fraction_remaining(self) -> float | None:
+        fractions = [
+            fraction
+            for crit in self.criteria
+            if (fraction := crit.fraction_remaining()) is not None
+        ]
+        return min(fractions) if fractions else None
```
N-Wouda:
```suggestion
        fractions = [crit.fraction_remaining() for crit in self.criteria]
        return min([f for f in fractions if f is not None], default=None)
```
Looks simpler to me, but this is a bit subjective

pyvrp/solve.py
```diff
@@ -31,68 +32,71 @@
...
     neighbourhood
         Neighbourhood parameters.
     node_ops
         Node operators to use in the search.
     route_ops
         Route operators to use in the search.
+    perturbation_ops
+        Perturbation operators to use in the search.
+    num_perturbations
+        Number of perturbations to apply in each iteration. Default 10.
```
N-Wouda, line 52: What does this mean?

pyvrp/stop/StoppingCriterion.py
N-Wouda: I considered that and think that was my initial naming for this, but `fraction_remaining` seems clearer than some generic 'progress' value.

pyvrp/stop/StoppingCriterion.py
```diff
@@ -6,6 +6,19 @@ class StoppingCriterion(Protocol):  # pragma: no cover
...
+        """
+        Returns the fraction of the stopping criterion that is remaining.
```
N-Wouda:
```suggestion
        Returns the fraction of the stopping criterion's budget that remains.
```

N-Wouda: Maybe include something about a "budget" here

tests/stop/test_NoImprovement.py
```diff
@@ -69,3 +69,54 @@ def test_n_max_iterations_with_single_improvement(n, k):
...
+    assert_equal(stop.fraction_remaining(), 1)
+
+    # New solution is not improving, so fraction remaining should decrease.
+    stop(2)
+    assert_equal(stop.fraction_remaining(), 0.99)
+
+    # Found a new best solution, so the iteration counter resets and so does
+    # the fraction remaining.
+    stop(0)
+    assert_equal(stop.fraction_remaining(), 1)
```
N-Wouda, line 92: Not really a review comment, but I just realised that using this criterion together with MBAT could result in somewhat unexpected/fun dynamic behaviour 😆.

tests/test_Model.py
```diff
@@ -1053,7 +1051,7 @@ def test_instance_with_multi_trip_and_release_times(mtvrptw_release_times):
            https://doi.org/10.1287/trsc.2022.1161.
     """
     m = Model.from_data(mtvrptw_release_times)
-    res = m.solve(stop=MaxIterations(5))
+    res = m.solve(stop=MaxIterations(20))  # TODO decrease iters later
```
N-Wouda, line 1054: Is later now?

pyvrp/IteratedLocalSearch.py
leonlan: Yes, in earlier tests, computing the best and average costs from history took up to 10% with `deque`. We run a lot of iterations in ILS.

---

It doesn't really seem to be that much anymore in more recent profiles (1-2%). Can keep it like this for convenience.

tests/stop/test_NoImprovement.py
leonlan: This actually works surprisingly well when combined with MaxRuntime. The idea is to have an overall weight decay (as determined by `MaxRuntime.fraction_remaining()`) while the `NoImprovement.fraction_remaining()` is used to do more exploitation on the short-term when the search is stagnating. 

I played around combining MaxRuntime and NoImprovement, which resulted in a very good solution to [PR20B](https://github.com/PyVRP/Instances/pull/13/files#) -- about 1% better than just having MaxRuntime. I think we can try to investigate this further in a later iteration.

tests/test_Model.py
leonlan: Not yet, after #860 :-)

pyvrp/plotting/plot_objectives.py
leonlan: `alpha=0.2`
image


`alpha=0.3`
image


I think `alpha=0.2` looks best.

pyvrp/ProgressPrinter.py
leonlan:
```
PyVRP v0.12.0a0

Solving an instance with:
    1 depot
    100 clients
    100 vehicles (1 vehicle type)

                  |               Cost (Feasible)
    Iters    Time |      Current    Candidate         Best
H    1303      5s |    28575 (Y)    28465 (N)    27892 (Y)
     2625     10s |    28245 (Y)    28446 (Y)    27892 (Y)
R                 |                restart
H    3950     15s |    28242 (Y)    28567 (Y)    27742 (Y)
R                 |                restart
     5320     20s |    28410 (Y)    28553 (N)    27742 (Y)
     6703     25s |    28251 (Y)    28905 (Y)    27742 (Y)
R                 |                restart
     8080     30s |    27862 (Y)    29737 (Y)    27742 (Y)
R                 |                restart
     9463     35s |    28052 (Y)    28310 (Y)    27742 (Y)
```

docs/source/api/pyvrp.rst
leonlan: I rephrased this a bit more to highlight the most important classes in this module.

pyvrp/solve.py
leonlan: Yeah, uhh, not a clear definition yet. Within `NeighbourRemoval`, it is defined as the number of clients to be removed. But it could change when we have different perturbation operators 🤷. For instance, we also need to have a "change vehicle type" perturbation operator, where we change the vehicle type of one route. `num_perturbations` could then be interpreted as the number of clients in the route -- I don't know yet for sure. The perturbation interface is still a bit odd.

examples/using_pyvrp_components.ipynb
leonlan: Kinda, but I just updated the notebook. We should revise this again once the perturbation interface is better defined.

pyvrp/solve.py
leonlan: Not sure if this parameter should be attached on LocalSearch or on the operator. But this is something for later.

docs/source/api/pyvrp.rst
```diff
@@ -5,11 +5,11 @@
 PyVRP
 =====
 
-The top-level :mod:`pyvrp` module exposes several core classes needed to run the VRP solver.
-These include the core :class:`~pyvrp.GeneticAlgorithm.GeneticAlgorithm`, and the :class:`~pyvrp.Population.Population` that manages a :class:`~pyvrp._pyvrp.Solution` pool.
-Most classes take parameter objects that allow for advanced configuration - but sensible defaults are also provided.
-Finally, after running, the :class:`~pyvrp.GeneticAlgorithm.GeneticAlgorithm` returns a :class:`~pyvrp.Result.Result` object.
-This object can be used to obtain the best observed solution, and detailed runtime statistics.
+The top-level :mod:`pyvrp` module exposes several of PyVRP's core components.
+These include the :class:`~pyvrp._pyvrp.ProblemData` class for defining a VRP instance, the :class:`~pyvrp.IteratedLocalSearch.IteratedLocalSearch` class that implements the solver, and the :class:`~pyvrp._pyvrp.Solution` class for representing solutions.
```
N-Wouda:
```suggestion
These include the :class:`~pyvrp._pyvrp.ProblemData` class defining a VRP instance, the :class:`~pyvrp.IteratedLocalSearch.IteratedLocalSearch` class that implements the solver, and the :class:`~pyvrp._pyvrp.Solution` class representing VRP solutions.
```
is a bit nit but I don't like this whole "X for Y" thing - X does Y, so we can just say that instead.

pyvrp/IteratedLocalSearch.py
N-Wouda: Should be OK. I wouldn't change this right now. We can always improve it later in a follow-up PR.

pyvrp/IteratedLocalSearch.py
```diff
@@ -1 +1,260 @@
...
+            # Always accept candidates if we don't yet have a feasible best.
+            return True
+
+        if not candidate.is_feasible():
+            # Don't accept infeasible solutions and don't register their
+            # cost, as it would skew the history costs.
+            return False
+
+        cand_cost = self._cost_evaluator.cost(candidate)
+        self._history.append(cand_cost)
```
N-Wouda, line 178: This looks like a better place for pushing to the history, yes.

Should we also push the first feasible solution to the history (first if in this method)?

pyvrp/Statistics.py
```diff
@@ -73,25 +46,39 @@ def __eq__(self, other: object) -> bool:
...
-            and self.infeas_stats == other.infeas_stats
+            and self.data == other.data
         )
 
+    def __iter__(self) -> Generator[_Datum, None, None]:
+        """
+        Iterates over the collected data points.
+        """
+        for datum in self.data:
+            yield datum
```
N-Wouda:
```suggestion
        yield from self.data
```

pyvrp/solve.py
N-Wouda: Yeah OK let's figure this out in a later PR.

PR #847, opened 2025-06-05
Make CodSpeed solve benchmark run for more iterations

N-Wouda:
This PR makes our call to `solve()` run for 25 iterations, instead of 1, in the benchmarks. I noticed very little work was being done in the local search, while I think that's one of the main reasons to have this benchmark in the first place.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -169,7 +169,8 @@ dmypy.json
 debug/
 release/
 
-# Profiler results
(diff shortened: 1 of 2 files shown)


N-Wouda:
I think I'm missing something about how CodSpeed works, but for some reason it's just now showing the iterations at all. Very strange.

PR #850, opened 2025-06-06
Rename all nb to num

### LINKED ISSUE #848: Use num instead of nb everywhere
We typically abbreviate "number" as "num". E.g. `num_available`, `num_clients`, etc. A few holdovers from the early HGS days still use "nb" instead of "num". E.g., `nb_close`, `nb_elite`, etc.

I think we should just rename those.

N-Wouda:
This PR:

- [x] Closes #848.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/GeneticAlgorithm.py
+++ b/pyvrp/GeneticAlgorithm.py
@@ -32,7 +32,7 @@ class GeneticAlgorithmParams:
         Probability (in :math:`[0, 1]`) of repairing an infeasible solution.
         If the reparation makes the solution feasible, it is also added to
         the population in the same iteration.
-    num_iter_no_improvement
+    num_iters_no_improvement
         Number of iterations without any improvement needed before a restart
         occurs.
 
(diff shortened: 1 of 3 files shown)


pyvrp/search/neighbourhood.py
```diff
@@ -24,7 +24,7 @@ class NeighbourhoodParams:
         Penalty weight given to the minimum time warp aspect of the proximity
         calculation. A large time warp indicates the clients are far apart in
         duration/time.
-    nb_granular
+    num_neighbours
```
N-Wouda, line 27: Could've updated this to `num_granular`, but I think `num_neighbours` is much more descriptive.

pyvrp/GeneticAlgorithm.py
```diff
@@ -32,33 +32,33 @@ class GeneticAlgorithmParams:
         Probability (in :math:`[0, 1]`) of repairing an infeasible solution.
         If the reparation makes the solution feasible, it is also added to
         the population in the same iteration.
-    nb_iter_no_improvement
+    num_iter_no_improvement
```
leonlan: I have a preference for `num_iters_no_improvement`, similar to how we call it `iters_no_improvement` in GA.

PR #852, opened 2025-06-07
Add test for possible bug

N-Wouda:
This PR traces a possible bug that I do not yet fully understand. Based on a data instance I received that triggers a delta cost before/after violation, so there definitely is something here that isn't quite right.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/search/Route.h
+++ b/pyvrp/cpp/search/Route.h
@@ -1022,15 +1022,15 @@ std::pair<Duration, Duration> Route::Proposal<Segments...>::duration() const
 
             if (other.endsAtReloadDepot())
             {
-                // We can only finalise the current segment at the depot, so we
-                // need to travel there. The depot time windows restrictions are
-                // already handled by the other segment.
-                ds = DurationSegment::merge(edgeDur, {}, ds);
+                // The other segment ends at a reload depot, so we go there and
+                // finalise the current segment. We first travel there. We need
+                // to end the segment within the depot's time windows to
+                // properly account for any release time on our segment.
+                ProblemData::Depot const &depot = data.location(other.last());
+                ds = DurationSegment::merge(edgeDur, {depot}, ds);
                 ds = ds.finaliseFront();
 
-                // We finalise by travelling to the depot, so the remaining
-                // travel duration is now zero.
-                edgeDur = 0;
+                edgeDur = 0;  // we are already there!
             }
 
             ds = DurationSegment::merge(edgeDur, other.duration(profile), ds);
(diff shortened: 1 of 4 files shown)


N-Wouda:
I have already isolated a failing test case from the larger data instance I received. Now I need to understand what's going on in the move evaluation :)

pyvrp/cpp/search/Route.h
```diff
@@ -1027,29 +1055,25 @@ std::pair<Duration, Duration> Route::Proposal<Segments...>::duration() const
...
 
     auto const fn = [&](auto &&segment, auto &&...args)
     {
         auto ls = segment.load(dimension);
-        if (segment.last() < data.numDepots() && segment.size() != 1)
-            // Ends at a depot that is *not* the start depot. We should not
-            // finalise at the start depot, since that would immediately remove
-            // any initial load we might have.
+        if (segment.endsAtReloadDepot())
             ls = ls.finalise(capacity);
```
N-Wouda, line 1065: This is now much simpler to follow, too. Nice side effect.

pyvrp/cpp/search/Route.h
```diff
@@ -985,14 +1013,14 @@ std::pair<Duration, Duration> Route::Proposal<Segments...>::duration() const
...
 
         auto const merge = [&](auto const &self, auto &&other, auto &&...args)
         {
             auto edgeDur = matrix(other.last(), first);
 
-            if (other.last() < data.numDepots())  // other ends at a depot
+            if (other.endsAtReloadDepot())
             {
                 // We can only finalise the current segment at the depot, so we
                 // need to travel there. The depot time windows restrictions are
```
N-Wouda: This returns us to the depot to finalise, but we do not need to do that if other is just the start depot.

pyvrp/cpp/search/Route.h
```diff
@@ -985,32 +1013,32 @@ std::pair<Duration, Duration> Route::Proposal<Segments...>::duration() const
...
-                // We can only finalise the current segment at the depot, so we
-                // need to travel there. The depot time windows restrictions are
-                // already handled by the other segment.
-                ds = DurationSegment::merge(edgeDur, {}, ds);
+                // The other segment ends at a reload depot, so we go there and
+                // finalise the current segment. We first travel there. We need
+                // to end the segment within the depot's time windows to
+                // properly account for any release time on our segment.
+                ProblemData::Depot const &depot = data.location(other.last());
+                ds = DurationSegment::merge(edgeDur, {depot}, ds);
```
N-Wouda, line 1030: We need to finalise *after* we are at the depot, and in the depot's time windows. The way finalising works is that we put the artificial release time segment *before* the actual client/depot segment, so the depot needs to be merged in, too, before we finalise.

PR #854, opened 2025-06-10
Do not account for empty route costs in cost computations

### LINKED ISSUE #853: Performance degradation v0.11 involving vehicle types with different start and end depots
Hello!

Thanks for this awesome package.

It took several hours to understand that it wasn't my data instances that had changed enough, but that I had updated pyvrp to 0.11.2 while playing with my (other) dependencies.

I have a couple of instances I'm testing:

* small: 30 depots, 117 clients, 31 vehicles = 31 vehicle types
* large: 136 depots, 765 clients, 208 vehicles = 208 vehicle types

In version 0.10.1, pyvrp found initial feasible solutions in seconds (1s for the small, 40s for the large). and in less then 500 iterations for both.

In version 0.11.2, after 20 min it stops without having found any feasible solution in either instance. And it shows warnings of reaching scaling limits on objective function penalties.

For now, I pinned the 0.10.1 version. But if you care about these performance regressions and need some test cases to play, I can try to export the datasets. What is the easiest way to export them?

Franco

leonlan:
This PR fixes a bug identified in #853 by changing `Proposal` and `costEvaluator.deltaCost` to evaluate empty routes as having no cost.

Empty routes can have costs, specifically when their associated vehicle type has different start and end depots with non-zero distance/duration between. It is then also possible to have excess distance and time warp if the vehicle type has maximum route distance or duration constraints.

Having empty routes with costs caused a strange situation when computing delta costs with the instances provided in #853. If an empty route already had some time warp, then inserting a client could reduce its time warp, while the new route still had some time warp. The search would continuously insert clients in empty routes because the delta cost showed to be improving. However, this introduced new routes in the solution with time warp, whereas the empty routes were actually not part of the solution.

The proposed fix is to special-case `Proposal` and `costEvaluator.deltaCost` to assume zero cost whenever a route (proposal) is empty. This also corresponds with how we handle empty routes: we don't allow empty routes to be part of a solution, i.e., they don't contribute to a solution's cost.

I also double-checked that the instances from #853 can be solved feasibly after this change.

- [x] Closes #853.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/tests/search/test_Route.py
+++ b/tests/search/test_Route.py
@@ -1151,36 +1151,3 @@ def test_has_duration_cost(
 
     route = Route(data, 0, 0)
     assert_equal(route.has_duration_cost(), expected)
-
-
-def test_empty_route_statistics():
-    """
-    TODO #853
-    """
-    mat = [
-        [0, 10, 10],
-        [10, 0, 10],
-        [10, 10, 0],
-    ]
-    data = ProblemData(
-        depots=[Depot(x=0, y=0), Depot(x=0, y=0)],
-        clients=[Client(x=0, y=0)],
-        vehicle_types=[
-            VehicleType(
-                num_available=1,
-                start_depot=0,
-                end_depot=1,
-                max_duration=0,
-            )
-        ],
-        duration_matrices=[mat],
-        distance_matrices=[mat],
-    )
-
-    route = make_search_route(data, [])
-    assert_equal(route.time_warp(), 0)
-    assert_equal(route.duration(), 0)
-
-    route = make_search_route(data, [2])
-    assert_equal(route.time_warp(), 20)
-    assert_equal(route.duration(), 20)


pyvrp/cpp/CostEvaluator.h
N-Wouda, line 1: Should we also update `CostEvaluator::penalisedCost`?

pyvrp/cpp/search/Route.h
```diff
@@ -953,6 +966,12 @@ Route::Proposal<Segments...>::Proposal(Segments &&...segments)
     assert(last.last() == route->endDepot());      // must end at route end
 }
 
+template <Segment... Segments> bool Route::Proposal<Segments...>::empty() const
+{
+    auto const size = std::apply(
+        [](auto &&...args) { return (args.size() + ...); }, segments_);
```
N-Wouda: Maybe split this off into a `Proposal::size()` member? And then check here that `size() == 2`?

pyvrp/cpp/search/Route.h
```diff
@@ -953,6 +966,12 @@ Route::Proposal<Segments...>::Proposal(Segments &&...segments)
...
+    return size == 2;  // only contains start and end depot
+}
 template <Segment... Segments>
```
N-Wouda:
```suggestion
}

template <Segment... Segments>
```

pyvrp/cpp/CostEvaluator.h
```diff
@@ -256,14 +256,17 @@ bool CostEvaluator::deltaCost(Cost &out, T<Args...> const &proposal) const
 {
     auto const *route = proposal.route();
```
N-Wouda:
```suggestion
```
Maybe, nit, in line with the two-route deltaCost?

pyvrp/cpp/search/Route.cpp
```diff
@@ -335,11 +335,12 @@ void Route::update()
...
 
         load_[dim] = 0;
-        excessLoad_[dim]
-            = loadBefore[dim][nodes.size() - 1].excessLoad(capacity);
         for (auto it = depots_.begin() + 1; it != depots_.end(); ++it)
             load_[dim] += loadBefore[dim][it->idx()].load();
 
+        excessLoad_[dim]
+            = loadBefore[dim][nodes.size() - 1].excessLoad(capacity);
+
```
N-Wouda: Is this change required?

tests/search/test_Exchange.py
```diff
@@ -625,3 +625,39 @@ def test_bug_release_time_shift_time_windows():
     op = Exchange10(data)
     cost_eval = CostEvaluator([], 1, 0)
     assert_equal(op.evaluate(route1[1], route2[0], cost_eval), 0)
+
+
+def test_empty_route_delta_cost_bug():
+    """
+    Fixes a bug identified in #853 where the empty route's costs are
+    incorrectly included in delta cost evaluations.
```
N-Wouda:
```suggestion
    Fixes a bug identified in #853 where the empty route's costs are
    incorrectly included in delta cost evaluations.
```
```suggestion
    Tests that a bug identified in #853 where the empty route's costs are
    incorrectly included in delta cost evaluations has been fixed.
```

tests/search/test_primitives.py
```diff
@@ -80,7 +80,10 @@ def test_insert_cost_between_different_depots(ok_small_multi_depot):
     cost_eval = CostEvaluator([0], 0, 0)
     dist_mat = data.distance_matrix(0)
 
-    delta = dist_mat[0, 2] + dist_mat[2, 1] - dist_mat[0, 1]
+    # The delta cost is the distance from the start depot to the new node, and
+    # back to the end depot. The original route was empty so it should not
+    # contribute to the delta cost.
```
N-Wouda:
```suggestion
    # back to the end depot. The route is currently empty so it does not
    # contribute to the delta cost.
```

tests/search/test_primitives.py
```diff
@@ -262,3 +265,37 @@ def test_remove_consecutive_reload_depots(ok_small_multiple_trips):
 
     # And similarly removing the second depot should also be a no-op.
     assert_equal(remove_cost(route[4], data, CostEvaluator([1000], 0, 0)), 0)
+
+
+def test_empty_route_delta_cost_bug():
+    """
+    Fixes a bug identified in #853 where inserting or removing a client
+    involving an empty route incorrectly included the empty route's costs.
```
N-Wouda: Same with "Tests that X has been fixed" - the test fixes nothing, but it does check that something has been done.

N-Wouda (commented):
LGTM, but I think we should also update `CostEvaluator::penalisedCost`.

pyvrp/cpp/CostEvaluator.h
```diff
@@ -225,6 +225,13 @@ Cost CostEvaluator::distPenalty(Distance distance, Distance maxDistance) const
...
+    if (arg.empty())
+    {
+        if constexpr (PrizeCostEvaluatable<T>)
+            return arg.uncollectedPrizes();
+        return 0;
+    }
+
     // Standard objective plus infeasibility-related penalty terms.
     auto const cost = arg.distanceCost() + arg.durationCost()
                       + (!arg.empty() ? arg.fixedVehicleCost() : 0)
```
N-Wouda:
```suggestion
                      + arg.fixedVehicleCost()
```
right?

leonlan: Right!

N-Wouda (approved):

PR #860, opened 2025-06-14
Evaluate promising U's for node operators

leonlan:
This PR restricts local search to only evaluate promising moves. Specifically, rather than always evaluating all (U, V) pairs for all clients U, we keep track of U's that are promising in terms of resulting in improving moves.

- [x] Part of #778.

Notes
- We lose local optimality of local search. This affects a large number of tests. We should consider removing all "candidate" statuses only if we perturb. 

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/search/Route.h
+++ b/pyvrp/cpp/search/Route.h
@@ -603,7 +603,7 @@ inline Route::Node *p(Route::Node *node)
 
 inline Route::Node const *p(Route::Node const *node)
 {
-    auto &route = *node->route();
+    auto const &route = *node->route();
     return route[node->idx() - 1];
 }
 


pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -383,6 +446,15 @@ void LocalSearch::insert(Route::Node *U,
...
         update(UAfter->route(), UAfter->route());
+
+        if (!U->isDepot())
+            promising[U->client()] = true;
+
+        if (!p(U)->isDepot())
+            promising[p(U)->client()] = true;
+
+        if (!n(U)->isDepot())
+            promising[n(U)->client()] = true;
```
leonlan, line 457: TODO: this segfaults in MTVRPTW instances

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -191,6 +200,9 @@ bool LocalSearch::applyNodeOps(Route::Node *U,
                 = costEvaluator.penalisedCost(*rU)
                   + Cost(rU != rV) * costEvaluator.penalisedCost(*rV);
 
+            markPromising(U);
+            markPromising(V);
```
N-Wouda, line 204: No need to put this everywhere, I think. We can overload `LS::update` to take two nodes `U` and `V`, mark promising on both, and then call into `update(U->route(), V->route())`. `update()` is already called everywhere, much easier to roll it into that.

And then once `LS::intensify` is gone, we can remove the route-based `update`.

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -383,9 +397,24 @@ void LocalSearch::insert(Route::Node *U,
...
+    assert(U->route());
+
+    if (!U->isDepot())
+        promising[U->client()] = true;
+
+    if (!U->isStartDepot() && !p(U)->isDepot())
+        promising[p(U)->client()] = true;
+
+    if (!U->isEndDepot() && !n(U)->isDepot())
+        promising[n(U)->client()] = true;
```
N-Wouda: Does it matter if `U`, `p(U)`, or `n(U)` are depots? We never evaluate those, right?

pyvrp/cpp/search/NeighbourRemoval.cpp
```diff
@@ -19,6 +20,14 @@ void pyvrp::search::NeighbourRemoval::operator()(
         if (!U->route())
             continue;
 
+        promising[U->client()] = true;
+
+        if (!p(U)->isDepot())
+            promising[p(U)->client()] = true;
+
+        if (!n(U)->isDepot())
+            promising[n(U)->client()] = true;
```
N-Wouda: This duplicates `LS::markPromising`, but I don't know how to merge the two easily.

pyvrp/cpp/search/LocalSearch.cpp
leonlan: Correct, it's an artefact from an earlier version where I checked for `promising.any()` to terminate the search loop.

pyvrp/cpp/search/LocalSearch.cpp
leonlan: I tried this, but I encountered two issues:
- If a node is removed, we can no longer mark its old predecessor and successor nodes since the node is not on a route. So marking promising nodes through an update call _after_ removal is not possible (unless we track the pointers to the neighboring nodes). There's also a special case in `applyDepotRemovalMoves` where removing the reload depot and then marking through `update(U)` is invalid since that node is erased.
- There is a special case where V can become invalid: during `applyNodeOps` with V being the start depot, adding a new reload depot with TripRelocate can invalidate V. That's why I put `markPromising` before `nodeOp.apply`.

N-Wouda: Oh, right

PR #862, opened 2025-06-18
Use a `display_interval` parameter to print iterations

### LINKED ISSUE #861: Make ProgressPrinter's time/iterations between prints configurable
Currently we print output every 500 iterations:

https://github.com/PyVRP/PyVRP/blob/0613aedf1797e7e8994bdd6f70ea8b1825cf3ca2/pyvrp/ProgressPrinter.py#L57-L61

This works plenty well enough for small-ish instances, but not for large ones. I regularly look at time-between-prints of 30s to minutes. That's a bit annoying - I would like more feedback than that. It seems other users agree: there are two recent forks, [here](https://github.com/lassirati/PyVRP) and [here](https://github.com/la8804/PyVRP), that also update the iterations between prints.

I think we should make this a parameter argument somehow, so that it can be configured by the user.

N-Wouda:
This PR:

- [x] Closes #861.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/ProgressPrinter.py
+++ b/pyvrp/ProgressPrinter.py
@@ -48,6 +48,9 @@ class ProgressPrinter:
     """
 
     def __init__(self, should_print: bool, display_interval: float):
+        if display_interval < 0:
+            raise ValueError("Expected display_interval >= 0.")
+
         self._print = should_print
         self._display_interval = display_interval
         self._last_print_time = perf_counter()
--- a/tests/data/test_config.toml
+++ b/tests/data/test_config.toml
@@ -1,3 +1,5 @@
+display_interval = 10.0
+
 node_ops = [
     "Exchange10",
     "SwapTails",
--- a/tests/test_ProgressPrinter.py
+++ b/tests/test_ProgressPrinter.py
@@ -1,5 +1,5 @@
 import pytest
-from numpy.testing import assert_, assert_equal
+from numpy.testing import assert_, assert_equal, assert_raises
 
 from pyvrp import (
     CostEvaluator,
@@ -11,6 +11,15 @@ from pyvrp import (
 )
 from pyvrp.ProgressPrinter import ProgressPrinter
 from pyvrp.diversity import broken_pairs_distance as bpd
+
+
+def test_raises_negative_display_interval():
+    """
+    Tests that ProgressPrinter raises when provided with a negative
+    display_interval argument.
+    """
+    with assert_raises(ValueError):
+        ProgressPrinter(True, -1.0)
 
 
 def test_start(ok_small, capsys):
--- a/tests/test_solve.py
+++ b/tests/test_solve.py
@@ -1,4 +1,4 @@
-from numpy.testing import assert_, assert_equal
+from numpy.testing import assert_, assert_allclose, assert_equal
 
 from pyvrp.GeneticAlgorithm import GeneticAlgorithmParams
 from pyvrp.PenaltyManager import PenaltyParams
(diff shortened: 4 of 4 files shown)


N-Wouda:
Here's an example (printing every 5s by default, just like Gurobi):
```
PyVRP v0.12.0a0

Solving an instance with:
    1 depot
    100 clients
    25 vehicles (1 vehicle type)

H     367      5s |  39     7966     7773 |  37     8844     7611
H     872     10s |  34     7911     7761 |  35     8934     7711
     1358     15s |  41     7901     7761 |  54     8501     7661
     1856     20s |  56     7866     7761 |  39     7734     7615
     2300     25s |  29     7878     7761 |  60     7707     7511
     2707     30s |  39     7902     7761 |  39     7823     7726
     3085     35s |  37     7849     7761 |  35     7713     7626
     3540     40s |  35     7854     7761 |  62     7730     7676
     3943     45s |  47     7900     7761 |  40     7777     7702
     4402     50s |  35     7845     7761 |  65     7689     7503
     4848     55s |  44     7891     7761 |  47     7768     7714

Search terminated in 60.02s after 5288 iterations.
Best-found solution has cost 7761.

Solution results
================
    # routes: 4
     # trips: 4
   # clients: 100
   objective: 7761
    distance: 7761
    duration: 17761
# iterations: 5288
    run-time: 60.02 seconds
```

pyvrp/solve.py
```diff
@@ -132,6 +132,7 @@ def solve(
     seed: int = 0,
     collect_stats: bool = True,
     display: bool = False,
+    display_interval: float = 5.0,
     params: SolveParams = SolveParams(),
```
N-Wouda: Or we put this on `SolveParams`? That'd not be a breaking change.

pyvrp/ProgressPrinter.py
```diff
@@ -42,10 +43,14 @@ class ProgressPrinter:
     should_print
         Whether to print information to the console. When ``False``, nothing is
         printed.
+    display_interval
+        Time (in seconds) between iteration logs.
     """
 
-    def __init__(self, should_print: bool):
+    def __init__(self, should_print: bool, display_interval):
```
leonlan:
```suggestion
    def __init__(self, should_print: bool, display_interval: float):
```

pyvrp/ProgressPrinter.py
```diff
@@ -79,6 +85,7 @@ def iteration(self, stats: Statistics):
         )
         print(msg)
 
+        self._last_print_time += interval
```
leonlan: It took me a few minutes to understand what's going on here, because it's pretty subtle that `interval` increases to at least `display_interval` due to the shortcut. I think it's clearer to define a variable `current_time = perf_counter()` and to assign `self._last_print_time = current_time` here.

pyvrp/solve.py
leonlan: Will we introduce more parameters like this one in the future? I just went through [Gurobi's parameters](https://docs.gurobi.com/projects/optimizer/en/current/concepts/parameters/groups.html#secparametergroups) to get a feeling of what we might want to include in the future but I don't think we will add many more non-algorithm-related parameters.

If we put `display_interval` under `SolveParams`, do we put `display_interval` in a parameter group or just have it as a stand-alone? ~I think a group would be better to avoid breaking changes again in the future. Something like `ProgressPrinterParams`?~ (Edit: I don't think breaking this is an issue.)

leonlan: Something like a `stats_interval` param will also be useful when we introduce ILS: the number of iterations in ILS is an order of magnitude than in HGS. That could be another reason to have new parameters to be put on `SolveParams`.

PR #865, opened 2025-06-30
Improve ILS performance

leonlan:
Follow-up after #860. This PR improves the performance of ILS through some simple parameter tweaks. In particular, 
- PenaltyManager has smoother updates. We run many more iterations, so this allows us to update more often and thus react quicker to changes. 
- The number of neighbours is also increased since we now only consider promising U's.
- The route operators are commented out as we no longer rely on them.

Checklist:
- [x] Is part of #533.

<details>

**Notes**:

In particular:

</details>


leonlan:
Expected codspeed changes because we increased the number of neighbours from 40 to 60.

PR #869, opened 2025-07-11
Check before every load penalty evaluation if we can shortcut

### LINKED ISSUE #868: Investigate performance effects of multiple load dimensions
For me, so I don't forget. Some synthetic data with various load dimensions, just to see what the effects of that are and whether we can make things a bit faster.

N-Wouda:
This PR:

- [x] Closes #868.

<details>

**Notes**:

In particular:

</details>


leonlan (approved):

wouterkool:
@N-Wouda I was just checking this - shouldn't we also shortcut *after* evaluating the load penalty, before evaluating the duration? Currently (especially with 1 load dimension) we seem not to do this?

PR #877, opened 2025-07-21
Improve perturbation operators

leonlan:
This PR
- Introduces the `OptionalInsert` perturbation operator, which perturbs a solution by (forcefully) inserting optional clients. This is important for prize-collecting VRPs.
- Changes ``NeighbourRemoval`` to also remove, for each neighbour U, it's predecessor and successor p(U) and n(U). This is important for the late stages of search.
- Changes the number of perturbations to be randomly sampled rather than being fixed for all operators. This is important for TW instances, which sometimes need higher perturbations to escape local optima.

Furthermore:
- [x] Part of #778.

## Benchmarks

Table below compares v0.11.2 (HGS) with the current version for various time limits. After this, ILS has performance parity with HGS on academic benchmarks (that is, the time limits we currently use for our benchmarks). On shorter time limits (<10 min), ILS was already better and continues to be.

| Commit   | Time     | MDVRPTW | HFVRP  | GVRP   | VRPB   | MTVRPTWR | CVRP   | PCVRPTW | VRPTW  |
|----------|----------|---------|--------|--------|--------|----------|--------|---------|--------|
| v0.11.2  | 1 min    | 5.65%   | 6.79%  | 1.28%  | 3.39%  | 3.38%    | 1.53%  | 2.93%   | 6.00%  |
| **929a1347** (this PR) | 1 min | 2.14% | 1.68% | 0.44% | 1.43% | 1.38% | 0.85% | 2.35% | 2.39% |
| v0.11.2         | 10 min   | 1.72%   | 1.49%  | 0.18%  | 1.38%  | 2.30%    | 0.71%  | 0.58%   | 1.54%  |
| **929a1347** (this PR) | 10 min | 0.84% | 0.74% | 0.14% | 0.57% | 0.49% | 0.37% | 0.77% | 1.02% |
| v0.11.2   | academic | 0.75% | 0.76% | 0.10% | 0.58% | 2.53% | 0.25% | 0.16% | 0.58% | 
| **929a1347** (this PR) | academic | 0.47% | 0.57% | 0.10% | 0.39% | 0.49% | 0.30% | 0.35% | 0.60% |

<details>
<summary> Older benchmarks </summary>

| Commit   | Time     | MDVRPTW | HFVRP  | GVRP   | VRPB   | MTVRPTWR | CVRP   | PCVRPTW | VRPTW  |
|----------|----------|---------|--------|--------|--------|----------|--------|---------|--------|
| e62f635 (feat/iterated-local-search)  | 1 min    | 1.94%   | 2.06%  | 0.49%  | 1.11%  | 1.08%    | 0.74%  | 23.66%  | 2.53%  |
|          | 10 min   | 0.83%   | 1.13%  | 0.20%  | 0.66%  | 0.36%    | 0.35%  | 21.94%  | 1.75%  |
| 466f999 (optional insert) | 1 min    | 2.21%   | 1.79%  | 0.67%  | 1.39%  | 0.76%    | 0.70%  | 1.89%   | 3.12%  |
|          | 10 min   | 1.00%   | 0.82%  | 0.40%  | 1.00%  | 0.27%    | 0.40%  | 1.05%   | 2.20%  |
| 7d82334 (strings)  | 1 min    | 2.23%   | 1.57%  | 0.45%  | 1.26%  | -        | 0.79%  | 1.63%   | 2.10%  |
|          | 2 min    | 1.68%   | 1.21%  | 0.35%  | 0.99%  | -        | 0.64%  | 1.28%   | 1.66%  |
|          | 10 min   | 0.92%   | 0.66%  | 0.16%  | 0.54%  | -        | 0.37%  | 0.77%   | 1.06%  |
| 21636fa1 (dynamic neighbours) | 1 min    | 2.33%   | inf%   | 0.46%  | 1.47%  | 1.48%    | 0.89%  | 2.35%   | 2.39%  |
|          | 2 min    | 1.64%   | 1.26%  | 0.32%  | 1.08%  | 1.08%    | 0.67%  | 1.45%   | 1.71%  |
|          | 5 min    | 1.11%   | 0.89%  | 0.19%  | 0.74%  | 0.74%    | 0.51%  | 0.91%   | 1.29%  |
|          | 10 min   | 0.84%   | 0.70%  | 0.11%  | 0.58%  | 0.51%    | 0.39%  | 0.76%   | 1.03%  |
|      | academic | 0.51% | 0.51% | 0.09% | 0.49% | 0.49% | 0.28% | 0.37% | 0.60% |
| 3aa6a540 | 1 min | 1.99% | 1.68% | 0.47% | 1.37% | 1.34% | 0.78% | 1.98% | inf% |
| | 2 min | 1.51% | 1.24% | 0.32% | 1.05% | 0.99% | 0.61% | 1.26% | 1.62% |
| | 5 min | 1.06% | 0.87% | 0.17% | 0.73% | 0.65% | 0.44% | 0.93% | 1.21% |
| | 10 min | 0.82% | 0.65% | 0.12% | 0.57% | 0.49% | 0.35% | 0.73% | 1.02% |
| 6b1bd27d | 1 min | 2.71% | inf% | 0.48% | 1.74% | 2.44% | inf% | inf% | inf% |
| | 2 min | 2.04% | 1.82% | 0.36% | 1.36% | 1.95% | 1.07% | 1.35% | inf% |
| | 5 min | 1.42% | 1.37% | 0.23% | 0.95% | 1.44% | 0.86% | 0.99% | 1.54% |
| | 10 min | 1.11% | 1.17% | 0.14% | 0.74% | 1.14% | 0.72% | 0.79% | 1.30% |
| **929a1347** (new) | 1 min | 2.14% | inf% | 0.44% | 1.43% | 1.38% | 0.85% | 2.35% | inf% |
| | 2 min | 1.58% | 1.30% | 0.32% | 1.10% | 1.06% | 0.66% | 1.30% | 1.67% |
| | 5 min | 1.09% | 0.89% | 0.16% | 0.72% | 0.68% | 0.48% | 0.95% | 1.28% |
| | 10 min | 0.84% | 0.74% | 0.14% | 0.57% | 0.49% | 0.37% | 0.77% | 1.02% |
| | academic | 0.47% | 0.57% | 0.10% | 0.39% | 0.49% | 0.30% | 0.35% | 0.60% |

</details>

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/search/OptionalInsert.cpp
leonlan, line 1: This operator is really necessary for prize-collecting VRPs, which have local optima that are much harder to escape from. To overcome this, this operator "force inserts" optional clients into the solutions. I don't know why exactly, but it's particularly important to consider empty routes.

pyvrp/IteratedLocalSearch.py
leonlan, line 1: I slipped this change in here as well because it helps with performance: 
1. Infeasible candidates are also accepted. This helps quite a bit in TW instances.
2. Infeasible candidates are no longer repaired. There was no difference in performance keeping/removing this.

Because `target_feasible=0.85` is relatively high in ILS and the penalty updates are smoother, the penalised costs of infeasible solutions do not significantly affect MBAT's threshold value.

pyvrp/solve.py
```diff
@@ -203,12 +208,11 @@ def solve(
...
     for perturb_op in params.perturbation_ops:
-        ls.add_perturbation_operator(
-            perturb_op(data, params.num_perturbations)
-        )
+        if perturb_op.supports(data):
+            ls.add_perturbation_operator(perturb_op(data))
 
     pm = PenaltyManager.init_from(data, params.penalty)
-    init = Solution(data, [])  # type: ignore
+    init = ls(Solution(data, []), pm.booster_cost_evaluator())  # type: ignore
```
leonlan, line 215: Bit hacky, but this helps ILS to start with feasible solutions. This is particularly useful for tests that use a low number of iterations, because then the initial penalties don't get updated (enough).

pyvrp/IteratedLocalSearch.py
leonlan, line 1: I slipped this change in here as well because it helps with performance: 
1. Infeasible candidates are also accepted. This helps quite a bit in TW instances.
2. Infeasible candidates are no longer repaired. There was no difference in performance keeping/removing this.

Because `target_feasible=0.90` is relatively high in ILS and the penalty updates are smoother, the penalised costs of infeasible solutions do not significantly affect MBAT's threshold value.

pyvrp/cpp/search/NeighbourRemoval.cpp
leonlan, line 1: One issue with the old `NeighbourRemoval` was that it always removed the same clients. That's OK for early stages of search, but once we get close to "optimal", we need something that is adaptive. This PR changes the neighbour removal operator to delete each neighbour U, but also p(U) and n(U).

I also tried a string removal perturbation operator. This has similar performance, but I find this version of neighbour removals simpler.

pyvrp/cpp/search/OptionalInsert.cpp
leonlan, line 1: This operator is really necessary for prize-collecting VRPs, which have local optima that are much harder to escape from. Solving prize-collecting with NeighbourRemoval only does not work. Instead, this operator that "force inserts" optional clients into the solutions.

It's also important to insert into empty routes here. I'm not sure why exactly.

pyvrp/IteratedLocalSearch.py
N-Wouda: Is the `_booster_cost_evaluator` property still being used anywhere now?

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -169,8 +170,16 @@ void LocalSearch::perturb(CostEvaluator const &costEvaluator)
...
-    (*perturbOps[0])(
-        nodes, routes, costEvaluator, neighbours_, orderNodes, promising);
+    PerturbationContext context{nodes,
+                                routes,
+                                costEvaluator,
+                                neighbours_,
+                                orderNodes,
+                                orderVehTypes,
+                                promising,
+                                numPerturbations_};
```
N-Wouda: Yeah we need to figure out a way to reduce this

pyvrp/cpp/search/LocalSearch.h
```diff
@@ -56,6 +56,8 @@ class LocalSearch
     // numLocations, but nothing is stored for the depots!)
     Neighbours neighbours_;
 
+    size_t numPerturbations_ = 0;  // number of perturbations to apply
```
N-Wouda, line 59: I still don't really understand what this means or does. Reading it, I think it means the number of times we call the perturbation operators. But that's not it, it's more of a "degree of destruction" kind of thing.

pyvrp/cpp/search/OptionalInsert.cpp
```diff
@@ -0,0 +1,82 @@
+#include "OptionalInsert.h"
+#include "primitives.h"
+
+using pyvrp::search::OptionalInsert;
+
+void OptionalInsert::operator()(PerturbationContext const &context)
```
N-Wouda: There's lots of duplication here with `LocalSearch::insert`, I think. Can we somehow merge/abstract the relevant bits?

N-Wouda: (do we still have `repair::greedyRepair` in the ILS branch? Because that also somewhat duplicates this logic)

pyvrp/cpp/search/OptionalInsert.h
```diff
@@ -0,0 +1,34 @@
...
+
+public:
+    /**
+     * Creates an optional insert operator.
+     *
+     * Parameters
+     * ----------
+     * data
+     *     Problem data instance.
+     */
```
N-Wouda: This needs to be in the class-level docstring, if necessary. We do not document constructors like this (because it's never used in the bindings, only the class-level docstring is).

Class-level docstring also needs a first line type hint,
```
 * OptionalInsert(data: ProblemData)
```
should do

N-Wouda: Oh, I missed this with the other operators. Could you also update those?

pyvrp/cpp/search/PerturbationOperator.h
```diff
@@ -3,12 +3,28 @@
...
+{
+    std::vector<search::Route::Node> &nodes;
+    std::vector<search::Route> &routes;
+    CostEvaluator const &costEvaluator;
+    std::vector<std::vector<size_t>> const &neighbours;
+    std::vector<size_t> const &orderNodes;
+    std::vector<std::pair<size_t, size_t>> const &orderVehTypes;
+    DynamicBitset &promising;
+    size_t const numPerturbations;
+};
```
N-Wouda: I think this is OK to have for now as we still figure out how to develop ILS, but I'd like to get rid of this again at some point.

pyvrp/search/_search.pyi
```diff
@@ -36,14 +36,10 @@ class RouteOperator:
...
-    def __call__(
-        self,
-        nodes: list[Node],
-        routes: list[Route],
-        cost_evaluator: CostEvaluator,
-        neighbours: list[list[int]],
-        order_nodes: list[int],
-    ) -> None: ...
+    def __init__(self, data: ProblemData) -> None: ...
+    def __call__(self, context) -> None: ...
```
N-Wouda, line 40: Do we explain what the context is? Can we access it from Python (for tests?)?

pyvrp/solve.py
```diff
@@ -64,7 +64,7 @@ def __init__(
         perturbation_ops: list[
             type[PerturbationOperator]
         ] = PERTURBATION_OPERATORS,
-        num_perturbations: int = 10,
+        num_perturbations: int = 25,
```
N-Wouda, line 67: Should this be part of some `PerturbationParams` object, together with the perturb ops? It doesn't feel as independent from other parameters as, say, the `display_interval` argument did.

pyvrp/cpp/search/LocalSearch.h
leonlan: Yeah, it's more like a degree of destruction if we would be only removing clients. But we also perturb by inserting clients which is why I named it like this (but we should find a more appropriate name, I agree).

pyvrp/cpp/search/OptionalInsert.cpp
leonlan: I think `repair` can be completely removed with ILS.

pyvrp/search/_search.pyi
leonlan: I didn't bother providing bindings for this, because I'm hoping we can remove it in the next PRs. Do you think we should add it?

pyvrp/IteratedLocalSearch.py
leonlan: I only use it now to search around the initial solution (in `solve.py`), because the penalties don't update fast enough to obtain a feasible solution in tests with low number of iterations. I think we can remove it -- but we need another mechanism to make the initial solution (close to) feasible.

pyvrp/search/_search.pyi
N-Wouda: No should be OK if we refactor this anyway

pyvrp/IteratedLocalSearch.py
N-Wouda: No, I meant the property of the `IteratedLocalSearch` class - not the `PenaltyManager` method. I think the property is unused. But if we can remove it altogether somehow that might be even nicer.

pyvrp/cpp/search/OptionalInsert.cpp
leonlan: I made an attempt by introducing the `bestInsert` function at https://github.com/PyVRP/PyVRP/pull/877/commits/9e495cb7f33eb9e616565f7ab184da246fe49bc5. I'm not sure if this is better. We will only have two places where we use insertion (`LS::insert` and `InsertOptional`) so it's not that much duplication.

pyvrp/solve.py
leonlan: Getting the perturbation operators into PerturbationParams felt a bit tricky. Some things that came up during my attempt:
- What's the default for perturbation operators? I'd say all perturbation operators, but this results in LS having by default perturbation operators while having no node operators.
- When do we add perturbation operators if they are part of the params? If we add at construction then we get the above issue.

I think it's better to keep the perturbation_ops like the other operators, but `PerturbationParams` is ok to have to store the `num_perturbations` parameter.

pyvrp/cpp/search/primitives.h
```diff
@@ -80,6 +81,39 @@ Cost inplaceCost(Route::Node *U,
...
+ * neighbours
+ *     Neighbourhood structure of each client.
+ * nodes
+ *     Nodes in the solution.
+ * routes
+ *     Routes in the solution.
+ *
+ * Returns
+ * -------
+ * std::pair<Route::Node *, Cost>
```
N-Wouda: `tuple[Node, int]`?

pyvrp/cpp/search/bindings.cpp
```diff
@@ -536,4 +541,15 @@ PYBIND11_MODULE(_search, m)
           py::arg("data"),
           py::arg("cost_evaluator"),
           DOC(pyvrp, search, removeCost));
+
+    m.def("best_insert",
+          &bestInsert,
+          py::return_value_policy::copy,
```
N-Wouda: This copies the.. `std::pair`? Is that needed?

pyvrp/cpp/search/primitives.h
```diff
@@ -2,6 +2,7 @@
 #define PYVRP_SEARCH_PRIMITIVES_H
 
 #include "CostEvaluator.h"
+#include "DynamicBitset.h"
```
N-Wouda: Where is this used? I don't see it in the header.

pyvrp/cpp/search/InsertOptional.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include "PerturbationOperator.h"
+#include "ProblemData.h"
+
+namespace pyvrp::search
+{
+/**
+ * InsertOptional(data: ProblemData)
+ *
+ * This operator forcefully inserts optional clients into the solution.
```
N-Wouda:
```suggestion
 * This operator inserts optional clients into the solution.
```
Are we applying force? 😆

pyvrp/cpp/search/RemoveNeighbours.cpp
```diff
@@ -0,0 +1,39 @@
...
+            if (node->isDepot())
+                continue;
+
+            context.promising[node->client()] = true;
+            context.promising[p(node)->client()] = true;
+            context.promising[n(node)->client()] = true;
+
+            auto *route = node->route();
+            route->remove(node->idx());
+            route->update();
```
N-Wouda: This marks nodes as promising multiple times (but that's not terribly slow), but also calls `Route::update` repeatedly. That seems unnecessary. Perhaps something like
```cpp

auto *route = U->route();

for (auto *node : {n(U), U, p(U)})
{
  if (node->isDepot())
    continue;

  context.promising[p(node)->client()] = true;
  context.promising[node->client()] = true;
  context.promising[n(node)->client()] = true;

  route->remove(node->idx());
}

route->update();
```

We only need to update once since no (cached) attributes are used here during removal.

pyvrp/cpp/search/primitives.h
```diff
@@ -80,6 +81,39 @@ Cost inplaceCost(Route::Node *U,
 Cost removeCost(Route::Node *U,
                 ProblemData const &data,
                 CostEvaluator const &costEvaluator);
+
+/**
+ * Finds the best insertion position for U among its neighbors and empty routes.
```
N-Wouda:
```suggestion
 * Finds the best insertion position for U among its neighbours and in empty routes.
```

pyvrp/search/LocalSearch.py
```diff
@@ -12,6 +11,30 @@
...
+
+    Parameters
+    ----------
+    num_perturbations
+        Maximum number of perturbations to apply in each iteration. The actual
+        number of perturbations used is randomly chosen between 1 and this
+        value (inclusive). Default is 25.
+    """
+
+    num_perturbations: int = 25
```
N-Wouda: Yeah this might not be it, for just one parameter. Are we expecting to have more?

tests/test_solve.py
```diff
@@ -84,7 +86,7 @@ def test_solve_custom_params(rc208):
     expected by checking how solutions are accepted.
     """
 
-    def monotonically_decreasing(arr) -> np.bool:
+    def monotonically_decreasing(arr) -> np.bool_:
```
N-Wouda, line 87: `np.bool` and `np.bool_` are aliases, right? Is there a reason to prefer `np.bool_`?

leonlan: I'm a bit confused myself here. I got this error on numpy v1.26:
```
E           AttributeError: module 'numpy' has no attribute 'bool'.
````

Turns out that: 
- `np.bool` got deprecated in [v1.20](https://numpy.org/doc/2.0/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated)
- Removed in [v1.24](https://numpy.org/doc/2.0/release/1.24.0-notes.html), and
- Added back again as dtype in [v2.0](https://numpy.org/devdocs/release/2.0.0-notes.html#changes).

So there's a little gap in versions where `np.bool` won't work, so I think it's best to keep it as `np.bool_` for now.

pyvrp/search/LocalSearch.py
leonlan: We could always add more but I think this is the simplest implementation that works as well as HGS. I'm leaning toward keeping ILS as simple as possible to make future changes easier to address.

tests/test_solve.py
N-Wouda: Huh, what a roller coaster. Thanks for figuring this out!

N-Wouda (approved):
LGTM, let's refactor everything in the feature branch (not here)

PR #883, opened 2025-08-05
Add missing value argument

### LINKED ISSUE #683: Solver stops working
I'm not sure if it's a bug or if I'm doing something wrong. I have a larger model with about 150 nodes where the solver stops working after about 1 minute working. I have tried to identify the problem and transfer it to a smaller example. 
**Note** Due to the max duration, there should be no solution. Without using client_groups, the solver runs through and finds no solution. If clients_groups are used, it stops working.

**Code Sample**
```
from pyvrp import Model
from pyvrp.show_versions import show_versions

show_versions()

m = Model()
 
# If the group is set to None, the example works
Group = m.add_client_group(False) # None

depot = m.add_depot(x=0, y=0)

A = m.add_client(x=0, y=0, prize=1, required=False, group=None)
B = m.add_client(x=0, y=0, prize=1, required=False, group=None)
C = m.add_client(x=0, y=0, prize=1, required=False, group=None)
D = m.add_client(x=0, y=0, prize=1, required=False, group=Group)

#If the group is used and max duration is removed, it also works
m.add_vehicle_type(1, start_depot=depot, end_depot=depot, max_duration=20)

m.add_edge(depot, A, distance=0,duration=5)
m.add_edge(A, B, distance=0,duration=5)
m.add_edge(B, C, distance=0,duration=5)
m.add_edge(C, D, distance=0,duration=5)
m.add_edge(D, depot, distance=0,duration=5)
 
res = m.solve(stop=MaxRuntime(60), display=True)

```

**Output:**

```
INSTALLED VERSIONS
------------------
     pyvrp: 0.10.1
     numpy: 2.1.3
matplotlib: 3.9.2
    vrplib: 1.5.1
      tqdm: 4.67.0
     tomli: 2.0.1
    Python: 3.11.10
PyVRP v0.10.1

Solving an instance with:
    1 depot
    4 clients
    1 vehicle (1 vehicle type)

H     500      1s |  37        4        4 |  37 321337919358326464       94

```

The solver stops working without further output. Only when I remove the max_duration or the group the  solver works normally again.
In my larger example, I use more vehicles. I think the solver stops working after a while because it doesn't find solutions for all vehicles

**Note**: I also had the same problem with the older version 0.9.1

N-Wouda:
This PR:

- [x] Closes #683.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/Model.py
+++ b/pyvrp/Model.py
@@ -416,19 +416,25 @@ class Model:
         self._vehicle_types.append(vehicle_type)
         return vehicle_type
 
-    def data(self) -> ProblemData:
+    def data(self, missing_value: int = MAX_VALUE) -> ProblemData:
         """
         Creates and returns a :class:`~pyvrp._pyvrp.ProblemData` instance
         from this model's attributes.
+
+        Parameters
+        ----------
+        missing_value
+            Distance and duration value to use for missing edges. Defaults to
+            ``MAX_VALUE``, a large number.
         """
         locs = self.locations
         loc2idx = {id(loc): idx for idx, loc in enumerate(locs)}
 
         # First we create the base distance and duration matrices. These are
-        # shared by all routing profiles. If an edge was not specified, we use
-        # a large default value here.
-        base_distance = np.full((len(locs), len(locs)), MAX_VALUE, np.int64)
-        base_duration = np.full((len(locs), len(locs)), MAX_VALUE, np.int64)
+        # shared by all routing profiles.
+        shape = (len(locs), len(locs))
+        base_distance = np.full(shape, missing_value, np.int64)
+        base_duration = np.full(shape, missing_value, np.int64)
         np.fill_diagonal(base_distance, 0)
         np.fill_diagonal(base_duration, 0)
 
@@ -477,6 +483,7 @@ class Model:
         collect_stats: bool = True,
         display: bool = True,
         params: SolveParams = SolveParams(),
+        missing_value: int = MAX_VALUE,
     ) -> Result:
         """
         Solve this model.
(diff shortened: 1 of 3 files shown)


pyvrp/Model.py
```diff
@@ -416,19 +416,26 @@ def add_vehicle_type(
...
         """
         Creates and returns a :class:`~pyvrp._pyvrp.ProblemData` instance
         from this model's attributes.
+
+        Parameters
+        ----------
+        missing_value
+            Distance and duration value to use for missing edges. Defaults to
+            ``MAX_VALUE``, a large number. Note that this value cannot exceed
+            ``MAX_VALUE``.
```
leonlan: Should we also document `MAX_VALUE` in the API reference? It's better to be explicit about "the large value" so that the user also knows what the upper bound is.

```suggestion
            Distance and duration value to use for missing edges. Defaults to
            :const:`~pyvrp.constants.MAX_VALUE`. Note that this value cannot exceed
            :const:`~pyvrp.constants.MAX_VALUE`.
```

and adding this to `api/pyvrp.rst`:

```
.. automodule:: pyvrp.constants
   :members:
```

pyvrp/Model.py
```diff
@@ -496,14 +504,18 @@ def solve(
             ``collect_stats`` is also set, which it is by default.
         params
             Solver parameters to use. If not provided, a default will be used.
+        missing_value
+            Distance and duration value to use for missing edges. Defaults to
+            ``MAX_VALUE``, a large number.
```
leonlan: Per my comment above:

```suggestion
            :const:`~pyvrp.constants.MAX_VALUE`.
```

leonlan (approved):

PR #886, opened 2025-08-21
Construct initial solution and pass initial solution to params

### LINKED ISSUE #855: Support starting from known feasible/good solutions
- now that you're talking about initialization, it might also be good to allow the user to provide an initial solution, PyVRP can then guarantee to return a solution that is not worse than the initial solution. In my case which is what we call a reoptimization, the initial solution will be the solution to the original optimization.

_Originally posted by @ravenblackdusk in https://github.com/PyVRP/PyVRP/discussions/564#discussioncomment-13340383_

leonlan:
This PR:
- Removes `PenaltyManager.repair_booster` and introduces `PenaltyManager.max_cost_evaluator()`.
- Allows initial solutions to be passed to `SolveParams`.
- Constructs an initial solution using `PenaltyManager.max_cost_evaluator()`.
- Changes `LocalSearch::insert` to insert into empty routes.
- [Breaking] Moves `collect_stats` to `SolveParams`

Furthermore,
- [x] Is part of #778 and closes #855.

Performance compared to `feat/ils` target branch.

**1 min**
| Commit   | MDVRPTW | HFVRP  | GVRP   | VRPB   | MTVRPTWR | CVRP   | PCVRPTW | VRPTW  |
|----------|---------|--------|--------|--------|----------|--------|---------|--------|
| 34be5c0  | 2.08%   | inf%   | 0.45%  | 1.35%  | 1.50%    | 0.85%  | 2.54%   | inf%   |
| (current)| 2.11%   | 1.51%  | 0.44%  | 1.36%  | 1.08%    | 0.63%  | 2.48%   | 1.95%  |
| **Diff** | **+0.03%** | **-inf** | **-0.01%** | **+0.01%** | **-0.42%** | **-0.22%** | **-0.06%** | **-inf** |

**10 min**
| Commit   | MDVRPTW | HFVRP  | GVRP   | VRPB   | MTVRPTWR | CVRP   | PCVRPTW | VRPTW  |
|----------|---------|--------|--------|--------|----------|--------|---------|--------|
| 34be5c0  | 0.87%   | 0.72%  | 0.13%  | 0.57%  | 0.52%    | 0.38%  | 0.76%   | 1.03%  |
| (current)| 0.88%   | 0.59%  | 0.11%  | 0.66%  | 0.37%    | 0.30%  | 0.76%   | 0.91%  |
| **Diff** | **+0.01%** | **-0.13%** | **-0.02%** | **+0.09%** | **-0.15%** | **-0.08%** | **±0.00%** | **-0.12%** |

**Academic**
| Commit   | MDVRPTW | HFVRP  | GVRP   | VRPB   | MTVRPTWR | CVRP   | PCVRPTW | VRPTW  |
|----------|---------|--------|--------|--------|----------|--------|---------|--------|
| 34be5c0  | 0.48%   | 0.59%  | 0.09%  | 0.42%  | 0.52%    | 0.32%  | 0.35%   | 0.64%  |
| (current)| 0.52%   | 0.47%  | 0.11%  | 0.48%  | 0.35%    | 0.25%  | 0.37%   | 0.58%  |
| **Diff** | **+0.04%** | **-0.12%** | **+0.02%** | **+0.06%** | **-0.17%** | **-0.07%** | **+0.02%** | **-0.06%** |

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/Model.py
+++ b/pyvrp/Model.py
@@ -484,9 +484,9 @@ class Model:
         seed: int = 0,
         collect_stats: bool = True,
         display: bool = True,
-        initial_solution: Solution | None = None,
         params: SolveParams = SolveParams(),
         missing_value: int = MAX_VALUE,
+        initial_solution: Solution | None = None,
     ) -> Result:
         """
         Solve this model.
@@ -504,14 +504,14 @@ class Model:
             Whether to display information about the solver progress. Default
             ``True``. Progress information is only available when
             ``collect_stats`` is also set, which it is by default.
-        initial_solution
-            Solution to start the search from. If not provided, a default
-            solution will be created.
         params
             Solver parameters to use. If not provided, a default will be used.
         missing_value
             Distance and duration value to use for missing edges. Defaults to
             :const:`~pyvrp.constants.MAX_VALUE`, a large number.
+        initial_solution
+            Solution to start the search from. If not provided, a default
+            solution will be created.
 
         Returns
         -------
@@ -525,8 +525,8 @@ class Model:
             seed,
             collect_stats,
             display,
-            initial_solution,
             params,
+            initial_solution,
         )
 
 
--- a/pyvrp/Model.py
+++ b/pyvrp/Model.py
@@ -505,8 +505,8 @@ class Model:
             ``True``. Progress information is only available when
             ``collect_stats`` is also set, which it is by default.
         initial_solution
-            Initial solution to start the search from. If not provided, a
-            default initial solution will be created.
+            Solution to start the search from. If not provided, a default
+            solution will be created.
         params
             Solver parameters to use. If not provided, a default will be used.
         missing_value
--- a/pyvrp/Model.py
+++ b/pyvrp/Model.py
@@ -10,6 +10,7 @@ from pyvrp._pyvrp import (
     ClientGroup,
     Depot,
     ProblemData,
+    Solution,
     VehicleType,
 )
 from pyvrp.constants import MAX_VALUE
@@ -483,6 +484,7 @@ class Model:
         seed: int = 0,
         collect_stats: bool = True,
         display: bool = True,
+        initial_solution: Solution | None = None,
         params: SolveParams = SolveParams(),
         missing_value: int = MAX_VALUE,
     ) -> Result:
@@ -502,6 +504,9 @@ class Model:
             Whether to display information about the solver progress. Default
             ``True``. Progress information is only available when
             ``collect_stats`` is also set, which it is by default.
+        initial_solution
+            Initial solution to start the search from. If not provided, a
+            default initial solution will be created.
         params
             Solver parameters to use. If not provided, a default will be used.
         missing_value
@@ -514,8 +519,15 @@ class Model:
             A Result object, containing statistics (if collected) and the best
             found solution.
         """
-        data = self.data(missing_value)
-        return solve(data, stop, seed, collect_stats, display, params)
+        return solve(
+            self.data(missing_value),
+            stop,
+            seed,
+            collect_stats,
+            display,
+            initial_solution,
+            params,
+        )
 
 
 def _idx_by_id(item: object, container: Sequence[object]) -> int | None:
--- a/pyvrp/Model.py
+++ b/pyvrp/Model.py
@@ -10,7 +10,6 @@ from pyvrp._pyvrp import (
     ClientGroup,
     Depot,
     ProblemData,
-    Solution,
     VehicleType,
 )
 from pyvrp.constants import MAX_VALUE
@@ -485,7 +484,6 @@ class Model:
         display: bool = True,
         params: SolveParams = SolveParams(),
         missing_value: int = MAX_VALUE,
-        initial_solution: Solution | None = None,
     ) -> Result:
         """
         Solve this model.
@@ -505,9 +503,6 @@ class Model:
         missing_value
             Distance and duration value to use for missing edges. Defaults to
             :const:`~pyvrp.constants.MAX_VALUE`, a large number.
-        initial_solution
-            Solution to start the search from. If not provided, a default
-            solution will be created.
 
         Returns
         -------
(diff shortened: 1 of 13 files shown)


tests/test_Model.py
```diff
@@ -1064,7 +1064,7 @@ def test_instance_with_multi_trip_and_release_times(mtvrptw_release_times):
            https://doi.org/10.1287/trsc.2022.1161.
     """
     m = Model.from_data(mtvrptw_release_times)
-    res = m.solve(stop=MaxIterations(20))
+    res = m.solve(stop=MaxIterations(50))
```
leonlan: I just realized that `LS::insert` is pretty suboptimal for empty solutions, because it defaults to `routes[0][0]` as a fallback if there are no neighbors. As a consequence, what always happens in the 1st search iteration is that _all_ nodes are inserted in the first route.

leonlan: I propose the following for `LS::insert`:
- Evaluate inserting after all neighbors.
- Insert after all empty routes as a fallback.
- Insert in `routes[0][0]` as fallback if required.

pyvrp/cpp/search/LocalSearch.cpp
leonlan, line 1: This addition makes LS::insert work as expected for (empty) initial solutions. Before this change, the first LS search iteration would insert all clients to `routes[0]`, resulting in a TSP solution. The fix is to allow inserting into empty routes to create a proper VRP solution during the first search iteration.

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -403,6 +404,32 @@ void LocalSearch::insert(Route::Node *U,
...
+
+        auto const cost = insertCost(U, (*empty)[0], data, costEvaluator);
+        if (cost < bestCost)
+        {
+            bestCost = cost;
+            UAfter = (*empty)[0];
+        }
+
+        break;
+    }
```
leonlan, line 428: This addition makes `LS::insert` work as expected for (empty) initial solutions. Before this change, the first LS search iteration would insert all clients to `routes[0]`, resulting in a TSP solution. The fix is to allow inserting into empty routes to create a proper VRP solution during the first search iteration.

pyvrp/Model.py
```diff
@@ -507,15 +509,25 @@ def solve(
...
-        data = self.data(missing_value)
-        return solve(data, stop, seed, collect_stats, display, params)
+        return solve(
+            self.data(missing_value),
+            stop,
+            seed,
+            collect_stats,
+            display,
+            params,
+            initial_solution,
```
N-Wouda: Should we make the initial solution part of `params`?

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -403,6 +404,32 @@ void LocalSearch::insert(Route::Node *U,
...
+
+    if (required && !UAfter)
+        UAfter = routes[0][0];  // fallback if no insertion positions were found
```
N-Wouda: I think we can do this check inside the next if statement, something like
```cpp
if (!UAfter)
  UAfter = routes[0][0];
```
That should simplify the logic a little bit.

pyvrp/solve.py
```diff
@@ -212,8 +221,17 @@ def solve(
...
 
-    ils_args = (data, pm, rng, ls, init, params.ils)
+    if initial_solution is None:
+        # Using large penalties will help to find a feasible solution.
+        max_penalty = params.penalty.max_penalty
+        cost_eval = CostEvaluator(
+            [max_penalty] * data.num_load_dimensions,
+            max_penalty,
+            max_penalty,
+        )
```
N-Wouda: Should we move this into a method on the `PenaltyManager`? Perhaps something like a `fixed_cost_evaluator` with a `penalty: int` value argument? We could then do something like
```python
cost_eval = pm.fixed_cost_evaluator(params.penalty.max_penalty)
```
or similar.

N-Wouda: I think what bothers me about this is that we're now directly constructing a cost evaluator, whereas we already have an object managing all that.

N-Wouda: Or a `pm.max_cost_evaluator()` method, if we don't want to be able to specify the values and just want an evaluator with max penalties.

pyvrp/Model.py
```diff
@@ -495,13 +494,10 @@ def solve(
...
         seed
             Seed value to use for the random number stream. Default 0.
-        collect_stats
-            Whether to collect statistics about the solver's progress. Default
-            ``True``.
         display
             Whether to display information about the solver progress. Default
             ``True``. Progress information is only available when
-            ``collect_stats`` is also set, which it is by default.
+            ``params.collect_stats`` is also set, which it is by default.
```
N-Wouda: Maybe link to the attribute?

pyvrp/solve.py
```diff
@@ -121,10 +133,23 @@ def num_perturbations(self):
...
+
     @classmethod
     def from_file(cls, loc: str | pathlib.Path):
         """
         Loads the solver parameters from a TOML file.
+
+        Note
+        ----
+        The initial solution cannot be loaded from file, and will always be set
+        to ``None``.
```
N-Wouda:
```suggestion
        .. note::
           The initial solution cannot be loaded from file, and will always be
           set to ``None``.
```

pyvrp/solve.py
```diff
@@ -174,13 +200,10 @@ def solve(
...
     seed
         Seed value to use for the random number stream. Default 0.
-    collect_stats
-        Whether to collect statistics about the solver's progress. Default
-        ``True``.
     display
         Whether to display information about the solver progress. Default
         ``False``. Progress information is only available when
-        ``collect_stats`` is also set, which it is by default.
+        ``params.collect_stats`` is also set, which it is by default.
```
N-Wouda: Link to relevant attribute in docs?

pyvrp/solve.py
```diff
@@ -212,8 +235,16 @@ def solve(
             ls.add_perturbation_operator(perturb_op(data))
 
     pm = PenaltyManager.init_from(data, params.penalty)
-    init = ls(Solution(data, []), pm.booster_cost_evaluator())  # type: ignore
+    init = (
+        params.initial_solution
+        if params.initial_solution is not None
+        else ls(Solution(data, []), pm.max_cost_evaluator())  # type: ignore
+    )
```
N-Wouda: I think this'd be clearer with something like
```python
init = params.initial_solution
if init is None:
  init = ls(...)
```

N-Wouda (approved):
Very small documentation stuff

Issue #207, opened 2023-04-07
Tests fail on Mac OS


wouterkool:
#### Describe the bug

Running on Mac OS 13.2.1 and Python 3.11, several test cases fail. While I expect the plotting tests to fail, also some functional tests fail. I need to further investigate this but I'm having issues with attaching the debugger on Mac OS.

#### Code sample

```zsh
poetry run pytest
```

For full output, click details below:
<details>

```zsh
(pyvrp-py3.11) wouterkool@Wouters-MacBook-Pro PyVRP % pytest
========================================================================== test session starts ===========================================================================
platform darwin -- Python 3.11.2, pytest-7.2.2, pluggy-1.0.0
rootdir: /Users/wouterkool/dev/PyVRP, configfile: pyproject.toml, testpaths: pyvrp
plugins: cov-4.0.0, anyio-3.6.2
collected 254 items                                                                                                                                                      

pyvrp/crossover/tests/test_selective_route_exchange.py .............                                                                                               [  5%]
pyvrp/diversity/tests/test_broken_pairs_distance.py .                                                                                                              [  5%]
pyvrp/educate/tests/test_Exchange.py ..................F                                                                                                           [ 12%]
pyvrp/educate/tests/test_LocalSearch.py ............                                                                                                               [ 17%]
pyvrp/educate/tests/test_MoveTwoClientsReversed.py F........                                                                                                       [ 21%]
pyvrp/educate/tests/test_RelocateStar.py ....                                                                                                                      [ 22%]
pyvrp/educate/tests/test_SwapStar.py .........                                                                                                                     [ 26%]
pyvrp/educate/tests/test_TwoOpt.py .........                                                                                                                       [ 29%]
pyvrp/educate/tests/test_neighbourhood.py ........                                                                                                                 [ 33%]
pyvrp/plotting/tests/test_plotting.py ..FF.                                                                                                                        [ 35%]
pyvrp/stop/tests/test_MaxIterations.py ........                                                                                                                    [ 38%]
pyvrp/stop/tests/test_MaxRuntime.py ............                                                                                                                   [ 42%]
pyvrp/stop/tests/test_NoImprovement.py ...........                                                                                                                 [ 47%]
pyvrp/stop/tests/test_TimedNoImprovement.py ................                                                                                                       [ 53%]
pyvrp/tests/test_CostEvaluator.py .........                                                                                                                        [ 57%]
[...]

PR #889, opened 2025-08-26
Implement `RandomNumberGenerator::shuffle`

leonlan:
This PR implements the shuffle method for `RandomNumberGenerator`.

The motivation behind this is that each compiler implements `std::shuffle` differently, which results in different outputs between my local environment (Mac) and the CI (Linux). This has come up multiple times before (https://github.com/PyVRP/PyVRP/issues/207, https://github.com/PyVRP/PyVRP/pull/431). It got me frustrated this time so I finally decided to find a solution to this problem :-).

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/RandomNumberGenerator.h
leonlan, line 1: I didn't add tests because adding the bindings for this method would be quite complex. But I did check that this shuffles with a simple print statement :-)

pyvrp/cpp/RandomNumberGenerator.h
```diff
@@ -88,6 +97,13 @@ RandomNumberGenerator::result_type RandomNumberGenerator::randint(T high)
     static_assert(std::is_integral<T>::value);
     return operator()() % high;
 }
+
+template <typename RandomIt>
+void RandomNumberGenerator::shuffle(RandomIt first, RandomIt last)
+{
+    for (auto idx = last - first - 1; idx > 0; --idx)
+        std::swap(first[idx], first[randint(idx + 1)]);
```
N-Wouda, line 106: Perhaps add a link to [this](https://en.cppreference.com/w/cpp/algorithm/random_shuffle.html#Version_2)?

N-Wouda: (or at least some backlink to `std::shuffle` on https://en.cppreference.com, because it's helpful to know where this came from for correctness, since we don't have test support here)

pyvrp/cpp/RandomNumberGenerator.h
N-Wouda: Should be OK.

PR #898, opened 2025-09-23
Print lines missing coverage when running tests

N-Wouda:
Turns out there's also a `term-missing` option, instead of just `term`. The extra information about lines missing coverage seems useful.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/subprojects/.gitignore
+++ b/subprojects/.gitignore
@@ -1,1 +1,2 @@
 */
+*.wraplock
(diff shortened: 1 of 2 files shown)


subprojects/.gitignore
```diff
@@ -1 +1,2 @@
 */
+*.wraplock
```
N-Wouda, line 2: Meson added a lock file a while ago, but I think we can just ignore that - [they do the same](https://github.com/mesonbuild/meson/pull/13854/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947). Guess it's to ensure only one process (holding the lock file) is resolving stuff at the same time.

PR #908, opened 2025-10-02
Floating-point Coordinate type

### LINKED ISSUE #907: Allow different value types for different measures
Currently, we have a `Coordinate` type (a `Measure`, so underlying is an `int64`). But coordinates are often just lat or lon, which are typically just expressed as doubles. They're not really used in computations (except for centroids, which cast them to doubles anyway), and are mostly for plotting. I think we can relax these to just be doubles, if we allow the underlying value type of `Measure` to vary.

N-Wouda:
This PR:

- [x] Closes #907.

TODO:

- [x] Clean up
- [x] Retype `std::pair<double, double> centroid` to `std::pair<Coordinate, Coordinate>`

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/Route.cpp
+++ b/pyvrp/cpp/Route.cpp
@@ -7,6 +7,7 @@
 #include <fstream>
 #include <numeric>
 
+using pyvrp::Coordinate;
 using pyvrp::Cost;
 using pyvrp::Distance;
 using pyvrp::Duration;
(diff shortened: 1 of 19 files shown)


N-Wouda:
Fairly wonky stub for this idea. Needs more work but seems to be functional already.

pyvrp/cpp/Route.cpp
```diff
@@ -218,8 +219,9 @@ Route::Route(ProblemData const &data, Trips trips, size_t vehType)
         prizes_ += trip.prizes();
 
         auto const [x, y] = trip.centroid();
-        centroid_.first += (x * trip.size()) / size();
-        centroid_.second += (y * trip.size()) / size();
+        auto const div = std::max<size_t>(size(), 1);  // max 1 avoids zero div
+        centroid_.first += (x.get() * trip.size()) / div;
+        centroid_.second += (y.get() * trip.size()) / div;
```
N-Wouda: I think this was a subtle bug, because `size()` can be zero for empty routes.

N-Wouda: Haha, yeah, this was bugged:

image

N-Wouda:
Forgot about the tutorial, thanks!

PR #917, opened 2025-10-17
Add overtime cost

### LINKED ISSUE #913: Overtime cost
Hi !, 
I’m working on a VRP variant where all customers must be served, there are no time window constraints, and vehicles are uncapacitated. Each vehicle has a shift duration limit, but this is a soft constraint: if the route duration exceeds the limit, the route remains feasible, but an overtime cost is incurred. The overtime cost is computed as

max (0, route duration − shift limit) × 𝛽, where 𝛽 is the overtime cost per unit of time.

The objective is to minimize the total travel time plus overtime costs. I’m interested in understanding whether this variant can already be modeled in PyVRP or, if not, what code modifications might be required to support it (e.g., incorporating overtime as a soft constraint in the cost evaluation). Thank you!

N-Wouda:
This PR:

- [x] Closes #913.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/CostEvaluator.h
+++ b/pyvrp/cpp/CostEvaluator.h
@@ -121,27 +121,30 @@ public:
 
     /**
      * Hand-waving some details, each solution consists of a set of non-empty
-     * routes :math:`\mathcal{R}`. Each route :math:`R \in \mathcal{R}` is a
-     * sequence of edges, starting and ending at a depot. Each route :math:`R`
-     * has an assigned vehicle type, through which the route is equipped with a
-     * fixed vehicle cost :math:`f_R`, and unit distance and duration costs
-     * :math:`c^\text{distance}_R` and :math:`c^\text{duration}_R`,
-     * respectively. Let :math:`V_R = \{i : (i, j) \in R \}` be the set of
-     * locations visited by route :math:`R`, and :math:`d_R` and :math:`t_R`
-     * the total route distance and duration, respectively. The objective value
+     * routes :math:`\mathcal{R}`. Each route :math:`R \in \mathcal{R}` can be
+     * represented as a sequence of edges, starting and ending at a depot. A
+     * route :math:`R` has an assigned vehicle type that equips the route with
+     * fixed vehicle cost :math:`f_R`, and unit distance, duration and overtime
+     * costs :math:`c^\text{distance}_R`, :math:`c^\text{duration}_R`,
+     * :math:`c^\text{overtime}_R`, respectively. Let
+     * :math:`V_R = \{i : (i, j) \in R \}` be the set of locations visited by
+     * route :math:`R`, and :math:`d_R`, :math:`t_R`, and :math:`o_R` the total
+     * route distance, duration, and overtime, respectively. The objective value
      * is then given by
      *
      * .. math::
      *
      *    \sum_{R \in \mathcal{R}}
      *      \left[
-     *          f_R + c^\text{distance}_R d_R + c^\text{duration}_R t_R
+     *          f_R + c^\text{distance}_R d_R
+     *              + c^\text{duration}_R t_R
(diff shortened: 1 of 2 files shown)


N-Wouda:
This now seems functional but I want to tinker with it a bit more before it's ready to merge.

pyvrp/cpp/CostEvaluator.h
leonlan, line 1: `CostEvaluator.cost`'s docstring also needs to be updated.
image

tests/test_Solution.py
```diff
@@ -887,3 +886,28 @@ def test_distance_duration_cost_calculations(ok_small):
...
+    # overtime, if needed. This route takes 5'229 to complete, so the route
+    # should have 229 units of overtime.
+    route = Route(ok_small_overtime, [2, 4], 0)
+
+    assert_(not route.has_time_warp())
+    assert_equal(route.duration(), 5_229)
+    assert_equal(route.overtime(), 229)
+
+    # Duration cost includes the cost of overtime.
+    assert_equal(route.duration_cost(), 1 * 5_229 + 10 * 229)
```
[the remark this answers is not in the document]
leonlan: Thinking ahead: we might one day introduce soft time windows. For each client, we get some `lateArrivalCost` and `earlyArrivalCost`. These also quality as duration costs.

pyvrp/cpp/search/Route.h
```diff
@@ -448,18 +457,35 @@ class Route
...
      * Returns true if this route has duration-related cost components, either
      * via the objective or via penalised constraints. False otherwise.
      */
     [[nodiscard]] inline bool hasDurationCost() const;
 
     /**
-     * @return The maximum route duration that the vehicle servicing this route
-     *         supports.
+     * @return The (soft) maximum shit duration that the vehicle servicing this
+     *         route supports. This may optionally be extended with overtime.
```
leonlan:
```suggestion
     * @return The (soft) maximum shift duration that the vehicle servicing this
     *         route supports. This may optionally be extended with overtime.
```
💩

pyvrp/cpp/ProblemData.h
```diff
@@ -415,6 +419,12 @@ class ProblemData
      * max_reloads
      *     Maximum number of reloads the vehicle may perform on a route.
      *     Unconstrained if not explicitly provided.
+     * max_overtime
+     *     Maximum allowed overtime, on top of the :py:attr:`~shift_duration`.
+     *     Default 0, that is, overtime is not allowed.
+     * unit_overtime_cost
+     *     Cost of a unit of overtime. This is in addition to the regular
+     *     :py:attr:`~unit_duration_cost` of route durations. Default 0.
```
leonlan, line 427: Not something actionable, but I think that `unit_overtime_cost` can be slightly misleading. If regular rates are €20 and overtime rates are €30, I can imagine that users would put 20 and 30 for `unit_duration_cost` and `unit_overtime_cost`, instead of 20 and 10. 

I think this can be easily addressed by an example. Should I make one after merging this PR?

pyvrp/cpp/search/Route.h
N-Wouda: I see no problem here 😆

pyvrp/cpp/CostEvaluator.h
N-Wouda: Something like this?

image

pyvrp/cpp/ProblemData.h
N-Wouda: Yeah I think we can have an example for this to explain further, but the API docstrings are already quite clear. Just a shame nobody reads those :).

pyvrp/cpp/CostEvaluator.h
leonlan: LGTM

PR #931, opened 2025-11-13
Adding ReachedBKS stopping criterion

0nyr:
This PR:

- [ ] Closes #xxxx.

### PR description

This PR adds a new stopping criterion: `ReachedBKS` with clear tests and documentation. This criterion is a very handy when benchmarking large datasets with typically large time limits (1000s and above) since it allows to early exit a run that has reached the BKS without impacting the results.

NOTE: Tests are passing and the documentation was also built locally without issues.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,6 +111,9 @@ addopts = [
     "--numprocesses=auto",
 ]
 testpaths = "tests"
+markers = [
+    "slow: marks tests as slow (deselect with '-m \"not slow\"')",
+]
 
 
 [tool.coverage.run]


0nyr:
Dear maintainers, I keep getting an error with the CI step: `- uses: codecov/codecov-action@v5`. This CI failure is unrelated to my Python changes — tests passed but coverage XML generation fails due to a gcov/gcovr parse bug.

#### Summary

- Tests: 887 passed (see logs).
- Failure point: coverage step (gcovr parsing gcov output).
- Error excerpt:
    - "SuspiciousHits: pyvrp/cpp/search/Route.h:645 Got suspicious hit value ... This is caused by a bug in gcov tool (GCC bug 68080)."
    - "(ERROR) Exiting because of parse errors. You can run gcovr with --gcov-ignore-parse-errors=... to continue anyway."

#### Cause

- gcov produced suspicious hit counts in C++ header lines (GCC bug 68080). gcovr treats that as a parse error and exits non‑zero, which makes the Meson coverage step fail even though the tests themselves passed.

N-Wouda:
> This CI failure is unrelated to my Python changes — tests passed but coverage XML generation fails due to a gcov/gcovr parse bug.

I would not be surprised if the hit counts themselves are correct, and it's just gcovr's threshold that's the problem. The tests run for so long that the hit counts can get really large.

pyvrp/stop/ReachedBKS.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, bks_value: int):
+        """Initializes the ReachedBKS stopping criterion.
+
+        Parameters
+        ----------
+        bks_value
+            The objective value of the best known solution. Must be a
+            positive integer (PyVRP uses integer costs internally).
+        """
+        if bks_value <= 0:
```
N-Wouda: Why is zero not allowed? For instances that are largely about finding a feasible solution, a zero objective is not that uncommon.

0nyr:
That's a possible issue. I added a custom `@pytest.mark.slow` to that regard, which allows to filter out slow tests, which could be an answer to the issue. What do you suggest ?

tests/conftest.py
N-Wouda: Revert file

N-Wouda:
Hi @0nyr, I had begun a review, but I just realised that in calling code, a BKS value must be available somewhere to use this stopping criterion. Let's call that value `bks_value`. The stopping criterion that's defined in this PR can then be defined much more simply as 
```python
lambda best_cost: best_cost <= bks_value
```
Defining that lambda is no harder than importing a similar device from PyVRP, if we were to support it. What's the benefit of us providing something that can already be done so easily in calling code?

0nyr:
Good point. I would say for the same reasons that you did not define lambdas for other simple criterions `NoImprovement`, `MaxRuntime` etc. It allows for simple combinations of different criteria through you API:

```python
stop_criterion = MultipleCriteria([
    ReachedBKS(bks_value), # provided by the benchmark used
    MaxRuntime(1000)  # Stop after 60 seconds in case BKS is not reached
])
res = solve(rc208, stop=stop_criterion, seed=0)
```

This situation happens a lot for any code that needs to compare runtimes of PyVRP against another solver (which is exactly my use case). And I was surprise this was not available by default. And yes it requires an external source for solution, which is almost always the case when comparing solvers on benchmarks in the literature (only useful in academic research).

N-Wouda:
> I would say for the same reasons that you did not define lambdas for other simple criterions `NoImprovement`, `MaxRuntime` etc. 

Those criteria require a few more statements to update their internal state. The criterion proposed in this PR does not. I suggest you do the following:
> ```diff
> stop_criterion = MultipleCriteria([
> -   ReachedBKS(bks_value), # provided by the benchmark used
> +   lambda best_cost: best_cost <= bks_value,
>     MaxRuntime(1000)  # Stop after 60 seconds in case BKS is not reached
> ])
> res = solve(rc208, stop=stop_criterion, seed=0)
> ```
The lambda implements our `StoppingCriterion` protocol since that does not require much - it just needs to be callable with a `float` argument.

PR #938, opened 2025-12-05
Add replace move for optional clients

### LINKED ISSUE #882: `Replace` operator
Our local search operators currently only act on (U, V) node pairs that are both present in the solution. We're missing an opportunity here to evaluate node pairs where V is absent using node operators that have an 'inplace' mechanism.

A local search node operator named that can work with (U, V), where U is present and V is absent. One possibility that comes to mind is `SwapInPlace<N>`, which swaps `U, U+1, ..., U+N` with `V`.

I quickly prototyped the operator https://github.com/PyVRP/PyVRP/tree/swap-in-place, and the results show that this works pretty well:

Image

There are probably many edge cases that I haven't considered yet (e.g., groups).

N-Wouda:
This PR:

- [x] Closes #882.

See also #884 for a similar node operator.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/tests/search/test_LocalSearch.py
+++ b/tests/search/test_LocalSearch.py
@@ -441,6 +441,45 @@ def test_local_search_does_not_remove_required_clients():
     assert_(new_cost < sol_cost)
 
 
+def test_replacing_optional_client():
+    """
+    Tests that the local search evaluates moves where an optional client is
+    replaced with another that is not currently in the solution.
+    """
+    mat = [
+        [0, 0, 0],
+        [0, 0, 2],
+        [0, 2, 0],
+    ]
+    data = ProblemData(
+        clients=[
+            Client(0, 0, tw_early=0, tw_late=1, prize=1, required=False),
+            Client(0, 0, tw_early=0, tw_late=1, prize=5, required=False),
+        ],
+        depots=[Depot(0, 0)],
+        vehicle_types=[VehicleType()],
+        distance_matrices=[mat],
+        duration_matrices=[mat],
+    )
+
+    rng = RandomNumberGenerator(seed=42)
+    ls = LocalSearch(data, rng, compute_neighbours(data))
+    ls.add_node_operator(Exchange10(data))
+
+    # We start with a solution containing just client 1.
+    sol = Solution(data, [[1]])
+    assert_equal(sol.prizes(), 1)
+    assert_(sol.is_feasible())
+
+    # A unit of time warp has a penalty of 5 units, so it's never worthwhile to
+    # have both clients 1 and 2 in the solution. However, replacing client 1
+    # with 2 yields a prize of 5, rather than 1, at no additional cost.
+    cost_eval = CostEvaluator([], 5, 0)
+    improved = ls(sol, cost_eval)
+    assert_equal(improved.prizes(), 5)
+    assert_(improved.is_feasible())
+
+
 def test_mutually_exclusive_group(gtsp):
     """
     Smoke test that runs the local search on a medium-size TSP instance with


N-Wouda:
This definitely breaks something 😆. Let's figure that out after dinner.

N-Wouda:
OK so this sometimes cycles (a cycle of improving moves). That should not happen since it implies infinite improvement, so I need to figure out what we're evaluating wrong.

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -277,8 +277,36 @@ void LocalSearch::applyOptionalClientMoves(Route::Node *U,
         update(route, route);
     }
 
-    if (!U->route())
+    if (uData.required && !U->route())
+    {
         insert(U, costEvaluator, uData.required);
+        return;
+    }
+
```
leonlan, line 288: It's assumed that U is not in a route after this, but that's not the case. So the next loop will try to insert U's that are already in the solution.

leonlan: In other words, we should shortcut before the loop if `U` is in a route.

N-Wouda: Yeah, I had that locally already but that isn't it. There's something wrong in inplaceCost, I think

pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -277,8 +277,36 @@ void LocalSearch::applyOptionalClientMoves(Route::Node *U,
...
+        if (!route)
+            continue;
+
+        if (insertCost(U, V, data, costEvaluator) < 0)
+        {
+            route->insert(V->idx() + 1, U);
+            update(route, route);
+            return;
+        }
+
```
leonlan, line 304: If it has to do with `inplaceCost` I suspect it might have to do with swapping out required `V`s. We should only allow replacing if `V` is not required. 

```cpp
        ProblemData::Client const &clientData = data.location(vClient);
        if (clientData.required)  // then we cannot remove V.
            continue;
```

This can be a problem when there is a mix of optional and required clients, such as in the `okSmallPrizes` instance. Combined with the other shortcut, this passes all tests on my machine.

(I'm not sure yet _why_ this is needed though)

N-Wouda: I think removing required clients might be the problem, yeah. We probably shouldn't do that here.

leonlan:
I’m running a benchmark now. The performance gain seems to be a little less than the previous node operator version, but I prefer this approach because it’s much simpler. Will post the results in the afternoon after I get back home

pyvrp/cpp/search/LocalSearch.cpp
N-Wouda: That was it, thanks!

N-Wouda:
Uhh sure but this isn't done yet 😆. The insert/inplace evals don't have to be exact, for example. We can probably template those with a `bool exact = true` param, like in the `CostEvaluator`.

N-Wouda:
Allowing inexact evals with the primitives is way more work. I'm putting that in a stash for now.

N-Wouda:
OK I think this is reasonably OK. I need to add a test for swapping but I have to go now, so that'll have to come later.

leonlan:
image

Benchmark. PCVRPTW is solved consistently better with this swap move. I assume that with delta cost shortcutting, this can still shave off a few 0.0x% but I think it's already good as-is.

tests/search/test_LocalSearch.py
```diff
@@ -441,6 +441,45 @@ def test_local_search_does_not_remove_required_clients():
     assert_(new_cost < sol_cost)
 
 
+def test_replacing_optional_client():
```
N-Wouda, line 444: Checked that this fails on current main

PR #939, opened 2025-12-08
Simplify ILS class

N-Wouda:
Part of refactoring

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -62,4 +62,4 @@ jobs:
         run: ls -la dist/
       - name: Deploy to PyPI
         run: |
-          uv publish --token ${{ secrets.PYPI_TOKEN }}
+          uv publish
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![codecov](https://img.shields.io/codecov/c/github/PyVRP/PyVRP?style=flat-square&logo=codecov&label=Codecov)](https://codecov.io/gh/PyVRP/PyVRP)
 [![DOI:10.1287/ijoc.2023.0055](https://img.shields.io/badge/DOI-ijoc.2023.0055-green?style=flat-square&color=blue)](https://doi.org/10.1287/ijoc.2023.0055)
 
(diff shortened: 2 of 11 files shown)


pyvrp/IteratedLocalSearch.py
```diff
@@ -202,38 +119,58 @@ def run(
...
+
+            cand_cost = cost_eval.penalised_cost(candidate)
+            self._history[iters % self._params.history_length] = cand_cost
 
-            if self._accept(candidate, best, stop):
+            if not best.is_feasible():
+                # Best could be infeasible with a low cost at the start, so
+                # first we explore until we get to a feasible reference point.
+                current = candidate
+                continue
```
N-Wouda: @leonlan do you have a small (test) case or instance where this was helpful? I'd like to add a test for that specifically.

leonlan: Not yet. This is from https://github.com/PyVRP/PyVRP/pull/877#discussion_r2228303191. Will look into this a bit more.

pyvrp/IteratedLocalSearch.py
```diff
@@ -202,43 +117,84 @@ def run(
...
+            if (fraction := stop.fraction_remaining()) is not None:
+                weight *= fraction
+
+            best_weight = (1 - weight) * history.min()
+            mean_weight = weight * history.mean()
+            if cand_cost <= best_weight + mean_weight:
                 current = candidate
 
+            stats.collect(current, candidate, best, cost_eval)
+            print_progress.iteration(stats)
```
N-Wouda, line 166: This was subtly broken before - collection needs to happen all the way at the end, after we have done every change to current, cand, and best. The previous location did not track candidate replacing current via the acceptance criterion.

tests/test_IteratedLocalSearch.py
N-Wouda, line 1: There were a few tests here that I didn't really know how to salvage. I think I've replaced most of what was lost, however.

PR #942, opened 2025-12-10
Small stylistic changes to progress output

N-Wouda:
This PR:

- [ ] Closes #xxxx.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/Statistics.py
+++ b/pyvrp/Statistics.py
@@ -81,7 +81,7 @@ class Statistics:
         best
             The best solution.
         cost_evaluator
-            Cost evaluator to use.
+            CostEvaluator used to compute costs for solutions.
         """
         if not self._collect_stats:
             return


pyvrp/ProgressPrinter.py
```diff
@@ -34,6 +32,8 @@
 {summary}
 """
 
+_RESTART = "R                 |                      restart"
+
```
N-Wouda, line 36: Where it currently also is on main (I'm minimising unnecessary diffs)

pyvrp/Statistics.py
```diff
@@ -77,7 +81,7 @@ def collect(
         best
             The best solution.
         cost_evaluator
-            Cost evaluator to use.
+            CostEvaluator used to compute costs for solutions.
```
N-Wouda, line 84: Not pretty, but how it is on main and the new wording isn't much better than the old

pyvrp/Statistics.py
```diff
@@ -32,10 +32,14 @@ class Statistics:
         excessive memory use on long runs.
     """
 
+    runtimes: list[float]
+    num_iterations: int
+    data: list[_Datum]
```
N-Wouda, line 37: This is public state that's only initialised in `__init__`, so revert to this

PR #951, opened 2025-12-14
Perturbation manager

N-Wouda:
This PR:

- [ ] Closes #xxxx.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/search/PerturbationManager.h
+++ b/pyvrp/cpp/search/PerturbationManager.h
@@ -60,4 +60,4 @@ public:
 };
 }  // namespace pyvrp::search
 
-#endif  // PYVRP_SEARCH_LOCALSEARCH_H
+#endif  // PYVRP_SEARCH_PERTURBATIONMANAGER_H


N-Wouda:
Still a bit of a WIP as I figure out what responsibilities this class should have.

N-Wouda:
> Still a bit of a WIP as I figure out what responsibilities this class should have.

More than it does now, but that's still a few refactorings away (#952 and #953).

tests/search/test_LocalSearch.py
```diff
@@ -725,8 +727,8 @@ def test_perturb_inserts_clients(ok_small):
     """
     Tests that perturbing an empty solution inserts all missing clients.
     """
-    ls = cpp_LocalSearch(ok_small, compute_neighbours(ok_small))
-    ls.num_perturbations = 4
+    perturbation = PerturbationManager(PerturbationParams(4, 4))
+    ls = cpp_LocalSearch(ok_small, compute_neighbours(ok_small), perturbation)
```
leonlan: Doesn't really matter but `cpp_LocalSearch` is no longer needed with the new PerturbationManager right?

pyvrp/cpp/search/LocalSearch.h
```diff
@@ -55,7 +56,9 @@ class LocalSearch
     // numLocations, but nothing is stored for the depots!)
     Neighbours neighbours_;
 
-    size_t numPerturbations_ = 0;  // number of perturbations to apply
+    // Perturbation manager that determines the size of the perturbation during
+    // each LS invokation.
```
leonlan:
```suggestion
    // each LS invocation.
```

tests/search/test_LocalSearch.py
N-Wouda: No, this works fine without now. Only `test_perturb_switches_remove_insert` needs it still because that depends on an explicit evaluation order (orderNodes starts as [1, 2, 3, ...]).

PR #958, opened 2025-12-16
Minimise memory profile

### LINKED ISSUE #735: Performance on very large scale instances?
We currently regularly test on instances with up to 1K clients. But how does PyVRP perform on (much) larger instances? I know CVRPLIB also contains the large Flanders instances of Arnold, Gendreau and Sörensen (2017) which range from 3K to 30K clients. Perhaps we can run a benchmark on those sometime to see where the current bottlenecks are.

Relevant links:
- [This Hexaly post](https://www.hexaly.com/benchmark/large-scale-instances-capacitated-vehicle-routing-problem-cvrp)
- [CVRPLIB](http://vrp.galgos.inf.puc-rio.br/index.php/en/)

leonlan:
This PR:

- [x] Closes #735.

<details>

**Notes**:

In particular:

</details>


pyvrp/PenaltyManager.py
```diff
@@ -190,33 +190,41 @@ def init_from(
...
-            np.minimum(edge_costs, mat, out=edge_costs)
+            buf[:] = unit_dist * distances[prof].astype(np.float32)
+            buf += unit_dur * durations[prof].astype(np.float32)
+            np.minimum(edge_costs, buf, out=edge_costs)
 
         # Best edge cost/distance/duration over all vehicle types and profiles,
         # and then average that for the entire matrix to obtain an "average
         # best" edge cost/distance/duration.
         avg_cost = edge_costs.mean()
-        avg_distance = np.minimum.reduce(distances).mean()
```
leonlan, line 207: `np.minimum.reduce` has peak memory of `num_profiles` x `matrix.size`.

```python
import tracemalloc
import numpy as np

tracemalloc.start()

num_profiles = 10
num_locs = 1000
dists = [
    np.zeros((num_locs, num_locs), dtype=np.int64) for _ in range(num_profiles)
]
np.minimum.reduce(dists)

current, peak = tracemalloc.get_traced_memory()
print(f"Current memory usage: {current / 1024**2:.1f} MB")
print(f"Peak memory usage: {peak / 1024**2:.1f} MB")

tracemalloc.stop()
```

Returns
```
Current memory usage: 76.3 MB
Peak memory usage: 160.2 MB
```

Rewriting `reduce` making use of our own out variable can reduce this to `matrix.size`.

pyvrp/search/neighbourhood.py
```diff
@@ -157,30 +157,52 @@ def _compute_proximity(
...
-        mat = unit_dist * distances[prof] + unit_dur * durations[prof]
-        np.minimum(edge_costs, mat, out=edge_costs)
+    edge_costs = unit_dist * distances[prof].astype(np.float32)
+    edge_costs += unit_dur * durations[prof].astype(np.float32)
+    buf = np.empty_like(edge_costs)
 
-    # Minimum wait time and time warp of visiting j directly after i.
-    min_duration = np.minimum.reduce(durations)
-    min_wait = early[None, :] - min_duration - service[:, None] - late[:, None]
-    min_tw = early[:, None] + service[:, None] + min_duration - late[None, :]
```
leonlan, line 177: This is the most costly step (ignoring profiles), resulting in peak memory of 6x matrices of int64.

tests/test_PenaltyManager.py
leonlan, line 1: Checked that both tests (and both assertions) fail on main.

pyvrp/PenaltyManager.py
```diff
@@ -207,16 +206,28 @@ def init_from(
         first, *rest = unique_edge_costs
         unit_dist, unit_dur, prof = first
         edge_costs = unit_dist * distances[prof] + unit_dur * durations[prof]
+        buf = np.empty_like(edge_costs)
+
         for unit_dist, unit_dur, prof in rest:
-            mat = unit_dist * distances[prof] + unit_dur * durations[prof]
-            np.minimum(edge_costs, mat, out=edge_costs)
+            buf[:] = unit_dist * distances[prof] + unit_dur * durations[prof]
```
leonlan, line 201: I'm confused why this doesn't create a peak memory of 4 matrices:
- `edge_costs`
- `buf`
- `tmp1 = unit_dist * distances[prof]`
- `tmp2 = unit_dur * durations[prof]`

Memory profile only shows a peak of 2 matrices (and test also passes).

leonlan: https://numpy.org/doc/stable/reference/c-api/data_memory.html#example-of-memory-tracing-with-np-lib-tracemalloc-domain

N-Wouda:
I had a go at this locally, and the changes to PenaltyManager alone seem to 3x my local memory use (from 1x on current main). So clearly there's a massive difference between platforms/Python versions/numpy versions in how memory is (re)used.

I don't think we have the tools in Python to easily and reliably avoids allocations. We're certainly not doing anything too outrageous - the worst case for me was `compute_neighbours`, which on my machine did three matrix allocations in total. For a very large instance of say 10K locations that's a few additional GB. With a careful implementation we can probably avoid 1, _maybe_ 2 such allocations, but I don't really want to drop to C(++) for just this.

Since it appears to be heavily setup/machine dependent, and a few GB of additional memory use for really large instances (10K+ clients) really isn't that much with modern hardware, I'm leaning wontfix.

PR #961, opened 2025-12-18
Remove unnecessary const ref from Solution constructor

leonlan:
This PR:

- [x] Part of #735.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/bindings.cpp
+++ b/pyvrp/cpp/bindings.cpp
@@ -839,7 +839,8 @@ PYBIND11_MODULE(_pyvrp, m)
         .def(py::init<ProblemData const &, std::vector<Route>>(),
              py::arg("data"),
              py::arg("routes"))
-        .def(py::init<ProblemData const &, std::vector<std::vector<size_t>>>(),
+        .def(py::init<ProblemData const &,
+                      std::vector<std::vector<size_t>> const &>(),
              py::arg("data"),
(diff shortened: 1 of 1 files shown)


pyvrp/cpp/search/LocalSearch.cpp
leonlan, line 1: Antwerp1 instance, just load/export with an operatorless LS:

Without move semantics
```
Results over 5 runs of 2000 iters:
  Mean:   5.172 ms per iter
  Stdev:  0.110 ms per iter
```

With move semantics
```
Results over 5 runs of 2000 iters:
  Mean:   3.899 ms per iter
  Stdev:  0.172 ms per iter
```

pyvrp/cpp/Solution.cpp
```diff
@@ -228,17 +228,22 @@ Solution::Solution(ProblemData const &data,
 }
 
 Solution::Solution(ProblemData const &data, std::vector<Route> const &routes)
-    : routes_(routes), neighbours_(data.numLocations(), std::nullopt)
+    : Solution(data, std::vector<Route>(routes))  // delegate to move ctr
 {
-    if (routes.size() > data.numVehicles())
+}
```
leonlan: This adds one move operation and one function call. Not sure how problematic this is, or whether we should introduce a `validate()` member function and implement the body here.

PR #968, opened 2025-12-20
Skip allocations while loading by not reassigning unowned nodes

N-Wouda:
<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/search/Solution.cpp
```diff
@@ -25,15 +25,13 @@ bool operator==(pyvrp::search::Route const &route, pyvrp::Route const &solRoute)
...
-
-        if (tripIdx != 0 && trip.startDepot() != route[++idx]->client())
+        if (trip.startDepot() != route[idx++]->client())
             return false;  // not the same reload depot
 
         for (auto const visit : trip)
-            if (visit != route[++idx]->client())
+            if (visit != route[idx++]->client())
                 return false;
     }
```
N-Wouda: Think this works but we should probably have a test for this stuff somewhere because it's a bit tricky with the increments and different data layouts.

PR #805, opened 2025-05-16
Add depot service duration

N-Wouda:
This PR:

- [x] Closes #798.
- [ ] Adds and passes relevant tests.
- [ ] Has well-formatted code and documentation.

<details>

**Notes**:

Please read our [contributing guidelines](https://pyvrp.org/dev/contributing.html) first.
In particular:

- You must add tests when making code changes.
  This keeps the code coverage level up, and helps ensure the changes work as intended.
- When fixing a bug, you must add a test that would produce the bug in the master branch, and then show that it is fixed with the new code. 
- New code additions must be well formatted. Changes should pass the pre-commit workflow, which you can set up locally using [pre-commit](https://pre-commit.com/#intro). 
- Docstring additions must render correctly, including escapes and LaTeX.
- Finally, it is essential that all contributions in this PR are license-compatible with PyVRP's MIT license.
  Please check that this PR can be included into PyVRP under the MIT license.

</details>

PR #981, opened 2025-12-29
Add depot service duration

### LINKED ISSUE #798: Reintroduce depot service duration

I want to be able to specify load duration at the depots. This has been supported before, then got removed, then was reintroduced in #750, and then got removed again in #776. While the [FAQ](https://pyvrp.org/setup/faq.html#term-How-do-I-model-vehicle-load-or-service-duration-at-the-depots) details a workaround, the 'natural state' is for this field to exist.

The addition of a `service_duration: int` field to the `Depot` class. This service duration is incurred at the start of each trip.

The workaround listed in the FAQ.

N-Wouda:
This PR:

- [x] Closes #798.

See also #805. Hopefully this attempt fares better.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/tests/test_Route.py
+++ b/tests/test_Route.py
@@ -761,6 +761,18 @@ def test_small_example_from_cattaruzza_paper():
     assert_equal(route.service_duration(), 25 + 80)  # at clients and depots
     assert_equal(route.travel_duration(), 105)
 
+    # Numbers from the paper. Trip 3 starts at 125, 4 at 115, and the route
+    # ends at the end depot at 95.
+    schedule = route.schedule()
+    assert_equal(schedule[5].start_service, 125)  # trip 3
+    assert_equal(schedule[5].service_duration, 20)
+
+    assert_equal(schedule[7].start_service, 115)  # trip 4
+    assert_equal(schedule[7].service_duration, 20)
+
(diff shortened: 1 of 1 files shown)


N-Wouda:
I'm a little stuck here on how to handle this nicely. We want to depot service duration to apply only when *starting* at the depot, not when ending there. But that's rather tricky to do with our current reload depot nodes, because we make no distinction between a start and end depot there.

I'm trying to figure out if there's a relatively easy way to make this work without having to duplicate those nodes. If not, I'll try to split them up and see if that makes things a bit easier.

pyvrp/cpp/search/Route.h
```diff
@@ -823,8 +823,20 @@ Route::SegmentBetween::duration([[maybe_unused]] size_t profile) const
...
+            DurationSegment const durAt = {depot, 0};
+            durSegment
+                = DurationSegment::merge(mat(from, to), durSegment, durAt);
+        }
+        else
+        {
+            auto const &durAt = route_.durAt[step + 1];
+            durSegment
+                = DurationSegment::merge(mat(from, to), durSegment, durAt);
+        }
```
N-Wouda: Woudl rather not branch here, need to think about how to avoid that.

leonlan (approved):

Issue #963, opened 2025-12-18
Optionally skip perturbation in call to `LocalSearch::operator()`


N-Wouda:
We currently have no way to do this except by passing an 'empty' perturbation manager during construction, with `min=max=0` perturbations. Looking at #962 it'd be nice if we could disable it some other way, too.

N-Wouda:
Maybe add `exhaustive: bool = False` to `SearchMethod.__call__`? I don't think we'd need the `SearchMethod` to know about perturbation, per se, but allowing it to optionally implement a (non)exhaustive search seems reasonable.

N-Wouda:
Closed by #982

PR #982, opened 2026-01-04
feat(search): add exhaustive parameter to skip perturbation in LocalSearch

SkyTik:
## Summary
- Add `exhaustive` parameter to LocalSearch that skips perturbation step for deterministic improvement
- Update SearchMethod protocol to include exhaustive parameter

### DIFF
--- a/pyvrp/cpp/search/LocalSearch.h
+++ b/pyvrp/cpp/search/LocalSearch.h
@@ -151,15 +151,6 @@ public:
     /**
      * Iteratively calls ``search()`` and ``intensify()`` until no further
      * improvements are made.
-     *
-     * Parameters
-     * ----------
-     * solution
-     *     The solution to improve.
-     * costEvaluator
-     *     Cost evaluator to use.
-     * exhaustive
-     *     If true, skips the perturbation step. Default false.
      */
     pyvrp::Solution operator()(pyvrp::Solution const &solution,
                                CostEvaluator const &costEvaluator,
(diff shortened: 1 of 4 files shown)


N-Wouda:
Hi @SkyTik, thanks for your PR. I assume it's meant to close #963? I'm not sure yet how to resolve that issue since the `LocalSearch` implements the `SearchMethod` protocol, and other code relies on that protocol interface - not the specific `LocalSearch`. Do you have any ideas on how we can somehow get this behaviour - full search without perturbation - without having to rely on the `LocalSearch` type instead of `SearchMethod`?

SkyTik:
I think I can implement a new method for LocalSearch besides the default __call__, like this:

```
pyvrp::Solution LocalSearch::search_without_perturbation(pyvrp::Solution const &solution, CostEvaluator const &costEvaluator)
```

- __call__ → keeps perturbation (default behavior, satisfies SearchMethod protocol)
- search_without_perturbation() → skips perturbation

N-Wouda:
Perhaps we can add an `exhaustive: bool = False` argument to `SearchMethod.__call__`, and implement that for the LS as well? For the LS that has the same effect as skipping perturbation, but calling it exhaustive search seems somewhat more general - the code using the LS mostly doesn't know about perturbation.

SkyTik:
i think it's a good idea, i will update the implementation

N-Wouda (approved):
LGTM, thanks @SkyTik!

N-Wouda:
The failure is unrelated, I think, and has already been reported upstream: https://github.com/gcovr/gcovr/issues/1228.

PR #988, opened 2026-01-09
Full search on new best

### LINKED ISSUE #984: Do full search for final solution (or on new best?)
Kind of ties into #963. It's a good idea to do a complete local search on the final solution that we return as the best observed, to make sure there are no more easy, 'obvious' local improvements. We could implement this just before returning, or do it everytime we find a new best solution.

N-Wouda:
This PR:

- [x] Closes #984.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ pyvrp = "pyvrp.cli:main"
 build = [
     "meson>=1.0.0",
     "ninja>=1.11.1",
-    "gcovr>=7.2",
+    "gcovr>=7.2,<8.5",
     "docblock>=0.1.6",
 ]
 


pyvrp/IteratedLocalSearch.py
```diff
@@ -28,10 +28,14 @@ class IteratedLocalSearchParams:
...
         History length for the late acceptance hill-climbing stopping criterion
         used by the algorithm. Must be positive.
+    exhaustive_on_best
+        Whether to perform a more expensive, exhaustive search for newly found
+        best solutions.
     """
 
     num_iters_no_improvement: int = 50_000
     history_length: int = 500
+    exhaustive_on_best: bool = True
```
N-Wouda, line 38: Seems better (even on short runs) for instances up to 1K. A little worse for the XXL CVRP instances, since there full evaluations are really expensive, but that's only for runs <1min or so. On 10min runs there's nothing noticeable anymore in performance. Added a toggle just in case.

I want this at least for the final, returned solution. We shoudl always return at least a locally optimal solution. But doing it for every improving sol seems OK too (except for early on with the really large instances, but that's pretty niche anyway).

pyvrp/IteratedLocalSearch.py
```diff
@@ -136,20 +140,25 @@ def run(
...
+            self._pm.register(cand)
 
             iters_no_improvement += 1
-            if cost_eval.cost(candidate) < cost_eval.cost(best):  # new best
-                best = candidate
+            if cost_eval.cost(cand) < cost_eval.cost(best):
+                if self._params.exhaustive_on_best:
+                    # Candidate is already a new (global) best, but let's see
+                    # if we can improve it further via an exhaustive search.
+                    cand = self._search(cand, cost_eval, exhaustive=True)
```
leonlan, line 160: This could return an infeasible solution, so we have to check `if cost_eval.cost(cand) < cost_eval.cost(best):` again before assigning the fully searched candidate solution as best.

N-Wouda: Thanks, good catch. I realised this as I went to bed yesterday but didn't want to get up to change it anymore 😂.

leonlan:
| Commit  | Time     | CVRP-XXL | MDVRPTW | HFVRP | VRPB | MTVRPTWR | CVRP | PCVRPTW | VRPTW |
|---------|----------|----------|---------|-------|------|----------|------|---------|-------|
| 8c65c66 (main)| 1min     |    5.80% |   3.57% | 2.45% | 1.61% |    1.25% | 1.06% |   1.08% | 2.89% |
| 8c65c66 | 2min     |    5.11% |   2.07% | 1.67% | 1.13% |    0.70% | 0.70% |   0.70% | 2.14% |
| 8c65c66 | 10min    |    3.70% |   0.85% | 0.74% | 0.62% |    0.32% | 0.35% |   0.45% | 1.03% |
| 5154249 | 1min     |    7.23% |   3.55% | 2.42% | 1.63% |    1.18% | 1.04% |   1.07% | 2.84% |
| 5154249 | 2min     |    5.65% |   2.09% | 1.67% | 1.16% |    0.69% | 0.69% |   0.68% | 2.11% |
| 5154249 | 10min    |    3.68% |   0.91% | 0.73% | 0.63% |    0.32% | 0.35% |   0.43% | 1.05% |

Ran a quick bench to check, the performance differences are very minimal. FYI, I noticed about 200-500 best solutions during a regular search trajectory for with ~1K clients.

leonlan (approved):

Issue #1018, opened 2026-01-28
Separate UnaryOperator and BinaryOperator in LS


N-Wouda:
Basically, current `NodeOperator` aka `LocalSearchOperator<Route::Node>` should be `LocalSearchOperator<Route::Node, Route::Node>`. It can be renamed to `BinaryOperator`, I think, since it acts on pairs of (U, V) nodes.

We should add a `UnaryOperator` (`= LocalSearchOperator<Route::Node>`), too, acting on `U` alone. That lets us refactor a lot of moves out of the LS.

N-Wouda:
> We should add a UnaryOperator (= LocalSearchOperator<Route::Node>), too, acting on U alone. That lets us refactor a lot of moves out of the LS.

Everything is now in place for this, after #1019.

N-Wouda:
The unary operators replace the `applyMove` methods of the LS. Those currently mark lots of things as promising, but I'm not sure if they should. We only evaluate them when `U`'s already promising, and I think that's sufficient.

N-Wouda:
These can all become operators of some form, I think:

https://github.com/PyVRP/PyVRP/blob/d44efb2d46fa8a66e2468092f315e38cd16da150/pyvrp/cpp/search/LocalSearch.h#L50-L63

They're currently very deeply integrated into the LS, so that'll take a bit of rework to figure out a common pattern. I think some of these moves are actually a composite move, evaluating multiple different things at once.

N-Wouda:
In order:
- [x] `applyDepotRemovalMove` removes `U` if it's a reload depot that's not adding value. Can probably be updated to look at U's neighbours, check if they're depots, and test removing. Unary operator.
- [x] `applyEmptyRouteMoves` kind of special, should probably remain in the local search. Decision: keep in LS.
- [x] `applyOptionalClientMoves` does three things:
    - [x] tests removing U;
    - [x] tests inserting U after some neighbour V;
    - [x] tests replacing some other neighbour V of U with U;
- [x] `applyGroupMoves` works for mutually exclusive groups. It does the following: remove all but one of the group's members, in order of cost; then check if replacing the last member with U is better. This is an odd unary operator, but I'm not immediately sure how to get it to fit our operator model.

Let's start with `applyDepotRemovalMove` and `applyOptionalClientMoves`. We can go from there.

N-Wouda:
[...]

PR #1019, opened 2026-01-28
Rework NodeOperator to BinaryOperator, revisit insert-if-required

N-Wouda:
This PR:

- [x] Part of #1018.

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -328,33 +345,17 @@ void LocalSearch::applyGroupMoves(Route::Node *U,
...
-            {
-                searchSpace_.markPromising(client);
-                continue;
-            }
-        }
+        searchSpace_.markPromising(V);
+        auto *route = V->route();
+        route->remove(V->idx());
+        update(route, route);
+        return;
```
N-Wouda, line 358: As it turns out we've never handled optional groups all that well.

pyvrp/search/LocalSearch.py
```diff
@@ -40,17 +40,17 @@ def __init__(
         self._ls = _LocalSearch(data, neighbours, perturbation_manager)
         self._rng = rng
 
-    def add_node_operator(self, op: NodeOperator):
+    def add_operator(self, op: BinaryOperator):
```
leonlan, line 43: For my understanding, is the plan to make this

`def add_operator(self, op: UnaryOperator | BinaryOperator):`

once we have `UnaryOperator`?

leonlan: More generally, to what extent do we make the distinction between unary/binary operators, like we did with node/route operators?

leonlan (approved):

pyvrp/search/LocalSearch.py
N-Wouda: Yes, that was my thinking. We can just overload this in C++ and add a union type on the Python side. I think we can have just one list of operators, and not really go into detail too much that we have specific unary and binary operators. It's just a long list of them.

PR #1023, opened 2026-01-31
Add UnaryOperator, including RemoveOptional and RemoveAdjacentDepot operators

N-Wouda:
This PR:

- [x] Part of #1018.

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/search/RemoveAdjacentDepot.h
N-Wouda, line 1: Need to think of a way to remove multiple, consecutive reload depots. Those are typically free/0 cost but it's weird to have a route like `1 2 | | | 3` - should just be `1 2 | 3`.

N-Wouda: I think `evaluate() -> int` is not sufficiently granular. The decision of whether to apply the move can probably also better be made by the operator, so maybe an interface like `evaluate() -> tuple[int, bool]`?

We then only apply if the returned boolean argument is `true`, and require the delta cost evaluation to be exact in that case.

N-Wouda: Handled in #1024.

pyvrp/cpp/search/RemoveAdjacentDepot.cpp
```diff
@@ -0,0 +1,68 @@
...
+        Cost deltaCost = removeCost(n(U), data, costEvaluator);
+        if (deltaCost < bestCost)
+        {
+            bestCost = deltaCost;
+            move_ = MoveType::REMOVE_NEXT;
+        }
+    }
+
+    // Apply this move it's either better or neutral. It can be neutral if e.g.
+    // the same depot visited consecutively, but that's unnecessary.
```
leonlan:
```suggestion
    // the same depot is visited consecutively, but that's unnecessary.
```

pyvrp/cpp/search/RemoveAdjacentDepot.cpp
```diff
@@ -0,0 +1,68 @@
...
+    return std::make_pair(bestCost, bestCost <= 0);
+}
+
+void RemoveAdjacentDepot::apply(Route::Node *U) const
```
leonlan, line 47: Calling `apply()` before any improving moves is undefined behaviour. Is that something we need to protect against or make a comment about?

leonlan: It's already covered here:

```
    /**
     * Applies this operator to the given arguments. For improvements, should
     * only be called if <code>evaluate()</code> returns a negative delta cost.
     */
    // TODO remove arguments - always applies to most recently evaluated pair.
    virtual void apply(Args... args) const = 0;
```

But we can rephrase this I think

```
    /**
     * Applies this operator to the given arguments. Should only be called if
     * <code>evaluate()</code> returns when `should_apply=true`.
     */
    // TODO remove arguments - always applies to most recently evaluated pair.
    virtual void apply(Args... args) const = 0;
```

leonlan (approved):

PR #1026, opened 2026-02-02
Insert and replace optional binary operators

N-Wouda:
This PR:

- [x] Part of #1018.

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/search/LocalSearch.cpp
```diff
@@ -153,27 +153,30 @@ bool LocalSearch::applyBinaryOps(Route::Node *U,
...
-                  + Cost(rU != rV) * costEvaluator.penalisedCost(*rV);
+            [[maybe_unused]] Cost costAfter = 0;
+            costAfter += rU ? costEvaluator.penalisedCost(*rU) : 0;
+            costAfter += rV && rU != rV ? costEvaluator.penalisedCost(*rV) : 0;
 
             // When there is an improving move, the delta cost evaluation must
             // be exact. The resulting cost is then the sum of the cost before
             // the move, plus the delta cost.
-            assert(costAfter == costBefore + deltaCost);
+            // assert(costAfter == costBefore + deltaCost);
```
N-Wouda: Routes do not know about (un)collected prizes. They also have no way of knowing about a solution-level property like that. Wasn't a problem before because we didn't assign clients via binary operators, but we do now. And then this doesn't work.

Don't know how to fix this yet.

N-Wouda: I suppose that for routes we can look at only collected prizes, rather than total uncollected? The route's penalised cost can then be negative, but I see no other way to get this check to work.

Something for tomorrow.

N-Wouda:
Need to look at the LS again to clean it up further. And then write tests. But otherwise this should be OK now.

N-Wouda:
Still need to do a small benchmark on PCVRPTW and investigate why RTD sometimes fails. Bit unclear.

N-Wouda:
It fails because something's not right yet. My benchmark didn't complete.

pyvrp/cpp/search/Route.h
```diff
@@ -499,6 +500,11 @@ class Route
      */
     [[nodiscard]] inline Duration timeWarp() const;
 
+    /**
+     * @reutrn Total collected prizes on this route.
```
leonlan:
```suggestion
     * @return Total collected prizes on this route.
```

pyvrp/cpp/search/ReplaceOptional.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+using pyvrp::search::ReplaceOptional;
+
+std::pair<pyvrp::Cost, bool> ReplaceOptional::evaluate(
+    Route::Node *U, Route::Node *V, CostEvaluator const &costEvaluator)
+{
+    stats_.numEvaluations++;
+
+    if (U->route() || !V->route())
+        return std::make_pair(0, false);
```
leonlan:
```suggestion
    if (U->route() || !V->route() || !V->isDepot()) 
        return std::make_pair(0, false);
```

`V` can be a depot now, so this operator will do weird stuff when evaluated in `applyBinaryOps(U, p(V), ...)` or `applyEmptyRouteMove`. I think shortcutting here fixes the issue.

leonlan: This test fails non-deterministically now, but passes with the proposed fix.

```python
def test_skips_depot(ok_small_prizes):
    """
    Tests that InsertOptional skips depots.
    """
    route = make_search_route(ok_small_prizes, [1])

    op = ReplaceOptional(ok_small_prizes)
    cost_eval = CostEvaluator([0], 0, 0)

    # Should not be allowed to replace a depot.
    depot = route[0]
    client = Node(loc=3)
    assert_equal(op.evaluate(client, depot, cost_eval), (0, False))
```

N-Wouda: Pretty good chance that's it. I'm running it now.

N-Wouda:
Yeah now works. 1min benchmarks on PCVRPTW (1 seed) on my machine:

Current main (2529a58ac5dcfb8cccbf9d73ca1b3ea7ccc61f78):
```
     Avg. objective: 215217
    Avg. iterations: 69692
      Avg. run-time: 60.00s
       Total not OK: 0
```

This PR (182d757564eb7fd0686eb2dd92704f923ee18f72):
```
     Avg. objective: 215230
    Avg. iterations: 71796
      Avg. run-time: 60.00s
       Total not OK: 0
```
Looks fine to me.

leonlan (approved):

PR #1041, opened 2026-02-12
Pickle DynamicBitset

leonlan:
This PR:

- [ ] Closes #xxxx.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/bindings.cpp
+++ b/pyvrp/cpp/bindings.cpp
@@ -17,7 +17,6 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
-#include <climits>
 #include <memory>
 #include <sstream>
 #include <string>
--- a/pyvrp/cpp/bindings.cpp
+++ b/pyvrp/cpp/bindings.cpp
@@ -77,10 +77,9 @@ PYBIND11_MODULE(_pyvrp, m)
                     blocks.push_back(block.to_ullong());
                 return blocks;
             },
-            [](std::vector<unsigned long long> const &blocks)
-            {
-                return DynamicBitset(std::vector<DynamicBitset::Block>(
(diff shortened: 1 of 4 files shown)


tests/test_DynamicBitset.py
```diff
@@ -221,3 +223,17 @@ def test_reset():
...
     bitset.reset()
     assert_equal(bitset.count(), 0)
+
+
+def test_pickle():
+    """
+    Tests that DynamicBitset can be pickled and unpickled correctly.
+    """
+
+    bitset = DynamicBitset(128)
```
N-Wouda, line 232:
```suggestion
    """
    bitset = DynamicBitset(128)
```

pyvrp/cpp/bindings.cpp
```diff
@@ -69,7 +69,20 @@ PYBIND11_MODULE(_pyvrp, m)
...
+            [](DynamicBitset const &bitset) {  // __getstate__
+                std::vector<unsigned long long> blocks;
+                for (auto const &block : bitset.data())
+                    blocks.push_back(block.to_ullong());
+                return blocks;
+            },
+            [](std::vector<unsigned long long> const &blocks) {  // __setstate__
+                std::vector<DynamicBitset::Block> data;
+                for (auto val : blocks)
+                    data.emplace_back(val);
```
N-Wouda:
```suggestion
                std::vector<DynamicBitset::Block> data(blocks.begin(), blocks.end());
```
I think this'll work because a bitset can be constructed from an ull.

N-Wouda: And then you can also just
```
return {blocks.begin(), blocks.end()};
```
I think, because that can be upgraded to a DynamicBitset via its data constructor.

N-Wouda (approved):

Issue #1044, opened 2026-02-15
Piecewise linear duration costs


FormelessPuppy41:
Hi PyVRP contributors,

I would like to request an extension to the current overtime cost modelling to support configurable overtime penalty methods, rather than only a constant linear `unit_overtime_cost`.

The goal is to enable balanced workloads by penalising larger overtime deviations more heavily (e.g., 45 minutes of overtime should be “worse” than 15 minutes, even if both are beyond the same `shift_duration` threshold), while keeping the current behaviour as the default for backwards compatibility.

## Motivation

Currently, overtime cost is linear in overtime: `overtime_cost = unit_overtime_cost * overtime`.

This makes it hard to express workload balancing where larger deviations should be penalised more heavily. For example:

* Route A has 15 minutes overtime and Route B has 45 minutes overtime.
* With a linear coefficient, both are penalised with the same per-unit scaling, so the search has relatively weak incentive to redistribute workload.
* With a convex / non-linear penalty, PyVRP is encouraged to move clients from the high-overtime route to the low-overtime route (e.g., aiming for about 30 minutes and 30 minutes), because reducing large overtime yields a larger cost improvement.

## Proposal

Introduce an overtime penalty configuration whose default reproduces the current behaviour (full backwards compatibility), but which can be swapped for other methods.

_High-level change:_
* Current: `overtime_cost = unit_overtime_cost * overtime`.
* Proposed: `overtime_cost = unit_overtime_cost * method(overtime)`,
  where `method()` is configurable.

To avoid performance issues from Python callbacks during move evaluation, these methods should be built-in and evaluated in C++, with parameters passed via the Python API.

_Possible API design_

One option is to extend `VehicleType` with either:

1. A simple selector, for example: `overtime_penalty_method = "linear"` (default) / `"quadratic"` / `"piece_wise_linear"` / `"polynomial"` / ...
2. A configuration object, for example: `overtime_penalty = { type: ... (e.g. "polynomial"), params: ... (e.g. a=(1,2), b=(3,1), c=(1,0) for x^2 + 3x + 1) }`

[...]

PR #1046, opened 2026-02-18
Piecewise linear function class

### LINKED ISSUE #925: Piecewise linear functions
For #867 and #642 we'll need piecewise linear functions. @wouterkool reminded me a few days ago that such functions can also be a nice way to let the user specify duration and/or distance costs: they generalise the `unit_distance_cost`, `unit_duration_cost`, and `unit_overtime_cost` parameters we have now.

This issue is blocking for when (if?) we tackle #867 or #642.

FormelessPuppy41:
This PR:

- [x] Closes #925.

## Summary

This PR introduces piecewise-linear duration cost modelling and wires it through the C++ core, Python bindings, and search heuristics.

Main goals covered:
- Add a reusable piecewise-linear function implementation.
- Add a dedicated `DurationCostFunction` built on that PWL core.
- Keep legacy `unit_duration_cost` / `unit_overtime_cost` behavior via conversion to `DurationCostFunction`.
- Enforce clear API behavior when custom duration cost functions are provided.
- Update route/search logic and tests accordingly.

### 1) New core C++ modules
- Added `pyvrp/cpp/PiecewiseLinearFunction.h/.cpp`
- Added `pyvrp/cpp/DurationCostFunction.h/.cpp`

These implement:
- validated piecewise-linear representation,
- cumulative duration-cost evaluation semantics,
- `DurationCostFunction::fromLinear(...)` compatibility with legacy duration/overtime scalar costs.

### 2) VehicleType / ProblemData integration
- Updated `pyvrp/cpp/ProblemData.h/.cpp`:
  - `VehicleType` now stores `durationCostFunction`,
  - duration-cost resolution is centralized (`resolveDurationCostFunction(...)`),
  - strict input behavior: either custom duration cost function OR legacy scalar duration/overtime costs (not both),
  - replace/constructor pathways updated consistently.

### 3) Route and search behavior
- Updated `pyvrp/cpp/search/Route.h` (and related files):
  - `hasDurationCost()` now uses the duration-cost function/constraints as source of truth.
- Updated `pyvrp/cpp/Route.cpp`, `pyvrp/cpp/search/Route.cpp` as needed for consistent duration-cost handling.
- Updated search-side usage in:
  - `pyvrp/PenaltyManager.py`
  - `pyvrp/search/neighbourhood.py`
  to use a duration-cost slope accessor instead of relying on legacy `unit_duration_cost`.

### 4) Python bindings and typing surface
- Updated `pyvrp/cpp/bindings.cpp` for new types, serialization paths, and replacement behavior.
- Updated `pyvrp/_pyvrp.pyi` to expose:
  - `PiecewiseLinearFunction`,
  - `DurationCostFunction`,
  - `VehicleType.duration_cost_function`,
  - `VehicleType.duration_cost_slope`.
- Updated `pyvrp/Model.py`, `pyvrp/read.py`, `pyvrp/search/neighbourhood.py` for type consistency and pylance/static typing issues encountered during development.

### 5) Tests and docs
- Added `tests/test_DurationCostFunction.py`.
- Updated `tests/search/test_Route.py` (including `has_duration_cost()` expectations and regression comments).
- Updated docs:
  - `docs/source/api/pyvrp.rst`
  - `docs/source/setup/concepts.rst`
- Minor build setup updates:
  - `meson.build`

## Validation

I ran:
- targeted tests for changed modules, and
- full test suite (`uv run pytest -q -n0`),

and all tests passed locally. 

Note: the test output also shows repeated warnings like:

`profiling: ... .gcda: cannot merge previous GCDA file: corrupt arc tag (...)`

I’m not fully familiar with these warnings, so I’d appreciate confirmation on whether I need to modify the code or whether they can be 'ignored'.

## Reviewer notes / possible cleanup before merge

Some edits were mainly to make local development/type-checking run cleanly and may be optional for final merge:
- `pyproject.toml` (`numpy>=2.4.2` addition),
- typing-focused updates in `pyvrp/Model.py`, `pyvrp/_pyvrp.pyi`, `pyvrp/read.py`, `pyvrp/search/neighbourhood.py`.
- local `compile_commands.json` adjustments were only for local tooling and are not intended as functional PR content.

Note: 
I also added a 'FIXME' to docs/source/setup/concepts.rst to indicate that it might be of interest to include a figure for the DurationCostFunction mechanics. This FIXME must be removed before 'going online' or it will be shown on the website.

<details>

**Notes**:

In particular:

</details>


N-Wouda:
Hi @FormelessPuppy41, thanks for this - could you make the CI pass before I have a first look?

FormelessPuppy41:
Hi @N-Wouda , I’ve resolved the CI issues. There were two problems I wasn’t sure how to fix, so I consulted ChatGPT for guidance. I’ve included its suggested changes in this comment so you can quickly review what I updated and confirm whether the fixes are correct.





## CHAT RESPONSE:
You are not doing anything wrong. You have **two different CI issues**:

1. `Error: The operation was canceled` (clang-19 job)  
This usually means the job was canceled by GitHub (new push, re-run, or concurrency cancel), not a code/build error.

2. `apt-get ... 404 Not Found ... exit code 100` (gcc-10 job)  
This is an Ubuntu package mirror/index issue in CI, not a PyVRP logic issue.

What to do now:

1. Re-run failed jobs in GitHub Actions (or push an empty commit to retrigger).
2. If gcc-10 still fails with the same 404, add `apt-get update` before install in CI workflow.

In `.github/workflows/CI.yml`, update both compiler setup steps to:

```yaml
- if: matrix.compiler == 'clang'
  name: Set up clang
  run: |
    sudo apt-get update -o Acquire::Retries=3
    sudo apt-get install -y clang-${{ matrix.compiler-version }}
    echo "CC=clang-${{ matrix.compiler-version }}" >> $GITHUB_ENV
    echo "CXX=clang++-${{ matrix.compiler-version }}" >> $GITHUB_ENV

- if: matrix.compiler == 'gcc'
  name: Set up GCC
  run: |
    sudo apt-get update -o Acquire::Retries=3
    sudo apt-get install -y gcc-${{ matrix.compiler-version }} g++-${{ matrix.compiler-version }}
    ...
```

.github/workflows/CI.yml
```diff
@@ -60,13 +60,15 @@ jobs:
       - if: matrix.compiler == 'clang'
         name: Set up clang
         run: |
+          sudo apt-get update -o Acquire::Retries=3
```
N-Wouda:
```suggestion
          sudo apt-get update
```
I think this already suffices to obtain the right mirror location for gcc/clang.

.github/workflows/CI.yml
```diff
@@ -60,13 +60,15 @@ jobs:
...
           sudo apt-get install -y clang-${{ matrix.compiler-version }}
 
           echo "CC=clang-${{ matrix.compiler-version }}" >> $GITHUB_ENV
           echo "CXX=clang++-${{ matrix.compiler-version }}" >> $GITHUB_ENV
       - if: matrix.compiler == 'gcc'
         name: Set up GCC
         run: |
+          sudo apt-get update -o Acquire::Retries=3
```
N-Wouda:
```suggestion
          sudo apt-get update
```

pyvrp/cpp/search/Route.cpp
```diff
@@ -344,9 +344,11 @@ void Route::update()
     duration_ = durAfter[0].duration();
     timeWarp_ = durAfter[0].timeWarp(maxDuration());
 
-    auto const overtime = std::max<Duration>(duration_ - shiftDuration(), 0);
-    durationCost_ = unitDurationCost() * static_cast<Cost>(duration_)
-                    + unitOvertimeCost() * static_cast<Cost>(overtime);
+    // OLD CODE
+    // auto const overtime = std::max<Duration>(duration_ - shiftDuration(), 0);
+    // durationCost_ = unitDurationCost() * static_cast<Cost>(duration_)
+    //                + unitOvertimeCost() * static_cast<Cost>(overtime);
```
N-Wouda: Just remove old code; git tracks it in its history. There is no value for us in keeping it as e.g. a comment.

pyvrp/_pyvrp.pyi
```diff
@@ -130,6 +192,16 @@ class VehicleType:
...
     unit_overtime_cost: int
+    duration_cost_function: DurationCostFunction
+    # `duration_cost_slope` is derived from `duration_cost_function` and used
+    # for edge-based heuristics.
+    # It returns the slope of the first segment of the piecewise linear
+    # duration cost function. This is the per-duration-unit cost for small
+    # durations (e.g. within shift duration).
+    # It is used in `compute_proximity_matrix()` in
+    # `pyvrp.search.neighbourhood` and in `init_from()` in
+    # `pyvrp.PenaltyManager` instead of the previous `unit_duration_cost`.
```
N-Wouda: Don't add comments to type stubs. They're not read; this isn't code people look at - it's only for type hints in IDEs.

If you need to comment on the implementation, do so in the docstring (in this case, the docstring of `ProblemData::VehicleType`).

pyvrp/read.py
N-Wouda, line 1: Here, and in many other locations: don't touch places you don't need to be. These typing changes detract from the core of the PR. Revert basically everything that is not part of this core PR.

compile_commands.json
N-Wouda, line 1: Remove

pyproject.toml
```diff
@@ -37,6 +37,7 @@ dependencies = [
     "matplotlib>=2.2.0",
     "vrplib>=1.4.0",
     "tqdm>=4.64.1",
+    "numpy>=2.4.2",
```
N-Wouda: Remove

N-Wouda (commented):
I'm leaving these comments more generally; in its current form this is difficult to review.

I think we can start with just the PWL function, as a data structure we can later use. It doesn't need to be used yet: it just needs to compile, have tests, and support most of the intended use case. So, my suggestion is to keep only the parts needed to satisfy #925 in this PR. The follow-up needed for #1044 can be done afterwards.

That should reduce the size of this PR to only a few files, which is much more focused and far easier to review.

pyvrp/_pyvrp.pyi
FormelessPuppy41: Alright, thanks for the tip. I will remove the comments in the file.

FormelessPuppy41:
Hi @N-Wouda, I pushed new commits and the tests are running. I also reverted the changes that felt out of scope for the initial PWL PR.

No need to review yet; I mainly wanted to confirm CI/tests behave as expected. I’ll keep iterating and message you when I think it’s ready for review.

FormelessPuppy41:
Hi @N-Wouda, could you review the changes?

`PiecewiseLinearFunction` is the base layer, and `DurationCostFunction` specifies it with the domain-specific constraints needed for PyVRP. Tests for both classes are included in `tests/test_DurationCostFunction.py`.

All current changes pass the test suite. I added a few `GCOVR_EXCL_*` annotations for branches that are unreachable through the Python interface, and I reverted the changes that seemed out of scope for the initial PWL PR as you suggested.

I look forward to your feedback.

N-Wouda:
(I've just updated your branch; make sure to `git pull` before committing new changes!)

pyvrp/cpp/DurationCostFunction.h
N-Wouda, line 1: Do we need this? It is basically just a convenience constructor for turning the current unit duration/overtime arguments on VehicleType into a PWL function. I don't mind a breaking chang that removes those, so I don't think we need to have anything special in place for that use case.

Ideally, `VehicleType.duration_cost` becomes just a `PiecewiseLinearFunction`, not a `DurationCostFunction`.

pyvrp/cpp/ProblemData.cpp
```diff
@@ -272,7 +272,7 @@ ProblemData::Depot::~Depot() { delete[] name; }
 bool ProblemData::Depot::operator==(Depot const &other) const
 {
     // clang-format off
-    return x == other.x
```
N-Wouda, line 275: Thanks!

pyvrp/cpp/ProblemData.h
```diff
@@ -1,3 +1,4 @@
+
```
N-Wouda: Revert this change

pyvrp/cpp/Route.cpp
```diff
@@ -283,6 +283,7 @@ Route::Route(ProblemData const &data, Trips trips, size_t vehType)
 
     duration_ = ds.duration();
     overtime_ = std::max<Duration>(duration_ - vehData.shiftDuration, 0);
+
```
N-Wouda: Revert this change

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,131 @@
...
+ * Parameters
+ * ----------
+ * breakpoints
+ *     Strictly increasing list of segment start points.
+ * slopes
+ *     Segment slopes. Must have the same length as ``breakpoints``.
+ * intercept
+ *     Function value at ``breakpoints[0]``.
+ */
+class PiecewiseLinearFunction  // GCOVR_EXCL_LINE
```
N-Wouda:
```suggestion
class PiecewiseLinearFunction
```
Remove all `// GCOVR_EXCL_LINE` statements. I do not mind a few lines missing coverage, but I don't want that to be hidden in the reports.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,131 @@
...
+{
+public:
+    using Scalar = int64_t;
```
N-Wouda: This should probably be templated, as e.g.
```cpp
template <typename Domain>
class PiecewiseLinearFunction
```
We possibly want to separate `Domain` and `Codomain` types. For example, a `duration_cost` function has `Domain = pyvrp::Duration`, and `Codomain = pyvrp::Cost`.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,131 @@
...
+ * Creates a piecewise linear function over integer inputs.
+ *
+ * The given ``breakpoints`` define the start of each segment. Segment ``i``
+ * starts at ``breakpoints[i]`` and uses slope ``slopes[i]``. Segments are
+ * left-closed and right-open, except for the final segment, which is
+ * unbounded to the right.
+ *
+ * Let ``b_i`` denote breakpoints, ``s_i`` segment slopes, and ``v_i`` the
+ * function value at ``b_i``. This class stores and evaluates the cumulative
+ * function value:
```
N-Wouda: This isn't a general piecewise linear implementation. Those need slopes and intercepts for each segment, and don't evaluate cumulatively.

I'd suggest something like
```
PiecewiseLinearFunction(
  breakpoints: list[int] = [INT_MIN, INT_MAX],
  segments: list[tuple[int, int]] = [(0, 0)],
)
```
Then, each segment `idx` is associated with `[breakpoints[idx], breakpoints[idx + 1])` or some such.

Evaluating `f = PiecewiseLinearFunction(...)` is something like this, for `f(x)`:
```python
if x < breakpoints[0] or x > breakpoints[-1]:
    # unsure; perhaps we want to require f to be evaluateable 
    # across its entire domain instead
    raise

for idx, segment in enumerate(segments):
    if breakpoints[idx] <= x < breakpoints[idx + 1]:
        return segment[0] + segment[1] * x
```

Since the breakpoints are ordered, we could replace the loop with a binary search, but I doubt that'll add a lot of value since I expect that the number of segments will be small. I'd stick to a simple loop for now.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Function value at the first breakpoint.
+     */
+    [[nodiscard]] Scalar intercept() const;
+
+    /**
+     * Tests whether this function is the zero function.
+     */
+    [[nodiscard]] bool isZero() const;
```
N-Wouda: Why do we need this?

pyvrp/cpp/PiecewiseLinearFunction.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+using pyvrp::PiecewiseLinearFunction;
+
+namespace
+{
+// NOTE: #925/1044-FormPup41:
+// These overflow helpers are currently local to this file so the checked
+// arithmetic logic is explicit and compiler-independent. If more modules need
+// the same checked int64 arithmetic, this can be moved to a shared internal
+// utility.
```
N-Wouda: None of the modules need this. Just remove these checks, especially from the hot path in `operator()`. The PWL function shouldn't need to know too much about the underlying data types.

N-Wouda (commented):
I think this is a good start, especially since you didn't know any C++ before. Well done!

I have some comments on the design of the `PiecewiseLinearFunction` class. It should be more general than what it is now. I hope that my comments below can help you rewrite it towards a more general implementation.

I am unsure about how to handle bounds checking in `operator()`:
- We could raise when the requested value is outside the function domain. That is what [`std::domain_error`](https://en.cppreference.com/w/cpp/error/domain_error.html) is for.
- We could demand that a PWL function is defined across the entire allowable domain for the `Domain` type. Then the problem vanishes (and this is faster at runtime since there's never going to be an exception branch). But is that reasonable for all intended use cases? I don't know yet.

pyvrp/cpp/DurationCostFunction.h
FormelessPuppy41: Fair point. I was considering `DurationCostFunction` mainly as a future extension point in case we ever want a non-PWL duration cost. That said, it’s probably overkill right now since we can approximate most alternatives with PWL. I’m fine with making `VehicleType.duration_cost` just a `PiecewiseLinearFunction`.

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: `isZero()` is used in the integration where we normalize everything to a `DurationCostFunction`. If no custom duration-cost is provided, we treat it as the zero function and then build one from the legacy scalars (`unit_duration_cost`, `unit_overtime_cost`, `shift_duration`).

So `isZero()` is just the explicit predicate for “was a custom function provided?”, instead of manually checking intercept/slopes each time. In this PR it’s mostly plumbing (`DurationCostFunction::isZero()` delegates to it); the main use would land in the VehicleType follow-up.

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: Okay, I will remove the calls to // GCOVR_... from the `DurationCostFunction.cpp/.h` and `PiecewiseLinearFunction.cpp/.h` files.

pyvrp/cpp/PiecewiseLinearFunction.h
N-Wouda: Then this can probably be removed here since those "legacy scalars" can also just be removed. I don't mind these changes to be breaking.

pyvrp/cpp/DurationCostFunction.h
N-Wouda: It *might* happen that we allow some higher-order piecewise polynomial functions (e.g., quadratic, cubed, ...). But piecewise linear already offers a great deal of flexibility, especially compared to what we currently have. I'm fine with not worrying too much about a future extension in this direction - piecewise linear is probably all we need for quite a while.

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: Alright, I will keep this in mind when implementing the future PR!

FormelessPuppy41:
Hi @N-Wouda,

Per your feedback, I removed `DurationCostFunction.h/.cpp` and now use `PiecewiseLinearFunction` directly. I also removed `PiecewiseLinearFunction.cpp`, since the class is now templated and fully defined in the header.

For bounds behavior, the current implementation throws `std::domain_error` when queried outside the function domain. If you prefer, I can switch to the alternative approach and require the PWL to be defined over the full `Domain` range.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * Parameters
+ * ----------
+ * breakpoints
+ *     Strictly increasing breakpoints. Must contain at least two elements.
+ * segments
+ *     Segment definitions ``(intercept, slope)`` for each interval. Must have
+ *     size ``len(breakpoints) - 1``.
+ */
+template <typename Domain = int64_t, typename Codomain = int64_t>
```
N-Wouda:
```suggestion
template <typename Domain, typename Codomain>
```

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,217 @@
...
+class PiecewiseLinearFunction
+{
+public:
+    using Segment = std::pair<Codomain, Codomain>;
```
N-Wouda:
```suggestion
    using Segment = std::pair<Domain, Domain>;
```

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,217 @@
...
+        requires(!requires(Value const value) { value.get(); })
+    [[nodiscard]] static Value toRaw(Value const value);
+
+    template <typename Value, typename Raw>
+        requires std::is_arithmetic_v<Value>
+    [[nodiscard]] static Value fromRaw(Raw const value);
+
+    template <typename Value, typename Raw>
+        requires(!std::is_arithmetic_v<Value>)
+    [[nodiscard]] static Value fromRaw(Raw const value);
```
N-Wouda: Huh? Why do we need these?

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Segment intercept/slope pairs.
+     */
+    [[nodiscard]] std::vector<Segment> const &segments() const;
+
+    /**
+     * Tests whether this function is the zero function.
+     */
+    [[nodiscard]] bool isZero() const;
```
N-Wouda: Remove

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,217 @@
...
+        if (x < breakpoints_[idx + 1])
+            break;
+
+    if (idx == segments_.size())  // x == breakpoints_.back()
+        idx = segments_.size() - 1;
+
+    auto const &[intercept, slope] = segments_[idx];
+    auto const delta = toRaw(x - breakpoints_[idx]);
+    auto const value = toRaw(intercept) + toRaw(slope) * delta;
+    return fromRaw<Codomain>(value);
```
N-Wouda:
```suggestion
    return static_cast<Codomain>(value);
```

pyvrp/cpp/bindings.cpp
```diff
@@ -35,6 +39,9 @@ using pyvrp::Route;
 using pyvrp::Solution;
 using pyvrp::Trip;
 
+using PiecewiseLinearFunction
+    = pyvrp::PiecewiseLinearFunction<pyvrp::Duration, pyvrp::Cost>;
```
N-Wouda: Probably rather `PiecewiseLinearFunction<int64_t, int64_t>` here, since we're only going to use this object in tests.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,217 @@
...
+    std::vector<Domain> breakpoints, std::vector<Segment> segments)
+    : breakpoints_(std::move(breakpoints)), segments_(std::move(segments))
+{
+    if (breakpoints_.size() < 2)
+    {
+        auto const *msg = "breakpoints must contain at least two values.";
+        throw std::invalid_argument(msg);
+    }
+
+    if (segments_.size() + 1 != breakpoints_.size())
```
N-Wouda:
```suggestion
    if (segments_.size() != breakpoints_.size() - 1)
```
Suggestion cannot overflow and matches error message.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,217 @@
...
+    {
+        auto const *msg = "segments must have size breakpoints.size() - 1.";
+        throw std::invalid_argument(msg);
+    }
+
+    for (size_t idx = 1; idx != breakpoints_.size(); ++idx)
+        if (!(breakpoints_[idx - 1] < breakpoints_[idx]))
```
N-Wouda:
```suggestion
        if (breakpoints_[idx - 1] >= breakpoints_[idx])
```
It's awkward to negate a nested expression when you can easily simplify to directly use the negation.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,217 @@
...
+{
+    return breakpoints_;
+}
+
+template <typename Domain, typename Codomain>
+inline auto PiecewiseLinearFunction<Domain, Codomain>::segments() const
+    -> std::vector<Segment> const &
+{
+    return segments_;
+}
```
N-Wouda: No need to inline these; they're not used anywhere critical

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    for (size_t idx = 1; idx != breakpoints_.size(); ++idx)
+        if (!(breakpoints_[idx - 1] < breakpoints_[idx]))
+            throw std::invalid_argument(
+                "breakpoints must be strictly increasing.");
+}
+
+template <typename Domain, typename Codomain>
+Codomain PiecewiseLinearFunction<Domain, Codomain>::operator()(Domain x) const
```
N-Wouda: Do inline this - this is critical. You can add the `inline` keyword with the declaration, not here in the implementation. That's mostly about style, but doing it like that matches the rest of PyVRP.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -0,0 +1,217 @@
...
+template <typename Domain, typename Codomain>
+inline std::vector<Domain> const &
+PiecewiseLinearFunction<Domain, Codomain>::breakpoints() const
+{
+    return breakpoints_;
+}
+
+template <typename Domain, typename Codomain>
+inline auto PiecewiseLinearFunction<Domain, Codomain>::segments() const
+    -> std::vector<Segment> const &
```
N-Wouda:
```suggestion
inline std::vector<Segment> const &PiecewiseLinearFunction<Domain, Codomain>::segments() const
```
Why the sudden `auto` and arrow notation?

pyvrp/cpp/bindings.cpp
```diff
@@ -17,8 +18,11 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
+#include <cstdint>
+#include <limits>
 #include <memory>
 #include <sstream>
+#include <stdexcept>
```
N-Wouda: I don't think you need this particular header here.

N-Wouda:
```suggestion
```

N-Wouda (commented):
This is a lot closer to what I had in mind initially. I think we're pretty close now.

pyvrp/cpp/bindings.cpp
N-Wouda: (that might cause trouble when we add it to the vehicle type later on with a different specialisation, but I think we can worry about that then, not now)

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: These are just adapter helpers to let the same template work with both plain arithmetic types and PyVRP’s `Measure` types (`Duration`/`Cost`). `toRaw()` unwraps `.get()` when needed, and `fromRaw()` wraps the computed value back. They’re not strictly required, so I can remove them and simplify if you prefer.

N-Wouda: No: the PWL function should not have to know about, or special case, for Measure types or plain primitives. I think these adapters were only needed because there was a typing mistake in the segments - those ought to be of type `Domain`, not `Codomain`. I'm reasonably sure the problem goes away once that's fixed.

N-Wouda:
This looks good to me. I'm going to make a few small changes later tonight:
- Demand that PWL covers the full domain of the `Domain` type;
- Remove the breakpoint delta - that doesn't work for negative numbers, and isn't how PWL functions are typically defined.

I'll merge it in after those changes. Thanks @FormelessPuppy41!

N-Wouda (approved):

N-Wouda:
Thanks @FormelessPuppy41!

Issue #867, opened 2025-07-07
Time Dependent VRP


ljwolf:
#### Is your feature request related to a problem? Please describe

From #441, it seems there's a broader interest in a time-dependent VRP implementation. Is this still the case?

I will be looking into implementing this over the next few months and wondered if such an implementation would be of interest upstream?

#### Describe the solution you'd like

I think this should become part of the profile system. Thus, profiles might take some additional arguments (like, `tw_early`, `tw_late`?) that define the time period over which the profile applies. 

#### Describe alternatives you have considered

I think keeping track of this within the profile makes more sense than keeping track of it as a separate system, but I am not yet sure of the structure of the C++ implementation. 

I think that a continuous-time implementation (so, departure times calculated on demand for specific OD pairs at a specific time) would be quite challenging to solve, and probably out of scope.

N-Wouda:
Hi @ljwolf! We are interested in this, but time-dependent routing is not easy to implement at all. If you're interested in exploring this, you might want to look into [Visser and Spliet (2020)](https://pubsonline.informs.org/doi/10.1287/trsc.2019.0938) and, more recently, [Blauth et al. (2024)](https://www.sciencedirect.com/science/article/pii/S1572528624000276) for an overview of what's already known.

The ideas in those papers get us most of the way there for 'regular' VRPs with time constraints. But another difficulty is that we'd need to also make everything work nicely with trip-dependent release times. That seems quite difficult, too - it was already pretty hard to do in case of constant travel times when I implemented multi-trip a few months back.

ljwolf:
Yes, I'm aware of the release time issue... I'm not sure how I'd solve that at the first brush, but to be sure--an upstream implementation must play nicely with that?

I'd hoped to only implement time dependent departure costs for the common routing with traffic applications, like having a rush hour time cost and an otherwise usual time cost. 

I'll keep this updated as the work unfolds!

N-Wouda:
> an upstream implementation must play nicely with that?

Certainly not for a first attempt at implementing this, but eventually, hopefully yes. We use release times quite a bit in same-day delivery so it's a fairly important feature to us. I don't know how hard it'd be to support release times in a time-dependent setting - I haven't thought long enough about how to modify the algorithms outlined in existing literature for this. Perhaps it's not so hard, but in any case, we'll see once we have a concrete implementation of what's currently still fairly abstract (to me).

> I'll keep this updated as the work unfolds!

That sounds great!

[...]

PR #1052, opened 2026-02-24
Add points constructor to piecewise linear functions

FormelessPuppy41:
This PR is part of #925.


FormelessPuppy41:
Hi @N-Wouda, I implemented #1044 in this PR and all CI checks pass except `docs/readthedocs.org:pyvrp`.  
I did not modify files under `docs/`, so I suspect this may be unrelated to the code changes (the log shows a generic “Unknown problem” during the build).  

Could you please review the implementation and let me know if you want me to investigate the docs failure further?

N-Wouda:
From the docs build:

> Model.add_vehicle_type() got an unexpected keyword argument 'unit_duration_cost'

That's hardly a generic "Unknown problem". Did you update the example notebooks?

FormelessPuppy41:
Thanks for pointing that out. You’re right, I should have checked the raw logs first. I initially only saw the top-level “Unknown problem” in Read the Docs (`https://app.readthedocs.org/projects/pyvrp/builds/31538665/`), but the raw log clearly shows the failure comes from example notebooks still using legacy `unit_duration_cost` / `unit_overtime_cost` arguments. I updated those notebooks! 

Edit: I noticed the markdown in the notebook needs to be updated too. I will modify this now!

pyvrp/cpp/bindings.cpp
```diff
@@ -369,36 +366,76 @@ PYBIND11_MODULE(_pyvrp, m)
...
+            py::arg("profile") = py::none(),
+            py::arg("start_late") = py::none(),
+            py::arg("initial_load") = py::none(),
+            py::arg("reload_depots") = py::none(),
+            py::arg("max_reloads") = py::none(),
+            py::arg("max_overtime") = py::none(),
+            py::arg("duration_cost_function") = py::none(),
+            py::kw_only(),
+            py::arg("name") = py::none(),
+            DOC(pyvrp, ProblemData, VehicleType, replace))
```
N-Wouda: Why the lambda?

pyvrp/cpp/bindings.cpp
```diff
@@ -117,11 +119,10 @@ PYBIND11_MODULE(_pyvrp, m)
...
             {
-                using Breakpoints = std::vector<int64_t>;
-                using Segments = std::vector<PiecewiseLinearFunction::Segment>;
                 return PiecewiseLinearFunction(
-                    t[0].cast<Breakpoints>(),  // breakpoints
-                    t[1].cast<Segments>());    // segments
+                    t[0].cast<std::vector<pyvrp::Duration>>(),  // breakpoints
+                    t[1].cast<std::vector<
+                        PiecewiseLinearFunction::Segment>>());  // segments
             }));
```
N-Wouda, line 172: Revert

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -43,8 +44,11 @@ template <typename Dom, typename Co> class PiecewiseLinearFunction
 private:
     std::vector<Dom> breakpoints_;
     std::vector<Segment> segments_;
+    bool isZero_ = false;
 
 public:
+    PiecewiseLinearFunction();
```
N-Wouda: What are the semantics of this constructor?

N-Wouda: Ah, I got it from the implementation: it initialises a zero PWL function.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -63,9 +67,20 @@ template <typename Dom, typename Co> class PiecewiseLinearFunction
      */
     [[nodiscard]] std::vector<Segment> const &segments() const;
 
+    /**
+     * Returns whether this function is identically zero.
+     */
+    [[nodiscard]] bool isZero() const;
```
N-Wouda: From what I can tell you only need this inside `pyvrp::search::Route`. Just track it there, in that object's state. I don't think this is a meaningful addition to the PWL class.

pyvrp/cpp/ProblemData.cpp
```diff
@@ -472,6 +462,11 @@ size_t ProblemData::VehicleType::maxTrips() const
     return reloadDepots.empty() ? 1 : std::max(maxReloads, maxReloads + 1);
 }
 
+Cost ProblemData::VehicleType::durationCostSlope() const
+{
+    return static_cast<Cost>(durationCostFunction.segments().front().second);
+}
```
N-Wouda: I think we can remove this

pyvrp/_pyvrp.pyi
```diff
@@ -39,14 +39,15 @@ class DynamicBitset:
...
     ) -> None: ...
-    def __call__(self, x: np.int64) -> np.int64: ...
+    def __call__(self, x: int) -> int: ...
     @property
-    def breakpoints(self) -> list[np.int64]: ...
+    def breakpoints(self) -> list[int]: ...
     @property
-    def segments(self) -> list[tuple[np.int64, np.int64]]: ...
+    def segments(self) -> list[tuple[int, int]]: ...
+    def is_zero(self) -> bool: ...
```
N-Wouda: Revert these changes

pyvrp/search/neighbourhood.py
N-Wouda: Just use the duration cost function?

tests/test_PiecewiseLinearFunction.py
```diff
@@ -96,3 +106,52 @@ def test_raises_unsorted_breakpoints():
...
+def test_vehicle_type_replace_updates_duration_cost_function():
+    original = PiecewiseLinearFunction([0], [(0, 1)])
+    updated = PiecewiseLinearFunction([5, _INT_MAX], [(0, 1), (-45, 10)])
+    vehicle_type = VehicleType(duration_cost_function=original)
+
+    replaced = vehicle_type.replace(duration_cost_function=updated)
+    unchanged = vehicle_type.replace()
+
+    assert_equal(replaced.duration_cost_function, updated)
+    assert_equal(unchanged.duration_cost_function, original)
```
N-Wouda: What do vehicle type tests have to do here? Move them to `test_ProblemData.py`, if relevant.

tests/test_ProblemData.py
```diff
@@ -538,25 +542,15 @@ def test_vehicle_type_raises_invalid_data(
...
-)
-def test_vehicle_type_raises_negative_overtime_data(
-    max_overtime: int,
-    unit_overtime_cost: int,
-):
+@pytest.mark.parametrize("max_overtime", [-1])
+def test_vehicle_type_raises_negative_overtime_data(max_overtime: int):
     with assert_raises(ValueError):
-        VehicleType(
-            max_overtime=max_overtime,
```
N-Wouda, line 579: No need to parametrize this anymore

pyvrp/cpp/ProblemData.h
```diff
@@ -500,14 +499,13 @@ class ProblemData
...
-        std::vector<Load> const initialLoad;     // Initially used capacity
-        std::vector<size_t> const reloadDepots;  // Reload locations
-        size_t const maxReloads;                 // Maximum number of reloads
-        Duration const maxOvertime;              // Maximum allowed overtime
-        Cost const unitOvertimeCost;             // Cost per unit of overtime
+        std::vector<Load> const initialLoad;      // Initially used capacity
+        std::vector<size_t> const reloadDepots;   // Reload locations
+        size_t const maxReloads;                  // Maximum number of reloads
+        Duration const maxOvertime;               // Maximum allowed overtime
+        DurationCost const durationCostFunction;  // Duration cost fn
```
N-Wouda: Perhaps call this just `durationCost`. I don't think there's much value in the *function* part of the name; the type already covers that adequately.

pyvrp/cpp/ProblemData.cpp
```diff
@@ -310,14 +311,13 @@ ProblemData::VehicleType::VehicleType(size_t numAvailable,
...
       unitDistanceCost(unitDistanceCost),
-      unitDurationCost(unitDurationCost),
       profile(profile),
       startLate(startLate.value_or(twLate)),
       initialLoad(pad(initialLoad, capacity)),
       reloadDepots(reloadDepots),
       maxReloads(maxReloads),
       maxOvertime(maxOvertime),
-      unitOvertimeCost(unitOvertimeCost),
+      durationCostFunction(std::move(durationCostFunction)),
```
N-Wouda: We need this function to be non-decreasing, so this needs tests to ensure that is the case. I think a helper `bool PiecewiseLinearFunction::isMonotonicallyIncreasing() const` will be useful.

tests/conftest.py
```diff
@@ -145,8 +148,10 @@ def ok_small_overtime(ok_small):
         shift_duration=5_000,
         max_overtime=1_000,
         unit_distance_cost=0,
-        unit_duration_cost=1,
-        unit_overtime_cost=10,
+        duration_cost_function=PiecewiseLinearFunction(
+            [5_000, _INT_MAX],
+            [(0, 1), (-50_000, 11)],
+        ),
```
N-Wouda: Yeah this interface is not very pleasant. Gurobi has [an interface](https://docs.gurobi.com/projects/optimizer/en/current/concepts/modeling/objectives.html#subsubsectionpiecewiseobj) where the user provides (x, y) pairs, and they internally reconstruct the PWL function from those pairs. We could have a look at trying the same?

pyvrp/cpp/bindings.cpp
```diff
@@ -38,7 +38,7 @@ using pyvrp::Solution;
 using pyvrp::Trip;
 
 using PiecewiseLinearFunction
-    = pyvrp::PiecewiseLinearFunction<int64_t, int64_t>;
+    = pyvrp::PiecewiseLinearFunction<pyvrp::Duration, pyvrp::Cost>;
```
N-Wouda: What if later on I want to have a PWL function for distance -> cost?

N-Wouda (commented):
I think this mostly works but there's a few points I don't really understand, and think we might be able to do better.

pyvrp/cpp/bindings.cpp
FormelessPuppy41: The lambda is misplaced here; however, in the new update it will be used to properly 'cast' the PWL to a durationCost (previously durationCostFunction) such that later we can also use a distanceCost that is also a PWL.

FormelessPuppy41:
Hi @N-Wouda ,

Following your feedback, I moved away from exposing `breakpoints + segments` directly.  
I now see two candidate public interfaces, and I’d like your preference.

Option 1: users provide `(x, f(x))` points (this is the Gurobi-style interface you suggested, and what I currently implemented in my local branch)
```python
PiecewiseLinearFunction([
    (0, 0),      # start
    (20, 20),    # slope 1 on [0, 20)
    (30, 80),    # slope 6 on [20, 30)
    (30, 130),   # jump of +50 at x = 30
    (40, 190),   # slope 6 continues after 30
])
```
This works, but users must compute cumulative function values themselves.  
Jumps are modeled via non-decreasing `x` (repeated `x` values).

Option 2: users provide `(x, slope, optional_intercept)`  
- `x`: breakpoint  
- `slope`: unit cost from that breakpoint onward  
- `optional_intercept` (`z`): additional fixed offset at that breakpoint (default `0`)  

```python
PiecewiseLinearFunction([
    (0, 1),        # slope 1 from x = 0
    (20, 6),       # slope 6 from x = 20
    (30, 6, 50),   # keep slope 6, add +50 jump at x = 30
])
```
In this option, jumps are modeled through `z` (additional intercept), and segments are reconstructed internally.

I think option 2 might be more user friendly and closer to the current PyVRP style. What is your preference?

EDIT:
Also for option 2, we could expose two inputs in `VehicleType` (`__init__ `and `replace`): either a raw list `[(x, y[, z])]` or an already-constructed `PiecewiseLinearFunction` instance for `duration_cost_function`. We could also focus on implementing only one of the two options.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -63,9 +67,20 @@ template <typename Dom, typename Co> class PiecewiseLinearFunction
...
+
     bool operator==(PiecewiseLinearFunction const &other) const = default;
 };
 
+template <typename Dom, typename Co>
+PiecewiseLinearFunction<Dom, Co>::PiecewiseLinearFunction()
+    : PiecewiseLinearFunction({std::numeric_limits<Dom>::min()}, {{0, 0}})
+{
+}
+
```
FormelessPuppy41, line 135: I think in a previous thread niels said this should not be exposed.

pyvrp/cpp/ProblemData.h
```diff
@@ -431,9 +429,9 @@ class ProblemData
      * max_overtime
      *     Maximum allowed overtime, on top of the :py:attr:`~shift_duration`.
      *     Default 0, that is, overtime is not allowed.
-     * unit_overtime_cost
-     *     Cost of a unit of overtime. This is in addition to the regular
-     *     :py:attr:`~unit_duration_cost` of route durations. Default 0.
+     * duration_cost_function
+     *     Piecewise linear duration cost function used to evaluate route
+     *     duration costs for this vehicle type. Defaults to the zero function.
      * name
```
FormelessPuppy41, line 464: Shouldn't we also allow users to specify the [(x,y[,z])] values here and we build the PWL?

N-Wouda:
Thanks for thinking along! The second option seems a bit easier than the first because we then don't have to worry about the impicit slopes being integral. So perhaps we can try that approach first?

I'm a bit unsure about how to model the jumps with `z`, but that should become clearer with an implementation of the idea (and a few examples of its use).

FormelessPuppy41:
I thought so too. That is why I already implemented it. I have included an example in the `duration_constraints.ipynb` notebook with a non-zero `z` value.

docs/source/api/pyvrp.rst
```diff
@@ -107,6 +107,13 @@ This object stores the best observed solution and detailed runtime statistics.
       :members:
       :special-members: __call__
 
+      .. note::
+
+         The public constructor accepts ``(breakpoint, slope[, jump])`` points
+         only. Breakpoints must be strictly increasing. The optional ``jump``
+         (default ``0``) adds extra fixed cost at that breakpoint and to the
+         right.
```
N-Wouda: Revert

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -30,21 +42,37 @@ namespace pyvrp
...
 public:
     using Segment = std::pair<Dom, Dom>;
+    struct Point
+    {
+        Dom breakpoint;
+        Dom slope;
+        Dom jump;
+
+        bool operator==(Point const &other) const = default;
+    };
```
N-Wouda:
```suggestion
    using Point = std::tuple<Dom, Dom, Dom>;
```

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -30,21 +42,37 @@ namespace pyvrp
...
 
 private:
     std::vector<Dom> breakpoints_;
     std::vector<Segment> segments_;
 
+    [[nodiscard]] static Dom safeAdd(Dom lhs, Dom rhs, char const *message);
+    [[nodiscard]] static Dom
+    safeSubtract(Dom lhs, Dom rhs, char const *message);
+    [[nodiscard]] static Dom
+    safeMultiply(Dom lhs, Dom rhs, char const *message);
```
N-Wouda: Remove (again)

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -79,9 +211,8 @@ PiecewiseLinearFunction<Dom, Co>::PiecewiseLinearFunction(
             "Number of breakpoints and segments must match.");
 
     for (size_t idx = 0; idx != breakpoints_.size() - 1; ++idx)
-        if (breakpoints_[idx] >= breakpoints_[idx + 1])
-            throw std::invalid_argument(
-                "Breakpoints must be strictly increasing.");
+        if (breakpoints_[idx] > breakpoints_[idx + 1])
+            throw std::invalid_argument("Breakpoints must be non-decreasing.");
```
N-Wouda: Inconsistent with the point constructor; why change this?

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -110,6 +241,34 @@ PiecewiseLinearFunction<Dom, Co>::segments() const
...
+        auto const [nextIntercept, nextSlope] = segments_[idx];
+
+        auto const left = static_cast<long double>(prevIntercept)
+                          + static_cast<long double>(prevSlope) * breakpoint;
+        auto const right = static_cast<long double>(nextIntercept)
+                           + static_cast<long double>(nextSlope) * breakpoint;
+
+        if (right < left)
+            return false;
+    }
```
N-Wouda: Do *not* cast to double. Why have `Co` if it's not being used?

pyvrp/search/neighbourhood.py
```diff
@@ -155,20 +155,32 @@ def _compute_proximity(
...
+        breakpoints_list = cast("list[int]", state[0])
+        segments_list = cast("list[tuple[int, int]]", state[1])
+        key = (
             veh_type.unit_distance_cost,
-            veh_type.unit_duration_cost,
             veh_type.profile,
+            tuple(breakpoints_list),
+            tuple(segments_list),
+        )
+        unique_edge_costs.setdefault(key, key)
```
N-Wouda: There's `__eq__` on PWL functions already. Make PWL functions immutable (easy), and implement a reasonable hash value for them. This is very awkward.

pyvrp/search/neighbourhood.py
```diff
@@ -155,20 +155,32 @@ def _compute_proximity(
...
+    first, *rest = unique_edge_costs.values()
+    unit_dist, prof, breakpoints_tuple, segments_tuple = first
+    edge_costs = unit_dist * distances[prof] + _duration_cost_values(
+        breakpoints_tuple, segments_tuple, durations[prof]
+    )
+
+    for unit_dist, prof, breakpoints_tuple, segments_tuple in rest:
+        mat = unit_dist * distances[prof] + _duration_cost_values(
+            breakpoints_tuple, segments_tuple, durations[prof]
         )
```
N-Wouda: You might need to `np.vectorize` the PWL function, but it should be fine to evaluate?

N-Wouda (commented):
This has kind of gone of the rails. I think you need to simplify *a lot*. Remove the unnecessary changes, and start smaller. Perhaps we begin with just the alternative PWL function constructor? That seems to have caused a lot of further bloating in this PR.

FormelessPuppy41:
Hi @N-Wouda ,
I made the same mistake as in the previous PR by over-scoping😅. I will keep the scope of review requests narrow from now on.

I reverted all `test_*.py` changes except `tests/test_PiecewiseLinearFunction.py`, and I reverted the API/model changes so `unit_duration_cost` and `unit_overtime_cost` are reinstated.

For now, I will first finalize the PWL interface/behavior, then integrate it into the rest of PyVRP instead of trying it all at once.  
In the latest commits, I focused only on PWL: I added the new `PiecewiseLinearFunction(points=[(x, y[, z])])` constructor, removed the legacy `(breakpoints, segments)` constructor, and simplified `PiecewiseLinearFunction.h` in line with your feedback.

The notes below are TODOs for me after the PWL part is accepted (not action items for this review):

1. Add a `DurationCostFunction = ProblemData::VehicleType::DurationCost` mapping from the PWL class, so we could later support a `DistanceCostFunction` in the same way.  
2. Implement the earlier review comments on `neighbourhood.py`.  
3. Expose the PWL feature to users as `duration_cost` (not `duration_cost_function`) in the `VehicleType`.  
a. Decide what `VehicleType` should accept: a `PiecewiseLinearFunction` instance, a list of points (`[(x, y[, z])]`), or both.  
4. Extend the notebook examples: extend the current example by, e.g., adding a surcharge jump at duration 25 to model overtime/bonus pay.

If you already have a preference for 3a, I’d appreciate your guidance so I can keep that in mind.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -30,23 +43,23 @@ namespace pyvrp
...
 private:
     std::vector<Dom> breakpoints_;
     std::vector<Segment> segments_;
 
 public:
-    PiecewiseLinearFunction(std::vector<Dom> breakpoints,
-                            std::vector<Segment> segments);
+    PiecewiseLinearFunction();
+
+    PiecewiseLinearFunction(std::vector<Point> points);
```
N-Wouda, line 63: Add this constructor, keep the old one. Then revert some of the unnecessary changes.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -63,25 +76,70 @@ template <typename Dom, typename Co> class PiecewiseLinearFunction
...
 template <typename Dom, typename Co>
 PiecewiseLinearFunction<Dom, Co>::PiecewiseLinearFunction(
-    std::vector<Dom> breakpoints, std::vector<Segment> segments)
-    : breakpoints_(std::move(breakpoints)), segments_(std::move(segments))
+    std::vector<Point> points)
 {
-    if (segments_.empty())
-        throw std::invalid_argument("Need at least one segment.");
+    static_assert(std::is_integral_v<Dom>,
+                  "PiecewiseLinearFunction expects an integral domain type.");
```
N-Wouda: Why do we require this? What breaks for floating point PWL functions?

N-Wouda:
> 3\. Expose the PWL feature to users as `duration_cost` (not `duration_cost_function`) in the `VehicleType`.
>     a. Decide what `VehicleType` should accept: a `PiecewiseLinearFunction` instance, a list of points (`[(x, y[, z])]`), or both.

Definitely the function instance, not its constructor arguments. The indirection of the latter isn't really helpful.

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: My original reasoning was that in our use case the domain represents duration/time, which is integral, so I added the restriction on `Dom` for that reason. However, that is an application-specific assumption, not a requirement of a generic PWL definition, so I agree it should not be enforced here.

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: Just to confirm my understanding: 
we should keep both constructors; one taking `points` and one taking `breakpoints` and `segments`. Internally, the object can initialize itself based on the constructor that is used, while preserving the existing interface for `breakpoints_`, `segments_`, and `operator()(Dom x)`.

I’ll also reintroduce the tests for the `breakpoints` / `segments` constructor and revert the changes in the `bindings.cpp` file since the `breakpoints` / `segments` constructor is available again.

N-Wouda: Yes, I think so. We can expose both from C++ to Python (using `@typing.overload` in the type stubs to differentiate). Nothing changes for existing code, but we can use the points constructor when explaining the PWL function concept.

FormelessPuppy41: I updated the PWL implementation to expose both constructors: `PiecewiseLinearFunction(points=[(x, y[, z])])` and `PiecewiseLinearFunction(breakpoints, segments)`.  
Internally, the representation is now unified as strictly increasing `breakpoints` with `len(segments) = len(breakpoints) + 1`, so each breakpoint has a left and right segment, including the last one. This removes the need for a duplicated final breakpoint to model jumps, while keeping pickle/state based on `breakpoints` and `segments`.

FormelessPuppy41:
Hi @N-Wouda,

I am currently adding some additional tests for `PiecewiseLinearFunction.h`. Since we now have two constructors, should these tests use the explicit form
```python
PiecewiseLinearFunction(points=[(0, 1), (5, 1, 2), (10, 1, 3)])
PiecewiseLinearFunction(
    breakpoints=[0, 5],
    segments=[(0, 1), (0, 1), (3, 1)],
)
```
or the positional form
```python
PiecewiseLinearFunction([(0, 1), (5, 1, 2), (10, 1, 3)])
PiecewiseLinearFunction([0, 5], [(0, 1), (0, 1), (3, 1)])
```
I suspect the explicit form is a bit clearer, but I wanted to check your preference first.

N-Wouda:
We'll need both in the tests, so both? I don't really mind either way, as long as we have tests that check both forms result in the same object (for equivalent input data).

pyvrp/cpp/bindings.cpp
```diff
@@ -40,6 +41,31 @@ using pyvrp::Trip;
...
+            throw std::invalid_argument(
+                "Each point must be (breakpoint, slope) or "
+                "(breakpoint, slope, jump).");
+
+        slopePoints.push_back({seq[0].cast<int64_t>(),
+                               seq[1].cast<int64_t>(),
+                               size == 3 ? seq[2].cast<int64_t>() : 0});
+    }
+
+    return slopePoints;
```
N-Wouda: Probably best to just put this in the custom init directly, not here.

pyvrp/cpp/bindings.cpp
```diff
@@ -117,6 +150,9 @@ PYBIND11_MODULE(_pyvrp, m)
             },
             [](py::tuple t)  // __setstate__
             {
+                if (t.size() != 2)
+                    throw std::invalid_argument("Invalid state.");
+
```
N-Wouda:
```suggestion
```

FormelessPuppy41:
Hi @N-Wouda, I updated the PR with the latest changes and wanted to follow up on the review request. When you get a chance, I would be grateful if you could have a look!

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -1,50 +1,79 @@
...
+ * an additional fixed cost that is applied at that breakpoint and to the
+ * right. When omitted, ``jump`` defaults to zero.
+ *
+ * The first point also defines extrapolation to the left. In particular, for
+ * a first point ``(x_0, s_0, z_0)``, we have ``f(x_0) = z_0`` and slope
+ * ``s_0`` for :math:`x < x_0`.
+ *
+ * Internally, the function is represented via :math:`n` strictly increasing
+ * breakpoints :math:`(b_i)_{i = 0, \ldots, n - 1}` and :math:`n + 1` linear
+ * segments :math:`(f_i)_{i = 0, \ldots, n}`. Given :math:`x`, define
```
N-Wouda: Define on each constructor, not here (this doesn't work).

pyvrp/cpp/bindings.cpp
```diff
@@ -108,6 +139,8 @@ PYBIND11_MODULE(_pyvrp, m)
                                &PiecewiseLinearFunction::segments,
                                py::return_value_policy::reference_internal,
                                DOC(pyvrp, PiecewiseLinearFunction, segments))
+        .def("is_monotonically_increasing",
+             &PiecewiseLinearFunction::isMonotonicallyIncreasing)
```
N-Wouda: DOC?

pyvrp/cpp/PiecewiseLinearFunction.h
N-Wouda: See https://pyvrp--1052.org.readthedocs.build/en/1052/api/pyvrp.html#pyvrp._pyvrp.PiecewiseLinearFunction

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -63,9 +92,63 @@ template <typename Dom, typename Co> class PiecewiseLinearFunction
...
+    std::vector<Dom> values(points.size());
+    values[0] = std::get<2>(points[0]);  // f(x_0) equals the first jump value.
+
+    for (size_t idx = 1; idx != points.size(); ++idx)
+    {
+        auto const dx = std::get<0>(points[idx]) - std::get<0>(points[idx - 1]);
+        auto const delta = std::get<1>(points[idx - 1]) * dx;
+        auto const leftValue = values[idx - 1] + delta;
+        values[idx] = leftValue + std::get<2>(points[idx]);
+    }
```
N-Wouda: I think you can skip this allocation and instead just track the last value when building the segments. You don't need a vector for that.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -110,6 +188,34 @@ PiecewiseLinearFunction<Dom, Co>::segments() const
...
+    for (size_t idx = 0; idx != breakpoints_.size(); ++idx)
+    {
+        auto const breakpoint = static_cast<Co>(breakpoints_[idx]);
+        auto const [prevIntercept, prevSlope] = segments_[idx];
+        auto const [nextIntercept, nextSlope] = segments_[idx + 1];
+
+        auto const left = static_cast<Co>(prevIntercept)
+                          + static_cast<Co>(prevSlope) * breakpoint;
+        auto const right = static_cast<Co>(nextIntercept)
+                           + static_cast<Co>(nextSlope) * breakpoint;
```
N-Wouda: Pretty sure all this casting is completely unnecessary. breakpoint and segment have the same type.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -110,6 +188,34 @@ PiecewiseLinearFunction<Dom, Co>::segments() const
...
+
+template <typename Dom, typename Co>
+bool PiecewiseLinearFunction<Dom, Co>::isMonotonicallyIncreasing() const
+{
+    for (auto const &segment : segments_)
+    {
+        auto const slope = segment.second;
+        if (slope < 0)
+            return false;
+    }
```
N-Wouda:
```suggestion
    for (auto const [_, slope] : segments_)
        if (slope < 0)
            return false;
```

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -87,14 +169,10 @@ PiecewiseLinearFunction<Dom, Co>::PiecewiseLinearFunction(
...
-        if (x < breakpoints_[idx])
-        {
-            auto const [intercept, slope] = segments_[idx];
-            return static_cast<Co>(intercept + slope * x);
-        }
-
-    auto const [intercept, slope] = segments_.back();
+    auto const idx = static_cast<size_t>(
+        std::upper_bound(breakpoints_.begin(), breakpoints_.end(), x)
+        - breakpoints_.begin());
```
N-Wouda: Pretty sure a loop is faster for the data we'll actually have; why change this?

pyvrp/cpp/bindings.cpp
```diff
@@ -92,10 +118,15 @@ PYBIND11_MODULE(_pyvrp, m)
...
         m, "PiecewiseLinearFunction", DOC(pyvrp, PiecewiseLinearFunction))
+        .def(py::init<>())
         .def(py::init<std::vector<int64_t>,
                       std::vector<PiecewiseLinearFunction::Segment>>(),
              py::arg("breakpoints"),
              py::arg("segments"))
+        .def(py::init(
+                 [](py::iterable points)
+                 { return PiecewiseLinearFunction(parseSlopePoints(points)); }),
+             py::arg("points"))
```
N-Wouda: This oen first, then the breakpoints/segments init, then the empty one.

(pybind searches them in order, and the points constructor is most likely to be used)

tests/test_PiecewiseLinearFunction.py
N-Wouda, line 1: Don't update the existing tests unless they're no longer valid. In this case I think they still are? If so, add new ones for the new functionality.

(prefer to keep diffs small, and don't touch what you don't have to. That makes it much easier to review and keep track of changes.)

pyvrp/cpp/bindings.cpp
```diff
@@ -40,6 +41,31 @@ using pyvrp::Trip;
...
+    std::vector<PiecewiseLinearFunction::Point> slopePoints;
+
+    for (auto const &point : points)
+    {
+        auto const seq = py::cast<py::sequence>(point);
+        auto const size = seq.size();
+        if (size != 2 && size != 3)
+            throw std::invalid_argument(
+                "Each point must be (breakpoint, slope) or "
+                "(breakpoint, slope, jump).");
```
N-Wouda: It's now a 'jump' at the breakpoint $x$. Is it easier if we instead say that the optional third value is the value at $x$, so $f(x)$, and we'll interpolate it from the other points if not provided?

N-Wouda:
> Hi @N-Wouda, I updated the PR with the latest changes and wanted to follow up on the review request. When you get a chance, I would be grateful if you could have a look!

Hi @FormelessPuppy41, sorry - this fell of my radar a bit. I just had a look, thanks for reminding me!

pyvrp/cpp/bindings.cpp
FormelessPuppy41: This is definitely possible, and I think both interpretations are valid.

For example, at `x = 3h`, suppose we want slope `20 €/h` before 3h, a jump of `50 €` at 3h, and slope `30 €/h` after 3h.

- `f(x)` semantics: `PWL(points=[(0, 20), (3, 30, 110)])`
- `jump` semantics: `PWL(points=[(0, 20), (3, 30, 50)])`

With `f(x)` semantics, users must manually compute `f(3) = 20 * 3 + 50 = 110`.  
With `jump` semantics, users who think in absolute breakpoint values must manually compute `jump = 110 - 20 * 3 = 50`.

So when many breakpoints are involved, both approaches can be error-prone; the conversion burden just shifts depending on whether users think in local jumps or absolute values. So I think we should either decide which use case is most common and stick to that, or use a compromise, e.g. support both explicitly via `PWL.from_values(points=[...])` and `PWL.from_jumps(points=[...])` (or `PWL(points=[...], third_value="value" | "jump")`).

If you prefer to keep only one option, I can also switch to only `f(x)` semantics or keep only `jump` semantics. Please let me know what you prefer.

pyvrp/cpp/bindings.cpp
FormelessPuppy41: Should we also keep the constructor declarations/docblocks in `PiecewiseLinearFunction.h` in the same order as the bindings?

Right now I mapped docs by overload index in `bindings.cpp` (`1 = default`, `2 = points`, `3 = breakpoints+segments`) and it works. I’m wondering whether we should also keep `PiecewiseLinearFunction.h` ordered the same way to make this mapping more obvious and reduce the chance of mismatches later. What do you think?

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: I changed this to `std::upper_bound` because the breakpoints are sorted, and in an earlier review we discussed that binary search is a better default once the number of breakpoints grows.

In my personal use case I expect around `~50` breakpoints, and this function will be called many times, so `O(log n)` lookup seemed more reasonable. That said, for small `n` a linear scan can still be faster due to lower constant overhead. So, if you prefer a loop then I can revert the change.

FormelessPuppy41:
Hi @N-Wouda,

Thanks for the review. I’ve implemented your suggestions, except for the `jump` vs `f(x)` discussion and the `parseSlopePoints` helper, since that helper kind of depends on which semantics we choose there. I also kept the binary-search lookup for now, based on the rationale in that comment thread.

Could you share your preference on those remaining comments? Then I will implement the final changes and push the commit.

pyvrp/cpp/PiecewiseLinearFunction.h
N-Wouda: 50 breakpoints are ~7 cache lines of data, nicely sequentially organised inside the vector. I don't think your CPU meaningfully benefits from binsearch at that size. (it might, but I doubt it)

Perhaps a benchmark can tell? I'd write a small C++ program to run a few different sizes of PWL functions, and then roughly determine at which size binsearch is faster. Calling the PWL function from Python doesn't work for benchmarking, the Python -> C++ overhead is too large for that.

pyvrp/cpp/bindings.cpp
N-Wouda: Yeah probably, that is a bit more consistent.

pyvrp/cpp/bindings.cpp
N-Wouda: I think we should ideally have just one points-based constructor, so not both. My 'ideal' constructor is still the Gurobi-style one of $(x, f(x))$ points. I don't like the fiddlyness of the jumps, and it's a bit awkward that we need to rely on the previous segment's [intercept, slope] pairs at $x$ when technically $x$ is outside that segment's domain (it's the next breakpoint, and we defined segment domains to exclude those).

The domain issue can probably be fixed if we make domains inclusive of the next breakpoint, just like Gurobi. If we did $(x, f(x))$ points, we'd then just have soemthing like $(0, 0), (3, 60), (3, 110), (4, 140)$ with the Gurobi-interface. That looks a lot saner to me than optional jumps and the associated type messiness of differently sized tuples.

Would you be willing to try something like this? I think we can throw an exception if the implied slopes are non-integral.

pyvrp/cpp/bindings.cpp
FormelessPuppy41: Okay, i will also reordered the constructors.

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: Hi @N-Wouda ,
I let Claude run a small benchmark to compare `std::upper_bound` vs a linear scan. Code:

```cpp
// Compile: g++ -O2 -std=c++20 -o bench_pwl bench_pwl.cpp

#include "PiecewiseLinearFunction.h"
#include <chrono>
#include <cstdint>
#include <iostream>
#include <vector>

using PWL = pyvrp::PiecewiseLinearFunction<int64_t, int64_t>;

int64_t evalLinear(PWL const &fn, int64_t x)
{
    auto const &bp = fn.breakpoints();
    size_t idx = 0;
    while (idx < bp.size() && x >= bp[idx])
        ++idx;
    auto const [intercept, slope] = fn.segments()[idx];
    return intercept + slope * x;
}

template <typename F>
double bench(F &&f, int64_t n_iters)
{
    auto t0 = std::chrono::high_resolution_clock::now();
    volatile int64_t sink = 0;
    for (int64_t i = 0; i < n_iters; ++i)
        sink += f(i);
    auto t1 = std::chrono::high_resolution_clock::now();
    return std::chrono::duration<double, std::nano>(t1 - t0).count() / n_iters;
}

PWL makePWL(int n_breakpoints)
{
    std::vector<std::pair<int64_t, int64_t>> points;
    for (int i = 0; i <= n_breakpoints; ++i)
        points.push_back({i * 10, i * 10});
    return PWL(points);
}

int main()
{
    constexpr int64_t N = 5'000'000;
    std::cout << "breakpoints\tbinsearch(ns)\tlinear(ns)\n";
    for (int n : {1, 5, 10, 20, 30, 40, 50, 100, 200, 500})
    {
        auto fn = makePWL(n);
        double t_bin = bench([&](int64_t i) { return fn(i % (n * 10)); }, N);
        double t_lin = bench([&](int64_t i) { return evalLinear(fn, i % (n * 10)); }, N);
        std::cout << n << "\t\t" << t_bin << "\t\t" << t_lin << "\n";
    }
}
```

On my computer, this resulted in (with 5M iterations per size):

Breakpoints | Binary search (ns) | Linear scan (ns)
-- | -- | --
1 | 2.99 | 2.99
5 | 3.85 | 2.40
10 | 4.16 | 3.12
20 | 5.46 | 4.98
30 | 5.26 | 6.69
40 | 6.65 | 8.30
50 | 6.58 | 10.25
100 | 7.92 | 20.58
200 | 9.37 | 38.58
500 | 10.96 | 87.63

<br class="Apple-interchange-newline">

The results show that the crossover point is somewhere between 20 and 30 breakpoints. Around 30 breakpoints, binary search starts to outperform linear scan, and by 50 breakpoints it is already about 1.5–1.6× faster. The gap then widens steadily as the number of breakpoints increases.

For smaller inputs, linear scan still has an advantage: at 5 breakpoints it is about 1.6× faster, and the two approaches are essentially identical at 1 breakpoint.

Let me know whether you’d like to keep the current binary-search approach or switch to a linear scan!

pyvrp/cpp/bindings.cpp
FormelessPuppy41: Hi @N-Wouda,

I've updated the constructor to use Gurobi-style `(x, f(x))`` points. I left two open questions as FIXMEs in the code. I am happy to resolve them however you prefer:

1.  **Minimum number of points.** Currently at least two points are required, since we need two distinct x-values to determine the extrapolation slope. Should we keep it this way, or would a single-point input be useful (e.g. implicitly anchoring through `(0, 0))`?
2. **Legacy breakpoints/segments constructor.** We now expose three constructors: `PWL(points=...)`, `PWL(breakpoints=..., segments=...)`, and `PWL()`. I've kept the second one for now, but since `(x, f(x))` is now the primary interface, we could remove it to simplify the API. Do you want to keep it?

pyvrp/cpp/PiecewiseLinearFunction.h
N-Wouda: Let's do binsearch then! I tinkered with this locally a bit, and I get fairly similar results - for me the switch is conclusive around 30 or so breakpoints.

I think we'll, on average, have pretty simple PWL functions as input data - say <10 breakpoints in general. But with time dependent routing (#867) the number of breakpoints quickly increases as route segments are formed, and it's not unlikely that there could be hundreds of breakpoints needed to evaluate a full segment. Clearly a linear search cannot handle that, while binsearch should be fine.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -87,14 +227,10 @@ PiecewiseLinearFunction<Dom, Co>::PiecewiseLinearFunction(
...
```
N-Wouda:
```suggestion
    auto const idx = std::distance(
        breakpoints_.begin(),
        std::upper_bound(breakpoints_.begin(), breakpoints_.end(), x));
```

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -110,6 +246,31 @@ PiecewiseLinearFunction<Dom, Co>::segments() const
...
 }
+
+template <typename Dom, typename Co>
+bool PiecewiseLinearFunction<Dom, Co>::isMonotonicallyIncreasing() const
+{
+    for (auto const &[_, slope] : segments_)
+    {
+        if (slope < 0)
+            return false;
+    }
```
N-Wouda:
```suggestion
    for (auto const [_, slope] : segments_)
        if (slope < 0)
            return false;
```

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -63,9 +118,95 @@ template <typename Dom, typename Co> class PiecewiseLinearFunction
...
+    : PiecewiseLinearFunction({}, {{0, 0}})
+{
+}
+
+template <typename Dom, typename Co>
+PiecewiseLinearFunction<Dom, Co>::PiecewiseLinearFunction(
+    std::vector<Point> points)
+{
+    // FIXME: ask Niels whether a minimum of two points is preferred or a
+    // single-point value should be supported.
```
N-Wouda: I don't know tbh. What would the semantics of a single point be? How do we extrapolate from just one point?

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -1,53 +1,108 @@
...
     PiecewiseLinearFunction(std::vector<Dom> breakpoints,
                             std::vector<Segment> segments);
 
+    /**
+     * PiecewiseLinearFunction()
+     *
+     * Creates the zero function represented by a single segment with zero slope
+     * and zero intercept.
+     */
+    PiecewiseLinearFunction();
```
N-Wouda: Do we actually need this constructor?

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -1,53 +1,108 @@
...
  *    f(x) =
  *    \begin{cases}
- *       f_1(x) & \text{ if } x < b_1, \\
+ *       f_0(x) & \text{ if } x < b_0, \\
  *       f_i(x) & \text{ if } b_{i - 1} \le x < b_i
- *                \text{ for } i = 2, \ldots, n, \\
+ *                \text{ for } i = 1, \ldots, n - 1, \\
  *       f_n(x) & \text{ otherwise.}
  *    \end{cases}
  *
```
N-Wouda: This no longer shows up in the docs for me.

pyvrp/cpp/bindings.cpp
```diff
@@ -92,10 +93,30 @@ PYBIND11_MODULE(_pyvrp, m)
...
+                         if (seq.size() != 2)
+                             throw std::invalid_argument(
+                                 "Each point must be a (x, f(x)) pair.");
+                         valuePoints.push_back(
+                             {seq[0].cast<int64_t>(), seq[1].cast<int64_t>()});
+                     }
+                     return PiecewiseLinearFunction(valuePoints);
+                 }),
+             py::arg("points"),
+             DOC(pyvrp, PiecewiseLinearFunction, PiecewiseLinearFunction, 1))
```
N-Wouda: I think this is unnecessary now? Binding the actual constructor should work.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -1,53 +1,108 @@
...
+     *     points: list[tuple[np.int64, np.int64]],
+     * )
+     *
+     * Creates a piecewise linear function from ordered ``(x, f(x))`` points.
+     *
+     * Points must be non-decreasing in ``x``. Jumps are represented by two
+     * consecutive points sharing the same ``x``-coordinate. The first two and
+     * last two points must have distinct ``x``-coordinates (jumps at the
+     * endpoints are not supported). The first and last segments are used for
+     * extrapolation beyond the endpoints.
```
N-Wouda: Maybe add a link to the Gurobi docs for further information.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -1,53 +1,108 @@
...
+     * ------
+     * ValueError
+     *     When fewer than two points are provided, when the first two or last
+     *     two points share an ``x``-coordinate, when points are not
+     *     non-decreasing in ``x``, or when the implied slope is not integral.
+     */
+    PiecewiseLinearFunction(std::vector<Point> points);
+
+    // FIXME: consider removing this constructor once the points constructor is
+    // the standard interface.
```
N-Wouda: We'll need this one anyway for serialisation.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -63,9 +118,95 @@ template <typename Dom, typename Co> class PiecewiseLinearFunction
...
+            auto const xNext = points[j + 1].first;
+            auto const yNextLeft = points[j + 1].second;
+            slopeRight = (yNextLeft - yRight) / (xNext - x);
+            prevSlope = slopeRight;
+        }
+        else
+        {
+            // Right extrapolation: same slope as last segment.
+            slopeRight = prevSlope;
+        }
```
N-Wouda: Just assign before the if, remove this else branch.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -63,9 +118,95 @@ template <typename Dom, typename Co> class PiecewiseLinearFunction
...
+
+        breakpoints_.push_back(x);
+        segments_.push_back({yRight - slopeRight * x, slopeRight});
```
N-Wouda:
```suggestion
        segments_.emplace_back(yRight - slopeRight * x, slopeRight);
```

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -63,9 +118,95 @@ template <typename Dom, typename Co> class PiecewiseLinearFunction
...
+            if (dy % dx != 0)
+                throw std::invalid_argument("Implied slope is not integral.");
+        }
+    }
+
+    // Left extrapolation: slope of first segment (between first two points).
+    auto const firstSlope = (points[1].second - points[0].second)
+                            / (points[1].first - points[0].first);
+    segments_.push_back(
+        {points[0].second - firstSlope * points[0].first, firstSlope});
```
N-Wouda:
```suggestion
    segments_.emplace_back(
        points[0].second - firstSlope * points[0].first, firstSlope);
```

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -63,9 +118,95 @@ template <typename Dom, typename Co> class PiecewiseLinearFunction
...
+    Dom prevSlope = firstSlope;
+
+    while (i < points.size())
+    {
+        auto const x = points[i].first;
+
+        // Find the last point at this x (handles jumps).
+        size_t j = i;
+        while (j + 1 < points.size() && points[j + 1].first == x)
+            ++j;
```
N-Wouda: Use an iterator, remove outer while loop. You don't need `i` or `j`.
```suggestion
    for (auto it = points.begin(); it != points.end(); ++it)
    {
        auto const x = it->first;

        // Find the last point at this x (handles jumps).
        auto right = it;
        while (++right != points.end() && right->first == x)
             ;
```
etc.

tests/test_PiecewiseLinearFunction.py
N-Wouda: Could you have a look at this again and determine what's actually relevant/needed and what's just bloat? I don't see much value in having each test construct the same object in different ways.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -63,9 +118,95 @@ template <typename Dom, typename Co> class PiecewiseLinearFunction
...
+            // Right extrapolation: same slope as last segment.
+            slopeRight = prevSlope;
+        }
+
+        breakpoints_.push_back(x);
+        segments_.push_back({yRight - slopeRight * x, slopeRight});
+
+        i = j + 1;
+    }
+}
```
N-Wouda, line 134: Down  here, for good measure, maybe assert the following:
```cpp
assert(!segments_.empty());
assert(breakpoints_.size() + 1 == segments_.size());
```

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: Now that you mention it; no, not likely. 

I thought we could use this constructor to easily initialise a 'zero-function PWL', similar to having unit_duration/overtime_cost=0. However, we could just use one of the other constructors.

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: > 1. **Minimum number of points.** Currently at least two points are required, since we need two distinct x-values to determine the extrapolation slope. Should we keep it this way, or would a single-point input be useful (e.g. implicitly anchoring through `(0, 0))`?

We could support a single point by implicitly anchoring through `(0, 0)`, giving the slope as `f(x) / x`. For example, `PWL(points=[(1, 3)])` would define `f(x) = 3x`. This saves users from having to spell out `(0, 0) `explicitly for linear through origin cost functions. 

It only helps when the function passes through the origin, but in practice that covers most cases: the current `unit_duration_cost` and `unit_overtime_cost` are both linear through the origin, and any non-zero fixed component would typically go in `fixed_cost` on the vehicle type rather than be embedded in the PWL.

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: Before `commit 157dd1f` (`8e28a27`) I had `for (auto const [_, slope] : segments_)` (without `&`), which caused the `docs/readthedocs.org:pyvrp` workflow to fail. GCC 11 treats the unnecessary copy as an error under `-Werror=range-loop-construct`. I added `&` in `157dd1f` to fix it, this suggestion would reintroduce that failure.

pyvrp/cpp/PiecewiseLinearFunction.h
FormelessPuppy41: That is weird. For me it does show up at; https://pyvrp--1052.org.readthedocs.build/en/1052/api/pyvrp.html#pyvrp._pyvrp.PiecewiseLinearFunction

tests/test_PiecewiseLinearFunction.py
FormelessPuppy41: The test file had indeed become quite messy, so I restructured it to follow a logical order (validation → constructor equivalence → properties → evaluation → jump semantics → monotonicity → equality → pickle). In doing so I removed several tests that were redundant; they were either constructing the same object via both constructors without testing anything new, or duplicating assertions already covered elsewhere. I understand this results in a larger diff than ideal, but I think the improved structure justifies it.

FormelessPuppy41:
Hi @N-Wouda ,
Thanks for the review. I've addressed most comments inline. One thing that still needs a decision: the single-point input question. I have outlined a possible approach (implicitly anchoring through `(0, 0)` so that `PWL(points=[(1, 3)])` defines `f(x) = 3x`), but happy to keep the two-point minimum if you prefer. Let me know what you think.

N-Wouda:
@FormelessPuppy41 thanks - could you resolve addressed comments rather than 👍 on them?

pyvrp/cpp/PiecewiseLinearFunction.h
N-Wouda: Sorry, I meant to say that the constructors (and their signatures) are no longer showing up. Which makes sense, only this docstring is parsed for the docs.

pyvrp/cpp/PiecewiseLinearFunction.h
N-Wouda: These do not render for me.

pyvrp/cpp/PiecewiseLinearFunction.h
N-Wouda: > It only helps when the function passes through the origin, but in practice that covers most cases: the current `unit_duration_cost` and `unit_overtime_cost` are both linear through the origin, and any non-zero fixed component would typically go in `fixed_cost` on the vehicle type rather than be embedded in the PWL.

That covers only only your envisioned use case. For time-dependent routing and e-VRP this wouldn't work - for those, typically, $f(0) \ne 0$.

The only reasonable thing I can think of is that the single-point constructor defines a constant function. But we don't really need to add that now.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -74,11 +141,10 @@ PiecewiseLinearFunction<Dom, Co>::PiecewiseLinearFunction(
...
-        throw std::invalid_argument(
-            "Number of breakpoints and segments must match.");
+    if (breakpoints_.size() + 1 != segments_.size())
+        throw std::invalid_argument("Need one more segment than breakpoints.");
 
-    for (size_t idx = 0; idx != breakpoints_.size() - 1; ++idx)
+    for (size_t idx = 0; idx + 1 < breakpoints_.size(); ++idx)
         if (breakpoints_[idx] >= breakpoints_[idx + 1])
             throw std::invalid_argument(
                 "Breakpoints must be strictly increasing.");
```
N-Wouda, line 150: Need to check if this matches with the points constructor.

pyvrp/cpp/PiecewiseLinearFunction.h
```diff
@@ -87,14 +153,11 @@ PiecewiseLinearFunction<Dom, Co>::PiecewiseLinearFunction(
...
-            auto const [intercept, slope] = segments_[idx];
-            return static_cast<Co>(intercept + slope * x);
-        }
+    auto const idx = std::distance(
+        breakpoints_.begin(),
+        std::upper_bound(breakpoints_.begin(), breakpoints_.end(), x));
 
-    auto const [intercept, slope] = segments_.back();
+    auto const [intercept, slope] = segments_[idx];
     return static_cast<Co>(intercept + slope * x);
```
N-Wouda, line 161: If we get points $(10, 8)$ and $(10, 12)$, there's a jump at $10$, but both $f(10) = 8$ and $f(10) = 12$ are valid function values.

N-Wouda: Possibly by making the endpoint inclusive/overlapping/figuring out what's the best return value at each point. But that's for tomorrow, it is late now.

N-Wouda:
@FormelessPuppy41 I've rewritten this basically from scratch to fix a few remaining issues, particularly w.r.t. breakpoint and segment duplication. There's one more thing I haven't figured out (see comment above), but I'll look into that tomorrow.

After that it should be good to go. We'll probably need a few more casts here and there, but we can figure that out once we start using it for e.g. `duration_cost`.

pyvrp/cpp/PiecewiseLinearFunction.h
N-Wouda: We can't have both right now, except perhaps with floating point support (#623?). In that case, $f(10 - \epsilon) \approx 8$, while $f(10 + \epsilon) \approx 12$.

PR #1054, opened 2026-02-25
Move `compute_neighbours` to C++

### LINKED ISSUE #1049: Rewrite `compute_neighbours` in C++ to reduce memory overhead

Currently, `compute_neighbours` has a 2-3x memory overhead over just `data`. That's problematic for large instance sizes, and caused because we create a few temporary matrices to evaluate candidate neighbours. I think we can entirely avoid those matrices if we rewrite this in C++. We need this to be fast and low-memory, so that seems like the best way forward.

Reimplement `compute_neighbours` in C++. Keep existing functionality, so nothing changes from a user perspective.

N-Wouda:
This PR:

- [x] Closes #1049.

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/pyvrp/cpp/search/neighbourhood.cpp
+++ b/pyvrp/cpp/search/neighbourhood.cpp
@@ -1,0 +1,28 @@
+#include "neighbourhood.h"
+
+#include <stdexcept>
+
+using pyvrp::search::NeighbourhoodParams;
+
+NeighbourhoodParams::NeighbourhoodParams(double weightWaitTime,
+                                         double weightTimeWarp,
+                                         size_t numNeighbours,
+                                         bool symmetricProximity,
+                                         bool symmetricNeighbours)
+    : weightWaitTime(weightWaitTime),
+      weightTimeWarp(weightTimeWarp),
+      numNeighbours(numNeighbours),
+      symmetricProximity(symmetricProximity),
+      symmetricNeighbours(symmetricNeighbours)
+{
+    if (numNeighbours == 0)
+        throw std::invalid_argument("num_neighbours <= 0 not understood.");
+}
+
+std::vector<std::vector<size_t>> pyvrp::search::computeNeighbours(
+    [[maybe_unused]] pyvrp::ProblemData const &data,
+    [[maybe_unused]] NeighbourhoodParams const &params)
+{
+    // TODO
+    return {};
+}
--- a/pyvrp/cpp/search/neighbourhood.cpp
+++ b/pyvrp/cpp/search/neighbourhood.cpp
@@ -65,9 +65,8 @@ Matrix<double> computeProximity(ProblemData const &data)
                     = static_cast<double>(vehType.unitDurationCost)
                       * (dur + wait);
 
-                auto const timeWarp = iEarly + iService + dur - jLate;
-                if (timeWarp > 0)
-                    continue;
+                if (iEarly + iService + dur > jLate)  // then this arc is never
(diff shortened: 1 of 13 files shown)


pyvrp/cpp/search/neighbourhood.cpp
```diff
@@ -0,0 +1,146 @@
...
+            auto const jEarly = static_cast<double>(jData.twEarly);
+            auto const jLate = static_cast<double>(jData.twLate);
+
+            auto const minDur = static_cast<double>(minDuration);
+            auto const minWait = jEarly - minDur - iService - iLate;
+            auto const minTw = iEarly + iService + minDur - jLate;
+
+            prox(i, j) -= static_cast<double>(jData.prize);
+            prox(i, j) += params.weightWaitTime * std::max(minWait, 0.0);
+            prox(i, j) += params.weightTimeWarp * std::max(minTw, 0.0);
```
N-Wouda: Make wait time part of costs (per veh type, above), remove time warp factor. If the arc isn't even feasible we can just prune it.

leonlan:
| Commit  | Time     | CVRP-XXL | MDVRPTW | HFVRP | VRPB | MTVRPTWR | CVRP | PCVRPTW | VRPTW |
|---------|----------|----------|---------|-------|------|----------|------|---------|-------|
| dfbf81e main   | 1 min     |    5.65% |   2.97% | 1.87% | 1.50% |    1.22% | 0.80% |   1.52% | 2.90% |
| dfbf81e    | 2 min |    4.93% |   1.96% | 1.40% | 1.17% |    0.89% | 0.60% |   0.93% | 2.19% |
| dfbf81e | 10min |    3.85% |   1.30% | 0.81% | 0.85% |    0.35% | 0.40% |   0.59% | 1.44% |
| 4c67ae8 this PR | 1 min     |    5.68% |   2.98% | 1.88% | 1.50% |    1.21% | 0.81% |   1.49% | 2.98% |
| 4c67ae8 | 2 min|    4.98% |   2.00% | 1.41% | 1.17% |    0.88% | 0.61% |   0.97% | 2.25% |
| 4c67ae8 | 10 min |    3.86% |   1.31% | 0.81% | 0.85% |    0.37% | 0.41% |   0.66% | 1.42% |

tests/data/test_config.toml
```diff
@@ -13,10 +13,8 @@ history_length = 1
 
 [neighbourhood]
 weight_wait_time = 0
-weight_time_warp = 0
```
leonlan, line 16: For future reference: penalising wait time is useful in time window instances. We tried moving it into the duration cost instead, but since benchmark instances (and our defaults) don't use duration cost, this would make the neighbourhood less effective for users who don't specify it. Keeping it as an explicit parameter therefore ensures that we still get the desired effect.

pyvrp/search/neighbourhood.py
leonlan, line 1: Also for future reference (correct me if wrong): keeping this file because the docs don't render nicely otherwise

leonlan (approved):

tests/data/test_config.toml
leonlan: Including time warp was not useful so we removed this parameter.

pyvrp/search/neighbourhood.py
N-Wouda: Yeah, that's something we need to figure out. For some reason autodoc doesn't like object defaults when the object comes from C++.

(probably easy; but haven't had time to look into this)

PR #1061, opened 2026-03-04
Remove `Solution::neighbours`

N-Wouda:
An older holdout from our HGS days. It can go.

This PR:

- [ ] Closes #xxxx.

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/Solution.cpp
```diff
@@ -141,31 +122,21 @@ bool Solution::operator==(Solution const &other) const
...
-        if (client2vehType[route[0]] != route.vehicleType())
-            return false;
-
-    return true;
+    // Tests if the routes are permutations of each other. Quadratic (in the
+    // number of routes) in the worst case.
+    return std::is_permutation(routes_.begin(),
+                               routes_.end(),
+                               other.routes_.begin(),
+                               other.routes_.end());
```
N-Wouda, line 136: This is potentially quite slow, but that's why we do the attribute checks first.

And we don't use this ourselves in any critical place anyway.

leonlan (approved):

Issue #640, opened 2024-07-19
Rename Client to Task, abstract coordinates to Location


N-Wouda:
(a bit more open ended than the standard format)

Both our `Client` and `Depot` classes have $(x, y)$ coordinate data in them, and their indices are used to index the distance and duration matrices. That is slightly inconvenient, because it means users need to for example remember that depots are in the lower indices. Additionally, for clients that share a location, it means we need to duplicate the matrix entries. 

If we abstract the map data to a `Location` class, we can reference the locations from `VehicleType` and `Client`. This seems easy enough, and would remove the need for an explicit `Depot` object.

Further, I propose we rename `Client` to simply `Task`. All we care about is doing something at a location, and the `Client` name does not quite capture that. It's very.. academic.

N-Wouda:
I think we can still do this in 0.10.0; particularly the first part related to `Location`. Renaming `Client` to `Task` is not hard, it's just a boatload of work checking everything we've ever written.

N-Wouda:
[VROOM](https://github.com/VROOM-Project/vroom) uses `Job`, rather than `Task`. For transportation jobs (see #331) they use `Shipment`.

N-Wouda:
I think splitting `Location` (a place with coordinates and an entry in the distance and duration matrices) off from `Depot` and `Client` is very helpful, because it lets us duplicate depots and clients without also increasing the sizes of the distance and duration matrices. I don't think we need to rename depots and clients for now. Maybe in the future, but not immediately.

N-Wouda:
@PeterSR this has been asked before, see #650.

N-Wouda:
I think we can pick this up after 0.12. Since we're not dropping depots, I think we should just add a `Location` class storing coordinates (and an optional name?) with an associated entry into the dist/dur matrices. Depots and clients then need to be updated to take a location index. Pretty similar to how VROOM does this.

Not sure how much needs to be changed to make this happen. In any case, renaming `Client` to `Task` can happen after we do this location stuff.

N-Wouda:
@leonlan with ILS, the need for explicit coordinates kind of vanishes right? Assuming we can also remove `LS::intensify`?

leonlan:
> [@leonlan](https://github.com/leonlan) with ILS, the need for explicit coordinates kind of vanishes right? Assuming we can also remove `LS::intensify`?

Yes, assuming we remove `LS::intensify`.

N-Wouda:
[...]

PR #1069, opened 2026-03-10
Add location class

N-Wouda:
This PR:

- [x] Part of #640.

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/ProblemData.cpp
```diff
@@ -562,10 +596,14 @@ size_t ProblemData::numLoadDimensions() const { return numLoadDimensions_; }
 
 void ProblemData::validate() const
 {
+    if (numLocations() != numDepots() + numClients())  // TODO temp
+        throw std::invalid_argument(
+            "Need #locations to match #depots + #clients.");
```
N-Wouda, line 601: Temporarily need this because we still rely on `numLocations()` as the upper bound for the client indices. That'll go, but this PR is already pretty big.

PR #1072, opened 2026-03-10
Remove x/y coords from depots and clients

N-Wouda:
This PR:

- [ ] Closes #xxxx.

<details>

**Notes**:

In particular:

</details>


pyvrp/Model.py
leonlan, line 1: Do we also need to check that `model.add_edge` only adds locations that exist in this instance? We didn't do it before, but I just thought about it with the new location checks.

leonlan (approved):

pyvrp/Model.py
N-Wouda: Would be nicer but that adds another (linear) lookup to a fairly expensive loop - adding all edges is already quadratic. Adding an edge with unknown location goes wrong when calling `Model.data()`, since there the locations are turned into indices. So we already raise at some point, just a little later than where we add the broken edge.

We could add state to the model to track e.g. `obj2idx` explicitly, but that is a lot of extra work for IMHO limited gain. I think we can leave this as-is.

PR #1075, opened 2026-03-11
Use Activity in pyvrp.search

N-Wouda:
This PR:

- [ ] Closes #xxxx.

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/search/ReplaceOptional.cpp
```diff
@@ -14,16 +14,19 @@ std::pair<pyvrp::Cost, bool> ReplaceOptional::evaluate(
     if (U->route() || !V->route() || V->isDepot())
         return std::make_pair(0, false);
 
-    auto const &uData = data.client(U->client() - data.numDepots());
-    auto const &vData = data.client(V->client() - data.numDepots());
+    auto const [uType, uClient] = U->activity();
+    auto const [vType, vClient] = V->activity();
```
leonlan, line 18:
```suggestion
    auto const [_, uClient] = U->activity();
    auto const [_, vClient] = V->activity();
```

pyvrp/cli.py
```diff
@@ -59,11 +73,9 @@ def write_solution(where: Path, data: ProblemData, result: Result):
 
         routes = [f"Route #{idx + 1}:" for idx in range(data.num_vehicles)]
         for route in result.best.routes():
-            visits = [str(visit.location) for visit in route.schedule()]
-            visits = visits[1:-1]  # skip start and end depots
-
+            visits = route2vrplib(visits)
```
leonlan:
```suggestion
            visits = route2vrplib(route)
```

I think

tests/search/test_LocalSearch.py
```diff
@@ -406,8 +399,8 @@ def test_local_search_removes_useless_reload_depots(ok_small_multiple_trips):
     # The local search should have removed the reload depot from the first
     # route, because that was not providing any value.
     routes = improved.routes()
-    assert_(str(routes[0]), "1 3")
-    assert_(str(routes[1]), "2 4")
+    assert_(str(routes[0]), "0 2")
+    assert_(str(routes[1]), "1 3")
```
leonlan:
```suggestion
    assert_equal(str(routes[0]), "C0 C2")
    assert_equal(str(routes[1]), "C1 C3")
```

tests/search/test_RelocateWithDepot.py
```diff
@@ -367,7 +372,7 @@ def test_depot_service_duration(ok_small_multiple_trips):
 
     # Single route, no reload depots. This route has excess load, which the
     # cost evaluator penalises heavily, but is feasible w.r.t. duration.
-    route = make_search_route(data, [3, 2, 4, 1])
+    route = make_search_route(data, ["C2", "C1", "C3", "C0"])
     assert_(route.excess_load(), [8])
```
leonlan:
```suggestion
    assert_equal(route.excess_load(), [8])
```
(also the assert below should become `assert_equal`)

tests/search/test_Exchange.py
```diff
@@ -296,25 +292,25 @@ def test_relocate_to_heterogeneous_empty_route(ok_small):
...
+    neighbours[1].append(0)
 
     ls = LocalSearch(data, rng, neighbours)
     ls.add_operator(Exchange10(data))
 
-    # The initial solution has routes with loads [13, 5, 0, 0]
-    # with excess [1, 0, 0, 0]. Moving node 3 to route 4 will resolve all
-    # load penalties, but other moves would increase load penalties.
-    # Therefore, this requires moving to an empty route which is not the first.
+    # Thi solution has routes with loads [13, 5, 0, 0] and excess [1, 0, 0, 0].
```
leonlan:
```suggestion
    # This solution has routes with loads [13, 5, 0, 0] and excess [1, 0, 0, 0].
```

tests/search/test_Exchange.py
```diff
@@ -296,25 +292,25 @@ def test_relocate_to_heterogeneous_empty_route(ok_small):
...
+    # Moving C2 to C4 will resolve all load penalties, but other moves would
+    # increase load penalties. Therefore, this requires moving to an empty
+    # route which is not the first.
```
leonlan:
```suggestion
    # Moving C2 to route 4 will resolve all load penalties, but other moves would
    # increase load penalties. Therefore, this requires moving to an empty
    # route which is not the first.
```

pyvrp/cpp/Route.cpp
```diff
@@ -109,47 +49,32 @@ Duration Route::ScheduledVisit::waitDuration() const { return waitDuration_; }
...
+    schedule_.reserve(activities_.size());
 
-    if (trips_.size() > vehData.maxTrips())
-        throw std::invalid_argument("Vehicle cannot perform this many trips.");
+    auto const &vehData = data.vehicleType(vehicleType_);
+    auto const &durations = data.durationMatrix(vehData.profile);
 
-    if (trips_[0].startDepot() != startDepot_)
+    Duration releaseTime;
+    std::vector<Duration> releaseTimes;
```
leonlan:
```suggestion
    std::vector<Duration> releaseTimes; // per trip
```

Small comment would be helpful here (intuitively I'd think its tracked per activity).

pyvrp/read.py
```diff
@@ -332,7 +322,9 @@ def mutually_exclusive_groups(self) -> list[list[int]]:
             # cast it to a 2D array.
             groups = np.atleast_2d(groups).T
 
-        raw_groups = [[idx - 1 for idx in group] for group in groups]
+        raw_groups = [
+            [idx - self.num_depots - 1 for idx in group] for group in groups
+        ]
```
N-Wouda, line 327: I think this used to be wrong before but we got away with it because we don't have GVRP instances with more than one depot.

Now maps indices to [0, 1, ...], as it should. Previously mapped to [1, 2, ...] (and assumed only a single depot).

pyvrp/cpp/Route.h
leonlan:
```suggestion
    * activity : Activity
    *     Route activity.
```

pyvrp/cpp/Route.cpp
```diff
@@ -347,43 +330,41 @@ Route::Route(Trips trips,
...
-
-size_t Route::size() const
+bool Route::empty() const
 {
-    return std::accumulate(trips_.begin(),
-                           trips_.end(),
-                           0,
-                           [](size_t count, auto const &trip)
-                           { return count + trip.size(); });
+    return size() == 2;  // only start and end depot activites
```
leonlan:
```suggestion
    return size() == 2;  // only start and end depot activities
```

tests/search/test_SearchSpace.py
```diff
@@ -7,7 +7,7 @@
 from tests.helpers import make_search_route
 
 
-@pytest.mark.parametrize("size", [1, 2, 3, 4, 6, 7])  # num_clients + 1 == 5
+@pytest.mark.parametrize("size", [1, 2, 3, 5, 6, 7])  # num_clients == 5
```
leonlan:
```suggestion
@pytest.mark.parametrize("size", [1, 2, 3, 5, 6, 7])  # num_clients == 4
```

pyvrp/cpp/search/ReplaceOptional.cpp
N-Wouda: Can't be done, redefinition of symbol

leonlan (approved):

PR #1084, opened 2026-03-16
Make Activity a class, rather than struct

N-Wouda:
This PR:

- [ ] Closes #xxxx.

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/bindings.cpp
```diff
@@ -63,16 +63,13 @@ PYBIND11_MODULE(_pyvrp, m)
              py::arg("idx"))
         .def(py::init<std::string const &>(), py::arg("description"))
         .def(py::self == py::self, py::arg("other"))  // this is __eq__
-        .def("__iter__",
-             [](Activity const &activity)
-             { return py::iter(py::make_tuple(activity.type, activity.idx)); })
```
N-Wouda, line 68: The class no longer supports unpacking (that's now private data), and we don't really use it ourselves anywhere else. It also doesn't make sense for `ScheduledActivity`, since there you'd probably want to use the schedule data.

Best to just drop this, I think.

leonlan (approved):

PR #1087, opened 2026-03-21
Add duration cost as a PWL function

### LINKED ISSUE #1044: Piecewise linear duration costs

FormelessPuppy41:
This PR:

- [ ] Closes #1044.

**Summary**

Implements piecewise linear (PWL) duration costs (#1044):

- Adds a `PiecewiseLinearFunction<Duration, Cost> durationCost` field to `VehicleType`, alongside a precomputed `hasDurationCost` flag. The existing `unitDurationCost` and `unitOvertimeCost` scalars are kept; replacing them with PWL evaluation in the cost computation and neighbourhood is left for a follow-up PR.
- Adds `operator<=>` to `PiecewiseLinearFunction` to support ordering comparisons.
- Validates that `durationCost` is monotonically increasing on construction.
- Exposes `duration_cost` in the Python bindings as a constructor argument, readonly property, and `replace()` parameter, including pickle support.
- Adds `duration_cost` to `Model.add_vehicle_type()`.

<details>

**Notes**:

In particular:

</details>


FormelessPuppy41:
Hi @N-Wouda,
I have implemented the piecewise linear duration cost function. I chose not to remove the unit costs yet, as I think that warrants a separate PR.
Code coverage isn't complete yet, however, before trying to improve the tests, I would like your review of the `VehicleType.durationCost` implementation first.

pyvrp/cpp/ProblemData.h
```diff
@@ -536,7 +543,11 @@ class ProblemData
         Duration const maxOvertime;              // Maximum allowed overtime
         Cost const unitOvertimeCost;             // Cost per unit of overtime
         Duration const maxDuration;  // Maximum route duration, incl. overtime
-        char const *name;            // Type name (for reference)
+        PiecewiseLinearFunction<int64_t, int64_t> const
+            durationCost;  // Cost f(duration)
```
N-Wouda:
```suggestion
        PiecewiseLinearFunction<Duration, Cost> const
            durationCost;
```

N-Wouda: Maybe alias this inside `VehicleType`, and use the alias. e.g.
```cpp
struct VehicleType
{
    using DurationCost = PiecewiseLinearFunction<Duration, Cost>;
```

pyvrp/cpp/ProblemData.h
```diff
@@ -557,6 +568,9 @@ class ProblemData
                     size_t maxReloads = std::numeric_limits<size_t>::max(),
                     Duration maxOvertime = 0,
                     Cost unitOvertimeCost = 0,
+                    std::optional<PiecewiseLinearFunction<int64_t, int64_t>>
+                        durationCost
+                    = std::nullopt,
```
N-Wouda: Why optional? What happens now when users pass both unit costs and a duration cost function? What happens when they pass nothing?

pyvrp/cpp/ProblemData.cpp
```diff
@@ -381,6 +395,17 @@ ProblemData::VehicleType::VehicleType(size_t numAvailable,
...
+    // NOTE: @N-Wouda:
+    // Currently, we use the monotonicity of the duration cost function,
+    // however, some use cases might require non-negativity instead. For
+    // example, if a user prefers medium duration routes over short and long
+    // routes, they might want a duration cost function that has a minimum at
+    // some point greater than 0. What is your preference? Do you prefer to
+    // enforce monotonicity, or non-negativity?
+    if (!this->durationCost.isMonotonicallyIncreasing())
+        throw std::invalid_argument("duration_cost must be monotonically "
+                                    "increasing.");
```
N-Wouda: This does not need to be monotonically increasing, I think, but we do need non-negativity. Else a lot of the evaluation shortcuts break, and besides, what would negative costs mean in practice?

pyvrp/cpp/ProblemData.h
FormelessPuppy41: Currently, unit costs default to 0 when not provided. I wanted the same pattern for duration_cost: omitting it results in a zero-cost function as defined in `ProblemData.cpp:344`.
I kept the unit costs for now to keep the diff small and focused; if the current duration cost implementation is accepted then both unit costs will be removed. So, users will not be able to provide both unit costs and a duration cost function.

pyvrp/cpp/ProblemData.h
FormelessPuppy41: Using `PiecewiseLinearFunction<Duration, Cost>` would cause compilation errors inside the `PiecewiseLinearFunction` constructor. The arithmetic there computes `slope = dy / dx` (`Co / Dom`) and `dy % dx` (`Co % Dom`), but `Measure`'s `operator/` and `operator%` only accept matching types so `Cost / Duration` won't compile. 
Additionally, `Segment = std::pair<Dom, Dom>` (in `PiecewiseLinearFunction.h:54`) stores intercepts and slopes as `Duration` values, however, those should be `Cost` (`Co`) right?

I think there are two ways to fix the compilation errors;
1. Cast to `int64_t` inside `PiecewiseLinearFunction.h`.
2. Add cross-type operators to `Measure.h`.

However, even with either of these fixes, changing the alias to `<Duration, Cost>` would require updating the Python bindings because the registered Python` PiecewiseLinearFunction` class is` <int64_t, int64_t>`, and pybind11 can't implicitly convert between the two instantiations. The `duration_cost` property and the pickle code would both need explicit conversion lambdas.

For now, I have implemented the alias as:
```cpp
using DurationCost = PiecewiseLinearFunction<int64_t, int64_t>;
```
which hides the verbosity without touching `PiecewiseLinearFunction.h` or the bindings. 

What's your preference on the fix?

pyvrp/cpp/ProblemData.h
FormelessPuppy41: I have removed the unnecessary `optional`. The parameter now defaults directly to the zero-cost function inside `ProblemData.h:572` as 
```cpp
DurationCost durationCost = DurationCost(
                        {}, {DurationCost::Segment{0, 0}}),
```

FormelessPuppy41:
Hi @N-Wouda, I have implemented the suggested changes, except for changing the alias to `PiecewiseLinearFunction<Duration, Cost>`, the reasons are outlined in my comment above. Could you share your preference on how to proceed, and i will implement it from there (and I will resolve the `codecov` fail)?

pyvrp/cpp/ProblemData.h
N-Wouda: > Additionally, `Segment = std::pair<Dom, Dom>` (in `PiecewiseLinearFunction.h:54`) stores intercepts and slopes as `Duration` values, however, those should be `Cost` (`Co`) right?

Probably, yeah. That should resolve quite a bit of casting issues.

> I think there are two ways to fix the compilation errors;
> 
> 1. Cast to `int64_t` inside `PiecewiseLinearFunction.h`.
> 2. Add cross-type operators to `Measure.h`.

Or update the segment type, cast the input domain to the codomain, and then you're basically done.

> The `duration_cost` property and the pickle code would both need explicit conversion lambdas.

They need explicit *types*, i.e., `PiecewiseLinearFunction<Duration, Cost>`, not the general alias of `PiecewiseLinearFunction = PiecewiseLinearFunction<int64_t, int64_t>` - I think that alias can go (it's only adding confusion). Use explicit types in the bindings.

pybind already knows how to convert Python `int` to `Duration` and `Cost` via the custom type casters. My hope is that this'll work without any issues on the Python side.

Else we might need a custom type caster, like for the matrices/measure types. Let's first see what pybind11 makes of the explicit type.

N-Wouda: I hope we can avoid a custom type caster, but those are also not too hard to write. 🤞

FormelessPuppy41: I implemented the segments as `Segment = std::pair<Co, Co>`. I removed the general alias of `PiecewiseLinearFunction = PiecewiseLinearFunction<int64_t, int64_t>` and renamed the generic `PiecewiseLinearFunction` name to `DurationCost` inside the `bindings.cpp` file for explicit typing.

FormelessPuppy41:
Hi @N-Wouda, I have implemented all the suggested changes. The code now passes all CI checks including the codecov requirements. If you are happy with the current approach then I will move on to phasing out the unit costs in favour of `duration_cost`.

N-Wouda:
Hi, I'm sorry, but this now only works for `PiecewiseLinearFunction<Duration, Cost>`. There's no way to pass a general PWL function to anything else. Maybe something like
```cpp
    template <typename OtherDom, typename OtherCo>
        requires(std::convertible_to<OtherDom, Dom>
                 && std::convertible_to<OtherCo, Co>)
    PiecewiseLinearFunction &
    operator=(PiecewiseLinearFunction<OtherDom, OtherCo> const &other);
```
could work to convert between different PWL types, but I don't have time right now to explore further.

I don't feel like there's any progress being made in this PR (or, to be honest, in the previous PRs - they all took significant revision to end up in a reasonable state). Adding this in will probably take a few hours of my time; so far, the PRs have taken a multiple of that, with boilerplate LLM-generated code getting in the way at every turn. I'm pretty tired of that.

I suggest you implement whatever more you need in your own fork, and use that fork instead. We'll get around to using PWL functions in due time, but it won't be via this PR.

FormelessPuppy41:
Hi @N-Wouda , 
That is completely understandable. I will go ahead and implement what I need in my own fork.
Thank you for your time, reviews, and feedback!

N-Wouda:
Good luck!

PR #1088, opened 2026-03-23
feat(build): add free-threaded Python 3.14t support

### LINKED ISSUE #1053: Check and support free-threaded Python
I think we need to update [this](https://pybind11.readthedocs.io/en/stable/advanced/misc.html#free-threading-support), and then check everything works well.

Might require more, need to check.

SkyTik:
This PR:

- [x] Closes #1053.

## Summary

Add free-threaded Python (3.14t) support:

- Mark pybind11 extension modules (`_pyvrp`, `_search`) as GIL-free via `py::mod_gil_not_used()`
- Add CI job running full test suite with `PYTHON_GIL=0` on Python 3.14t
- Enable `cpython-freethreading` in cibuildwheel and build `cp314t-*` wheels
- Add thread-safety warnings to `PenaltyManager` and `RandomNumberGenerator`
- Minor code cleanups: simplify comparisons, use `std::is_integral_v`

<details>

**Notes**:

In particular:

</details>

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -115,34 +115,26 @@ jobs:
           uv run env PYTHONMALLOC=malloc valgrind --tool=memcheck --quiet $VALGRIND_CMD
 
   free-threaded:
-    name: Test with free-threaded Python 3.13
+    name: Test with free-threaded Python 3.14
     runs-on: ubuntu-24.04
     env:
       GCC_VERSION: "14"
-      # Point uv at the free-threaded interpreter installed by setup-python,
-      # so it does not fall back to the system /usr/bin/python3 (3.12).
-      UV_PYTHON: python3.13t
+      UV_PYTHON: python3.14t
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           persist-credentials: false
       - name: Set up free-threaded Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
-          python-version: "3.13t"
+          python-version: "3.14t"
       - name: Install uv
-        uses: astral-sh/setup-uv@v6
+        uses: astral-sh/setup-uv@v7
         with:
           version: "0.7.5"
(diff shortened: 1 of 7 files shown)


N-Wouda:
Hi @SkyTik, thanks - but can you follow our PR template? I.e., waht this closes, etc. I'm not sure why you removed it.

Regarding the PR: I think we should target 3.14t, not 3.13t. That should also solve the lxml and cffi dependency issues, since both build for 3.14t.

N-Wouda:
I read [this](https://pybind11.readthedocs.io/en/stable/advanced/misc.html#free-threading-support) just yet to get a better understanding of what free-threading takes. Particularly this bit is worrisome:

> Importantly, enabling your module to be used with free-threading is also your promise that your code is thread safe.

I don't think our native extensions are thread safe to a sufficient degree yet. Marking them as so with `py::mod_gil_not_used()` isn't a fix. The only way to comply with this requirement is to have synchronisation primitives, and there might be quite a few places that need these.

I do not think I can merge this in its current form, and I'm unsure if adding all the required locking will prove worthwhile. But that's an investigation that needs to happen in any case before we can mark everything as GIL-free.

SkyTik:
hey guys, sorry if there are any in-convenient, i like this repository and would like to make some contribution via Agentic Coding, if you find this is weird just tell me to stop. Thanks

N-Wouda:
Hi @SkyTik we're open to contributions, but they do need to work. With this I'm not sure it does, especially since we have no multithreaded tests yet that could fail when the GIL's not present. Removing the GIL requirement should probably begin with adding those. In any case, adding free-threading support is a bigger PR than just changing two lines in our bindings.

In the future I suggest you open or leave a comment on an issue before opening a PR. That helps scope things first.

SkyTik:
i see, i close this PR and start with the issue first, thanks  @N-Wouda

N-Wouda:
No problem, thanks anyway for the attempt @SkyTik!

PR #1103, opened 2026-05-05
Track unplanned clients

### LINKED ISSUE #1102: Track unplanned clients/shipments on Solution

Currently, there is no way to know which clients and/or shipments (#331) are unplanned in a solution without writing code to find that out. Knowing what's unplanned is a fairly basic need, so this is rather tedious to do.

It's much more ergonomical if we can provide a `unplanned: list[Activity]` that tracks unplanned clients and shipments.

N-Wouda:
This PR:

- [x] Closes #1102.

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/Solution.cpp
leonlan, line 1: Maybe add a `num_unplanned` property?

leonlan (approved):

pyvrp/cpp/Solution.cpp
N-Wouda: Not sure yet how to define that in combination with shipments (are those one or two unplanned entities? Is `num_unplanned` always `len(unplanned)`?

Maybe leave this for later when we have a bit more clarity.

PR #1098, opened 2026-04-24
Pickup and delivery

N-Wouda:
Stacked PR for #331. This PR eventually:

- [x] Closes #331.
- [x] Adds and passes relevant tests.
- [x] Has well-formatted code and documentation.

To update:
- [x] `ProblemData` and `Model`
- [x] `Route` and `Solution`
- [x] Update `Solution.make_random`
- [x] Plotting tools
- [x] Updates `pyvrp::search::Solution`
- [x] Optional shipment operators
- [x] `SearchSpace` and `neighbourhood`
- [x] `PerturbationManager`
- [x] Structural feasibility
- [x] Update `Exchange` and `SwapTails` to work with shipments
- [x] `LocalSearch`
- [x] New operators (`RelocateShipment`, `SwapShipment`)
     - [x] Add `RelocateShipment` move that relocates a shipment pair to somewhere else.
     - [x] Add a `SwapShipment` move that exchanges two shipments.
     - [x] Add `RelocatePickup` move that relocates the pickup node to a better place in its current route.
- [x] Add notebooks
- [x] Relevant documentation

Included:
- #1112
- #1113
- #1114
- #1119
- #1120 
- #1121
- #1122
- #1125
- #1130
- #1134
- #1135
- #1137
- #1138
- #1139
- #1140
- #1143
- #1145
- #1146
- #1148
- #1153 
- #1155
- #1156
- #1157
- #1159
- #1161
- #1162
- #1164
- #1165
- 

---

### 1-minute gaps

| Benchmark (# instances) | v0.13.4 | be31b1359 |
|---|---:|---:|
| CVRP (100) | 0.65% | 0.67% |
| VRPTW (60) | 2.28% | 2.31% |
| PCVRPTW (60) | 1.01% | 0.88% |
| MDVRPTW (28) | 1.85% | 2.69% |
| VRPB (90) | 1.19% | 1.47% |
| HFVRP (100) | 1.39% | 1.42% |
| MTVRPTWR (171) | 0.95% | 1.39% |
| Li1000 PDPTW (58) | - | 6.38% |
| GVRP (56) | 1.36% | 1.11% |

### 10-minute gaps

| Benchmark (# instances) | v0.13.4 | be31b1359 |
|---|---:|---:|
| CVRP (100) | 0.32% | 0.37% |
| VRPTW (60) | 1.12% | 1.10% |
| PCVRPTW (60) | 0.66% | 0.47% |
| MDVRPTW (28) | 0.95% | 1.33% |
[...]

PR #1112, opened 2026-06-03
Add Shipment class

N-Wouda:
Part of the stacked PR in #1098.

<details>

**Notes**:

In particular:

</details>


pyvrp/cpp/Shipment.cpp
```diff
@@ -0,0 +1,123 @@
...
+               deliveryTwLate,
+               deliveryServiceDuration),
+      amount(amount),
+      prize(prize),
+      required(required),
+      name(duplicate(name.data()))
+{
+    if (std::any_of(amount.begin(), amount.end(), isNegative<Load>))
+        throw std::invalid_argument("shipment amounts must be >= 0.");
+
```
leonlan, line 74: Should we also check here that the step TWs make sense?

If `deliveryTwLate < pickupTwEarly` then we know this shipment can't be feasible.

leonlan (approved):

pyvrp/cpp/Shipment.cpp
N-Wouda: Yeah think so, forgot about that check. I'll add this.

PR #1131, opened 2026-07-13
Add RelocateAlternative

leonlan:
This PR adds RelocateAlternative. This is useful for VRPs with mutually exclusive groups, and covers a part of the search space that wasn't covered before. We noticed it when benchmarking #1073; our benchmark results show a fairly big decrease in average gap:

| Experiment | Mean gap |
|---|---:|
| Main | 3.135% |
| RelocateAlternative | 2.051% |

The same holds for GVRP.

| Experiment | Mean gap |
|---|---:|
| Main | 1.064% |
| RelocateAlternative | 0.715% |


benchmarks/conftest.py
leonlan, line 1: Adding this benchmark covers two blind spots: mutually exclusive groups and TSP.

pyvrp/cpp/search/RelocateWithAlternative.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+std::string RelocateWithAlternative::name() const
+{
+    return "RelocateWithAlternative";
+}
+
+template <>
+bool pyvrp::search::supports<RelocateWithAlternative>(ProblemData const &data)
+{
+    return data.numGroups() > 0;
```
N-Wouda: This should check whether any of the groups are mutually exclusive. We do not currently have any other type of group, but I did plan at some point for that to change.

Something like
```cpp
for (auto const &group : data.groups())
  if (group.mutuallyExclusive)
    return true;

return false;
```

pyvrp/cpp/search/RelocateWithAlternative.cpp
```diff
@@ -0,0 +1,107 @@
...
+    auto const *vRoute = V->route();
+    auto const &uData = data.client(U->idx());
+    if (!uRoute || !vRoute || !uData.group || U == V || U == n(V))
+        return std::make_pair(0, false);
+
+    if (uRoute == vRoute && U->trip() != V->trip())
+        return std::make_pair(0, false);
+
+    auto const &group = data.group(*uData.group);
+    assert(group.mutuallyExclusive);
```
N-Wouda: I think we're wrong in asserting this if we ever change the group system, but we do so in other places too, so should be OK for now.

pyvrp/cpp/search/RelocateWithAlternative.cpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+        else
+        {
+            costEvaluator.deltaCost(
+                deltaCost,
+                Route::Proposal(uRoute->before(V->pos()),
+                                ClientSegment(data, client),
+                                uRoute->between(V->pos() + 1, U->pos() - 1),
+                                uRoute->after(U->pos() + 1)));
+        }
```
N-Wouda: Move if/else outside the loop, it does not depend on anything looping related

pyvrp/cpp/search/RelocateWithAlternative.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+        if (deltaCost < bestCost)
+        {
+            bestCost = deltaCost;
+            alternative_ = alternative;
+        }
```
N-Wouda: Do we need to do best improvement? This now exhausts groups for each evaluation, which can be expensive for big groups. Maybe first improvement?

pyvrp/cpp/search/RelocateWithAlternative.cpp
N-Wouda: then loop inside each branch (maybe separate member functions if necessary, this is all rather big already)

pyvrp/cpp/search/RelocateWithAlternative.h
```diff
@@ -0,0 +1,36 @@
...
+
+namespace pyvrp::search
+{
+/**
+ * RelocateWithAlternative(data: ProblemData)
+ *
+ * Evaluates replacing the current mutually exclusive group member :math:`U` by
+ * another member of the same group, while relocating it after :math:`V`.
+ */
+class RelocateWithAlternative : public BinaryOperator
```
N-Wouda: Maybe rename to `RelocateAlternative`.

tests/search/test_RelocateAlternative.py
```diff
@@ -0,0 +1,81 @@
...
+    # The old route costs 1544 + 1593 + 1475 = 4612. The proposed route
+    # C3 -> C2 costs 1476 + 828 + 2063 = 4367, for a delta of -245.
+    move = op.evaluate(sol.nodes[0], sol.nodes[3], cost_eval)
+    assert_equal(move, (-245, True))
+
+    op.apply(sol.nodes[0], sol.nodes[3])
+    assert_equal(str(sol.routes[0]), "C3 C2")
+
+
+def test_accounts_for_prizes_and_fixed_vehicle_costs():
```
N-Wouda, line 74: Can we also add a test where V is the start depot of an empty route (for that route's fixed cost component?)

pyvrp/cpp/search/RelocateWithAlternative.cpp
leonlan: I also changed this for ReplaceGroup

pyvrp/cpp/search/RelocateAlternative.cpp
```diff
@@ -0,0 +1,154 @@
...
+            costEvaluator.deltaCost(
+                deltaCost,
+                Route::Proposal(route->before(U->pos() - 1),
+                                route->between(U->pos() + 1, V->pos()),
+                                ClientSegment(data, client),
+                                route->after(V->pos() + 1)));
+
+            move_ = {deltaCost, alternative};
+            if (deltaCost < 0)
+                return;
```
N-Wouda:
```suggestion
            
            if (deltaCost < 0)
            {
                move_ = {deltaCost, alternative};
                return;
            }
```

pyvrp/cpp/search/RelocateAlternative.cpp
```diff
@@ -0,0 +1,154 @@
...
+void RelocateAlternative::evalWithinRoute(Route::Node *U,
+                                          Route::Node *V,
+                                          ClientGroup const &group,
+                                          CostEvaluator const &costEvaluator)
+{
+    auto const *route = U->route();
+    auto const &uData = data.client(U->idx());
+
+    if (U->pos() < V->pos())
+    {
```
N-Wouda: These brackets (for if and else) are unnecessary

pyvrp/cpp/search/RelocateAlternative.cpp
```diff
@@ -0,0 +1,154 @@
...
+    auto const *vRoute = V->route();
+    auto const &uData = data.client(U->idx());
+    if (!uRoute || !vRoute || !uData.group || U == V || U == n(V))
+        return std::make_pair(0, false);
+
+    if (uRoute == vRoute && U->trip() != V->trip())
+        return std::make_pair(0, false);
+
+    auto const &group = data.group(*uData.group);
+    assert(group.mutuallyExclusive);
```
N-Wouda: Do we still need this here? I think we can just get the group inside the later eval, rather than passing it as an argument.

pyvrp/cpp/search/RelocateAlternative.cpp
```diff
@@ -0,0 +1,154 @@
...
+    move_ = {};
+
+    auto const *uRoute = U->route();
+    auto const *vRoute = V->route();
+    auto const &uData = data.client(U->idx());
+    if (!uRoute || !vRoute || !uData.group || U == V || U == n(V))
+        return std::make_pair(0, false);
+
+    if (uRoute == vRoute && U->trip() != V->trip())
+        return std::make_pair(0, false);
```
N-Wouda: This can move into `evalWithinRoute`, I think.

N-Wouda (approved):
LGTM if perf is also OK

PR #1141, opened 2026-07-27
Use `std::stoul` in Activity constructor

leonlan:
Ran into this error with clang21.

```
../pyvrp/cpp/Activity.cpp:48:44: error: null passed to a callee that requires a non-null argument [-Werror,-Wnonnull]
   48 |       idx_(std::stol(description.empty() ? 0 : description.substr(1)))
      |               
```

### DIFF
--- a/pyvrp/cpp/Activity.cpp
+++ b/pyvrp/cpp/Activity.cpp
@@ -37,7 +37,7 @@ Activity::Activity(ActivityType type, size_t idx) : type_(type), idx_(idx) {}
 
 Activity::Activity(std::string const &description)
     : type_(char2type(description.empty() ? ' ' : description[0])),
-      idx_(std::stol(description.empty() ? "0" : description.substr(1)))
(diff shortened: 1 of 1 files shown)


pyvrp/cpp/Activity.cpp
```diff
@@ -45,7 +45,7 @@ Activity::Activity(ActivityType type, size_t idx) : type_(type), idx_(idx) {}
 
 Activity::Activity(std::string const &description)
     : type_(char2type(description.empty() ? ' ' : description[0])),
-      idx_(std::stol(description.empty() ? 0 : description.substr(1)))
+      idx_(std::stol(description.empty() ? "0" : description.substr(1)))
```
N-Wouda: Looking at this again, should this not be `std::stoul` or the like? `idx_` is unsigned.

I suppose something like this should work:
```suggestion
      idx_(description.empty() ? 0 : std::stoul(description.substr(1)))
```

PR #1047, opened 2026-02-19
Initialise from penalty midpoint

leonlan:
This PR:
- [x] Removes `init_from` to initialise from the penalty midpoint directly using `PenaltyParams.midpoint_penalties(data)`. This reduces memory significantly (see #958), similar performance, and is just generally simpler esp. with difficult heterogeneous instances.
- [x] Also fixes CI issues identified in #1046 
- [x] Adds and passes relevant tests.
- [ ] Has well-formatted code and documentation.

Benchmark comparisons v0.13.0 - changes branched off from v0.13.0, as well.

| Commit  | Time     | CVRP-XXL | MDVRPTW | HFVRP | VRPB | MTVRPTWR | CVRP | PCVRPTW | VRPTW |
|---------|----------|----------|---------|-------|------|----------|------|---------|-------|
| v0.13.0    | 1min |    5.45% |   2.47% | 1.73% | 1.39% |    0.86% | 0.76% |   0.90% | 2.30% |
| v0.13.0    | 2min |    4.60% |   1.66% | 1.18% | 1.03% |    0.66% | 0.53% |   0.69% | 1.64% |
| v0.13.0    | 10min |    3.05% |   1.14% | 0.64% | 0.71% |    0.33% | 0.35% |   0.48% | 1.03% |
| f5d0978 | midpoint-1min |    4.89% |   2.91% | 1.48% | 1.36% |    1.05% | 0.81% |   1.07% | 2.45% |
| f5d0978 | midpoint-2min |    4.25% |   1.74% | 0.97% | 1.01% |    0.76% | 0.56% |   0.75% | 1.81% |
| f5d0978 | midpoint-10min |    3.12% |   1.12% | 0.55% | 0.73% |    0.33% | 0.37% |   0.51% | 1.10% |

<details>

**Notes**:

Please read our [contributing guidelines](https://pyvrp.org/dev/contributing.html) first.
In particular:

- You must add tests when making code changes.
  This keeps the code coverage level up, and helps ensure the changes work as intended.
- When fixing a bug, you must add a test that would produce the bug in the master branch, and then show that it is fixed with the new code. 
- New code additions must be well formatted. Changes should pass the pre-commit workflow, which you can set up locally using [pre-commit](https://pre-commit.com/#intro). 
- Docstring additions must render correctly, including escapes and LaTeX.
- Finally, it is essential that all contributions in this PR are license-compatible with PyVRP's MIT license.
  Please check that this PR can be included into PyVRP under the MIT license.

</details>

PR #1171, opened 2026-08-12
Do not emit PenaltyWarnings when violations are decreasing

### LINKED ISSUE #930: Improve initial penalties
Penalty parameters currently don't let users configure initial penalties. This can be useful when the initial penalties calculated with `PenaltyManager.init_from` are too low and users need quick feasible solutions. The only option now is to set `min_penalty`, but then we can't have infeasible solutions.

`PenaltyManager.initial_penalties` should be configurable. We have to think about how this works together with `PenaltyManager.init_from`.

leonlan:
This PR changes `PenaltyManager` to only emit a `PenaltyBoundWarning` when violations have stopped decreasing. The motivation comes from #1047: penalties start high, and with random initial solutions they will almost certainly keep increasing during the first few thousand iterations. This hits the penalty bound early and emits warnings that mean nothing at that moment.

With this change, a warning is only emitted once the violations no longer decrease between updates. The search can improve random initial solutions toward feasibility without spurious warnings.

I ran some additional tests:
- Our regular instances in PyVRP/Instances. Main emits warnings on several HFVRP instances; this branch no longer does.
- VRPTW instances with the fleet fixed at the BKS number of vehicles. Main emits warnings in most cases; this branch no longer does (and finds feasible solutions at some point).
- Randomly generated infeasible instances. Warnings are still emitted, as expected.
- FastVRP instances with very large data scaling. Warnings are still emitted, as expected.

Furthermore,
- [x] Closes #930.

<details>

**Notes**:

In particular:

</details>


pyvrp/PenaltyManager.py
```diff
@@ -191,46 +191,57 @@ def _compute(self, penalty: float, feas_percentage: float) -> float:
...
-        avg = fmean(feas_list)
-        feas_list.clear()
-        return self._compute(penalty, avg)
+        self._prev_avg_violations[idx] = avg_violation
+        self._penalties[idx] = self._compute(penalty, feas_percentage)
 
     def register(self, sol: Solution):
         """
-        Registers the feasibility dimensions of the given solution.
+        Registers the violations for each dimension of the given solution.
```
N-Wouda:
```suggestion
        Registers the violations for each penalty dimension of the given solution.
```
we use the same terminology above, when explaning `_viol_lists`.

tests/test_PenaltyManager.py
```diff
@@ -346,19 +346,46 @@ def test_max_min_penalty(ok_small):
...
+    solutions = [
+        Solution(ok_small, [[1, 2, 0, 3]]),  # excess load 8
+        Solution(ok_small, [[0, 1, 3]]),  # excess load 5
+        Solution(ok_small, [[1, 2, 0]]),  # excess load 3
+    ]
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error", PenaltyBoundWarning)
+        for sol in solutions:
+            pm.register(sol)
```
N-Wouda: I think this reads a bit easier with the `recwarn` fixture. E.g.

```python
for sol in solutions:
    pm.register(sol)

# No warnings should have been raised.
assert_equal(len(recwarn), 0)
```

N-Wouda (approved):

tests/test_PenaltyManager.py
N-Wouda: (and similarly in the other test)

PR #1194, opened 2026-09-13
Remove search on the initial solution in `solve()`

### LINKED ISSUE #1193: solve() exceeds MaxRuntime on large instances
#### Describe the bug

I was benchmarking a few solvers against each other, each on a matched wall-clock budget. On an instance with 2392 locations and a 30 second budget, PyVRP returned after 34.1 seconds. The other five came back within about half a second of theirs, so PyVRP looked worse than it is.

The extra time goes to the work before the loop starts. `MaxRuntime` takes its first timestamp when it is first called, and by then the neighbourhoods are computed and the initial solution has been through an exhaustive local search. `result.runtime` reports 30.0 and matches its docstring, but the call itself took 34.

A smaller budget makes it easier to see. With 5 seconds:

#### Code sample

```python
import time

import numpy as np
from pyvrp import Client, Depot, Location, ProblemData, VehicleType, solve

rng = np.random.default_rng(42)
coords = rng.integers(0, 10_000, size=(2392, 2))
diff = coords[:, None, :] - coords[None, :, :]
dist = np.round(np.sqrt((diff**2).sum(-1))).astype(np.int64)

data = ProblemData(
    locations=[Location(x=int(x), y=int(y)) for x, y in coords],
    clients=[Client(location=idx) for idx in range(1, 2392)],
    depots=[Depot(location=0)],
    vehicle_types=[VehicleType(num_available=1)],
    distance_matrices=[dist],
    duration_matrices=[dist],
)

start = time.perf_counter()
res = solve(data, stop=MaxRuntime(5), seed=0, display=False)
print(f"wall {time.perf_counter() - start:.1f}s, result.runtime {res.runtime:.1f}s")
```

```
wall 8.6s, result.runtime 5.1s
```

It is not only a single-vehicle thing. The same script with `num_available=100` comes back at 5.9s. Smaller gap, still there.

I went looking for this in the docstrings and on the stopping criteria page and did not find it. From reading the code the same ordering is on main, though I have not built it to check.

If it is worth a change, one option is a line in the `MaxRuntime` docstring about the clock starting on the first call, so the budget reads as covering the search loop rather than the preparation before it. Starting the clock in `__init__` would close the gap, though that would then count whatever happens between constructing the criterion and passing it to `solve`.

```
INSTALLED VERSIONS
------------------
     pyvrp: 0.14.0
     numpy: 2.5.3
matplotlib: 3.11.2
    vrplib: 2.2.0
      tqdm: 4.70.1
    Python: 3.12.0
```

leonlan:
This PR:

- [x] Closes #1193.

1 seed, all instanceds. Results show that removing this doesn't affect performance. Solves the runtime discrepancies as shown in the comment below.

| family        | 1m main | 1m branch | delta | 10m main | 10m branch | delta |
| ------------- | ------- | --------- | ----- | -------- | ---------- | ----- |
| CVRP          | 0.65    | 0.62      | -0.03 | 0.31     | 0.32       | +0.01 |
| VRPTW         | 2.06    | 2.07      | +0.01 | 0.93     | 0.88       | -0.05 |
| PCVRPTW       | 0.82    | 0.77      | -0.05 | 0.42     | 0.43       | +0.01 |
| MDVRPTW       | 2.62    | 2.65      | +0.03 | 1.07     | 1.09       | +0.02 |
| VRPB          | 1.26    | 1.23      | -0.03 | 0.69     | 0.65       | -0.04 |
| HFVRP         | 1.45    | 1.50      | +0.05 | 0.59     | 0.60       | +0.01 |
| MTVRPTWR      | 1.14    | 1.18      | +0.04 | 0.49     | 0.49       | +0.00 |
| PDPTW li-1000 | 5.11    | 4.61      | -0.50 | 0.78     | 0.73       | -0.05 |
| GVRP+ARP      | 0.99    | 0.93      | -0.06 | 0.40     | 0.38       | -0.02 |
| mean          | 1.79    | 1.73      | -0.06 | 0.63     | 0.62       | -0.01 |

<details>

**Notes**:

In particular:

</details>


tanmayBeginsJourney:
Thanks for picking this up so fast. Removing the search entirely had not occurred to me.

I checked it against the case from #1193. Rather than building the branch I applied the diff to the 0.14.0 wheel, since the changed block is byte-identical on 0.14.0 and main. So this is 0.14.0 plus your two lines, not 8027716.

Same instance as the issue: 2392 locations, one vehicle, no capacities or time windows. 15 seeds per build, fresh process per run, builds interleaved seed by seed.

| budget | build | wall | overrun | cost mean | iters mean |
| --- | --- | --- | --- | --- | --- |
| 5s | 0.14.0 | 7.48 - 8.22 | 2.82 | 540344 | 284 |
| 5s | + your diff | 5.37 - 5.50 | 0.43 | 560173 | 127 |
| 30s | 0.14.0 | 32.45 - 33.37 | 2.80 | 419902 | 6469 |
| 30s | + your diff | 30.36 - 30.46 | 0.39 | 424890 | 5667 |

The overrun goes away, with no overlap between the builds at either budget across all 60 runs. The overhead it removes is about 2.8s regardless of budget, so it was 56% of a 5s budget and 9% of a 30s one. `result.runtime` reads 5.00-5.15 and 30.00-30.05 in every run of both builds, so it never saw any of this.

The cost column depends on the budget. At 5s the patched build is 3.67% worse (11 of 15 seeds, paired t = 3.99); at 30s it is 1.19% worse (t = 1.78), which I would not read as anything. One unconstrained single-vehicle instance is close to a TSP and is not your benchmark set, so take it for what it is.

`tests/stop/test_MaxRuntime.py` covers the criterion on its own and `tests/test_solve.py` uses `MaxIterations` throughout, so I could not find anything tying `solve()`'s own runtime to the budget. If a regression test would help, this one measures no time:

```python
res = solve(data, stop=MaxIterations(0), seed=0)
assert res.best == Solution.make_random(data, RandomNumberGenerator(seed=0))
```

It passes with your diff and fails without it, checked both ways on a 60-location instance. Happy to open a PR with that if it is useful.

I used an AI coding assistant for the measurement scripts.

N-Wouda (approved):

